PR #87, opened 2025-10-24
Added helm files

subhadeepTW:
Added helm files for bahmni-app-frontend

### DIFF
--- a/helm/templates/service.yaml
+++ b/helm/templates/service.yaml
@@ -10,8 +10,8 @@ spec:
   type: {{ .Values.service.type }}
   ports:
     - port: {{ .Values.service.port }}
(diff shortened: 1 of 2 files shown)


helm/templates/service.yaml
```diff
@@ -0,0 +1,17 @@
...
+  name: {{ .Chart.Name }}
+  labels:
+    app: {{ .Chart.Name }}
+    environment: {{ .Values.metadata.labels.environment }}
+
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: {{ .Values.service.port }}
+      targetPort: 8091
```
mohan-13: service port should be `80`. The container listens on 80 only

PR #93, opened 2025-11-05
BN-204 | Add. Action Buttons For Appointment Search Handler

subhashydv:
> Add Action buttons on registration page

**JIRA** → [BN-204](https://bahmni.atlassian.net/browse/BN-204)

### **Description**

 

This PR add action button on registration page based on defined configurations to allow user to move ahead in registration journey.

---

### **Key Features**
- Render Action button for appointment dropdown
- Buttons will be rendered based on configration
- It supports few enablement rules (privilege, status, app-date)
- Incase privilege are missing the button will not be rendered.
- Action on button will work based on configuration.

### DIFF
--- a/apps/registration/src/pages/patientSearchPage/index.tsx
+++ b/apps/registration/src/pages/patientSearchPage/index.tsx
@@ -60,18 +60,23 @@ const PatientSearchPage: React.FC = () => {
   const handleCreateNewPatient = () => {
     navigate('/registration/new');
   };
+
+  const getSearchFieldsFromConfig = async (selectedType: string) => {
+    const config = await getRegistrationConfig();
+
+    const fields =
+      selectedType === 'appointment'
+        ? (config?.patientSearch?.appointment ?? [])
+        : (config?.patientSearch?.customAttributes ?? []);
+
+    return fields;
+  };
+
   useEffect(() => {
     const loadSearchConfig = async () => {
-      const config = await getRegistrationConfig();
-      let fields: PatientSearchField[] = [];
-
-      if (selectedFieldType === 'appointment') {
-        fields = [...(config?.patientSearch?.appointment ?? [])];
-      } else {
-        fields = [...(config?.patientSearch?.customAttributes ?? [])];
-      }
-
-      setSearchFields(fields);
+      getSearchFieldsFromConfig(selectedFieldType).then((fields) => {
+        setSearchFields(fields);
+      });
     };
     loadSearchConfig();
   }, [selectedFieldType]);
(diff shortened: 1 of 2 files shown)


apps/registration/src/pages/patientSearchPage/appointmentSearchHandler.ts
```diff
@@ -0,0 +1,124 @@
...
+
+export const handleActionNavigation = (
+  navigationUrl: string,
+  options: Record<string, string>,
+  navigate: NavigateFunction,
+) => {
+  if (!navigationUrl) return;
+
+  const url = formatUrl(navigationUrl, options, true);
+  if (url.startsWith('#')) {
```
angshu: Should this not be flipped?
common convention of URL is /context/path-info etc. 
for routes within an app .. lets prefix with #

apps/registration/src/pages/patientSearchPage/appointmentSearchHandler.ts
```diff
@@ -0,0 +1,124 @@
...
+  userPrivileges: UserPrivilege[],
+): boolean => {
+  const privilegeRules =
+    action.enabledRule
+      ?.filter((rule) => rule.type === 'privilegeCheck')
+      .map((rule) => rule.values)
+      .flat() ?? [];
+
+  if (privilegeRules.length === 0) {
+    return true;
```
angshu: Should we be more stringent on what action options be shown? Through the config schema definition, we can document that for a configured action must have at least one privilege rule defined. 
I would stay defensive and return false for now

apps/registration/src/pages/patientSearchPage/index.tsx
```diff
@@ -165,44 +166,88 @@ const PatientSearchPage: React.FC = () => {
...
+                  );
+                }),
+              )}
+            </Stack>
+          );
       }
 
       const cellValue =
-        row[cellId as keyof PatientSearchViewModel<PatientSearchResult>];
+        row[cellId as keyof PatientSearchViewModel<AppointmentSearchResult>];
```
angshu: Same thing here.

apps/registration/src/pages/patientSearchPage/index.tsx
```diff
@@ -165,44 +166,88 @@ const PatientSearchPage: React.FC = () => {
...
     if (row.uuid) {
       navigateToPatient(row.uuid);
     }
   };
 
-  const renderCell = useCallback(
+  const renderPatientSearchResult = useCallback(
     (
-      row: PatientSearchViewModel<PatientSearchResult>,
+      row: PatientSearchViewModel<AppointmentSearchResult>,
```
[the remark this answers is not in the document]
kousalyapt-tw: AppointmentSearchResult extends the PatientSearchResult . We need the appointment related data to render the cells and so AppointmentSearchResult type is used.

mohan-13 (commented):
@angshu Shouldn't the strategies have unit tests?

PR #94, opened 2025-11-06
BN-201 | Refactor. Registration App Structure For Better Readability

sumaztwcode:
@bahmni/developers 
Kindly review the proposed changes. Your feedback is appreciated

### DIFF
--- a/apps/registration/src/components/forms/addressInfo/AddressInfo.tsx
+++ b/apps/registration/src/components/forms/addressInfo/AddressInfo.tsx
@@ -5,7 +5,14 @@ import {
   type AddressHierarchyEntry,
   type PatientAddress,
 } from '@bahmni-frontend/bahmni-services';
-import { useCallback, useImperativeHandle, useRef, useState } from 'react';
+import { useQuery } from '@tanstack/react-query';
+import {
+  useCallback,
+  useEffect,
+  useImperativeHandle,
+  useRef,
+  useState,
+} from 'react';
 import { AddressData } from '../../../models/patient';
 import type { AddressErrors } from '../../../models/validation';
 import styles from './styles/index.module.scss';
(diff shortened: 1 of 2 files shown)


apps/registration/src/components/forms/patientAdditionalInformation/styles/index.module.scss
```diff
@@ -0,0 +1,27 @@
+@use "@carbon/react/scss/spacing" as *;
+@use "@carbon/react/scss/theme" as *;
+@use "@carbon/react/scss/type" as *;
+
```
PratitiP: Instead of px and rem, can we use CDS spacing token. It will be consistent with other style files

apps/registration/src/components/forms/patientAddressInformation/PatientAddressInformation.tsx
```diff
@@ -0,0 +1,442 @@
...
+        window.clearTimeout(existing);
+      }
+
+      const id = window.setTimeout(async () => {
+        if (!searchText || searchText.length < 2) {
+          setSuggestions((prev) => ({ ...prev, [field]: [] }));
+          setShowSuggestions((prev) => ({ ...prev, [field]: false }));
+          return;
+        }
+
```
PratitiP: Can we use tanstack query here, istead of calling the service function directly

apps/registration/src/components/forms/patientContactInformation/PatientContactInformation.tsx
```diff
@@ -0,0 +1,84 @@
...
+    validate,
+    getData,
+  }));
+
+  return (
+    <div className={styles.formSection}>
+      <span className={styles.formSectionTitle}>
+        {t('CREATE_PATIENT_SECTION_CONTACT_INFO')}
+      </span>
+      <div className={styles.row}>
```
PratitiP: className={styles.phoneNumberField} can be directly passed to TextInput. 
I think extra div wrappers is not needed on both the textInput components

apps/registration/src/pages/createPatientPage/patientFormService.ts
PratitiP: patientFormService - as its just helping formatting or creating formData, can we put it in utils and name it differently. Service name feels like we are using service methods for API handling. 
suggestion - formDataCreator (something similar to clinical utils).

apps/registration/src/providers/RegistrationConfigProvider.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export const RegistrationConfigProvider: React.FC<
+  RegistrationConfigProviderProps
+> = ({ children, initialConfig }) => {
+  const [registrationConfig, setRegistrationConfig] =
+    useState<RegistrationConfig | null>(initialConfig ?? null);
+  const [isLoading, setIsLoading] = useState(!initialConfig);
+  const [error, setError] = useState<Error | null>(null);
+
```
PratitiP: Leverage tanstack query for getRegistrationConfig query function. Then no need to maintain seperate states for error and loading explicitely.

PR #111, opened 2025-11-21
Add publish package workflow

arshiyaTW2021:


.github/workflows/publish-packages.yml
```diff
@@ -0,0 +1,245 @@
...
+          fi
+          
+          if echo "$AFFECTED" | grep -q "widgets"; then
+            echo "has_widgets=true" >> $GITHUB_OUTPUT
+          else
+            echo "has_widgets=false" >> $GITHUB_OUTPUT
+          fi
+
+  # Publish @bahmni/services (no dependencies)
+  # Note: PACKAGE_NAME is the scoped npm name, PACKAGE_PATH is the directory path
```
mohan-13: I see that the job steps are all duplicated. This will be hard to maintain we we start adding apps and packages

arshiyaTW2021: As a next step, I have modified this and removed duplicacies.

PR #112, opened 2025-11-21
BAH - 4244

sumaztwcode:
Kindly review. Items not covered in this include issues identified by Karthik during desk check - 
Middle Name - validation Error should not be happen for non-mandatory
LastName - (NOT REQUIRED), still showing last name is required.
Age fields - (—) are accepting
Birth Time validation (12:— —)


PratitiP (approved):
Validations can be extracted out in separate file and can be simplifies more. 
Styles can be optimised.
These enhancements can be taken later.

PR #119, opened 2025-11-27
BAH-4250 Additional Identifiers section under Registration

sumaztwcode:


apps/registration/src/components/forms/additionalIdentifiers/AdditionalIdentifiers.tsx
```diff
@@ -0,0 +1,125 @@
...
+    return null;
+  };
+
+  return (
+    <div className={styles.formSection}>
+      <SortableDataTable
+        headers={headers}
+        rows={rows}
+        ariaLabel=""
+        renderCell={renderCell}
```
PratitiP: sortable prop is used for enabling sorting on the column headers. As we don't have headers here, I think we can skip this prop as well

apps/registration/src/hooks/useIdentifierTypes.ts
PratitiP: I think separate custom hook is not required as this query can be added directly added inline in useAdditionalIdentifiers.ts 
Its simple tanstack query call and no additional transformation needed here

apps/registration/src/components/forms/additionalIdentifiers/styles/index.module.scss
```diff
@@ -0,0 +1,50 @@
+@use "@carbon/react/scss/spacing" as *;
+@use "@carbon/react/scss/theme" as *;
+@use "@carbon/react/scss/type" as *;
+
+// Form Container
+
```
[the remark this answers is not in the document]
sumaztwcode: Taken care

apps/registration/src/hooks/useIdentifierTypes.ts
sumaztwcode: Taken care

PR #161, opened 2026-01-02
BAH-4365  Filtering with episodeuuid for radiology investigation orders

skjameela276:


packages/bahmni-services/src/radiologyInvestigationService/radiologyInvestigationService.ts
```diff
@@ -53,11 +46,22 @@ function formatRadiologyInvestigations(
...
   patientUUID: string,
+  category: string,
+  encounterUuids?: string[],
+  numberOfVisits?: number,
 ): Promise<RadiologyInvestigation[]> {
-  const bundle = await getPatientRadiologyInvestigationBundle(patientUUID);
+  const bundle = await getServiceRequests(
+    category,
+    patientUUID,
+    encounterUuids,
```
mohan-13: I dont see any assertions which actually validates that the encounteruuids are actually passed onto the GET call if passed

PR #303, opened 2026-03-17
BAH-4457 | Fix ResizeObserver loop error in Carbon ComboBox

keerthi-456:
Description:
  ## Root Cause

  The "ResizeObserver loop completed with undelivered notifications" error
  was caused by Carbon's `autoAlign` prop enabling `@floating-ui/react`'s
  `autoUpdate`, which installs a ResizeObserver on the input and dropdown
  elements. When the ComboBox's clearSelectedOnChange feature triggered
  rapid state transitions (item → null via queueMicrotask), it caused
  multiple synchronous DOM layout changes that exceeded the browser's
  ResizeObserver notification delivery budget.

  ## Fix

  ### 1. ResizeObserver patch — `distro/src/index.html`
  Patched `window.ResizeObserver` to run all callbacks asynchronously via
  `requestAnimationFrame`. This breaks the synchronous loop at the browser
  level — callbacks are deferred to the next animation frame so they can
  never trigger each other synchronously. This is a true fix, not
  suppression: the loop structurally cannot happen.

  ### 2. ComboBox cleanup — `packages/bahmni-design-system/src/atoms/comboBox/index.tsx`
  - Restored the `queueMicrotask` approach (from main) to properly signal
    Downshift to clear its inputValue after selection (item → null sequence)
  - Changed `|| null` to `?? null` to avoid incorrect falsy value coercion
  - Added `handleChange` wrapper to correctly intercept `onChange` without
    introducing side effects on form state

  ### 3. Removed suppression utility — `packages/bahmni-design-system/`
  - Deleted `src/utils/suppressResizeObserverErrors.ts`
  - Removed its export from `src/index.ts`
  - Removed all calls from `setupTests.ts`

  ### 4. Removed suppression from App files
  - Removed `suppressResizeObserverErrors` import and call from
    `apps/clinical/src/App.tsx`
  - Removed from `apps/appointments/src/App.tsx`
  - Removed from `apps/registration/src/App.tsx`
  - Removed from all three `setupTests.ts` files

  ### 5. Deleted incorrect App test files
  - Deleted `apps/clinical/src/__tests__/App.test.tsx`
  - Deleted `apps/appointments/src/__tests__/App.test.tsx`
  - Deleted `apps/registration/src/__tests__/App.test.tsx`
    (These only tested that suppression was called — wrong behavior)

  ### 6. Restored clinical test files
  - Restored all 6 clinical form test files and snapshots to their original
    form (stored variable pattern). The re-querying was needed only because
    of key remounting, which is no longer used.

  ### 7. ComboBox tests updated
  - Removed fake timer setup (no timers in implementation)
  - Removed ResizeObserver suppression tests (not ComboBox's responsibility)
  - Updated user-interaction test to simulate real parent behavior
    (parent updates selectedItem in onChange, matching all real form usage)


ROHIT-V-TW (changes requested):
## Review for BAH-4457

Good approach — key-remount is a clean solution to the item→null feedback loop and Downshift warnings. Found one critical issue with the error message string that may prevent the suppression from working in Chrome.

Found 1 blocking issue, 2 suggestions, and 1 nit.

**Acceptance Criteria**: 5/6 met (ResizeObserver suppression may not match Chrome's exact error string)

ravinderkabli (commented):
## Review for BAH-4457

Good fix that correctly identifies the root cause (Carbon's `autoAlign` + `floating-ui` ResizeObserver interaction) and replaces the global blanket suppression with a co-located one. The key-remount strategy is the right approach for avoiding Downshift's controlled/uncontrolled state confusion. Two logic concerns in the wrapper need attention.

Found 0 blocking issues, 2 suggestions, and 4 nits.

**Acceptance Criteria**: 5/6 met (test coverage gap for `handleChange`-triggered remount path)

### What's Done Well
- Clear root cause analysis in PR description
- Key remount is the correct React idiom for clean Downshift re-init
- `||` → `??` prevents falsy value coercion
- All 5 consumer test suites updated to re-query DOM after remount
- Timer cleanup via `clearTimerRef` with proper unmount guard

ravinderkabli (changes requested):

ravinderkabli (changes requested):
BAH-4457 
The fix is incorrect. It moves the problem without solving it, and introduces new regressions. 
**Suppression multiplied** — went from 1 global location (index.html) to 7 call sites across App.tsx  and setupTests.ts files       
**Wrong package** — suppressResizeObserverErrors is a window error handler, not a UI component; it doesn't belong  in bahmni-design-system   
**Tests rewritten to hide a regression** — 6 test files updated to re-query the DOM because the key remount detaches the element; this masks the DOM instability rather than fixing it  .

The analysis of BUG-4457 should have been from point of view of fixing it and not moving the script to other files.

ROHIT-V-TW:
Check few warnings raised by SonarCloud if valid. Will connect with you on this for a sanity and close this

ravinderkabli (approved):

ROHIT-V-TW (approved):

PR #321, opened 2026-04-02
BAH-4471 | Add. Search Functionality within Patient Dashboard

sravyarongala:
## Summary
- **Add patient search within the Clinical Dashboard**: A new search overlay accessible via the header search icon allows clinicians to search patients by Patient ID without leaving the consultation page. Results appear in a dropdown with keyboard navigation (Arrow keys, Enter, Escape) and click support, navigating directly to the selected patient's dashboard.
- **New components & hooks**: `PatientSearch` component with accessible combobox pattern (ARIA roles, `aria-activedescendant`), `usePatientSearch` hook backed by React Query, and accompanying SCSS styles.
- **Translations**: Added `en` and `es` locale keys for search placeholder, no-results message, loading state, and global action labels.
- **Registration form cleanup**: Simplified `PatientRegister` and `PatientRegisterSection` by removing unused layout/section config logic and associated styles/snapshots.
- **Design system & shared packages**: Minor ComboBox fix, removed flaky ComboBox snapshot tests, bumped `bahmni-services` version, fixed `userService` bug.
- **Test improvements across the board**: Added missing mocks, rewrote flaky integration tests (Diagnoses, Conditions, VitalFlowSheet, ProgramDetails, etc.), improved test isolation, and increased coverage for services (`privilegeService`, `providerService`, `vitalFlowSheetService`) and utility modules.

## Test plan
- [ ] Verify search icon in the Clinical Dashboard header opens the search overlay
- [ ] Search by a known Patient ID → confirm matching patient appears in results
- [ ] Click or press Enter on a result → navigates to that patient's consultation page
- [ ] Press Escape or click outside → closes the search panel
- [ ] Search with no matches → "No matching records" message displayed
- [ ] Verify arrow key navigation highlights results correctly
- [ ] Run full test suite (`yarn test`) — all new and existing tests pass


apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -0,0 +1,192 @@
...
+    setInputValue('');
+    setSubmittedTerm('');
+  };
+
+  const handleResultClick = (result: PatientSearchResult) => {
+    navigate(`/clinical/${result.uuid}`);
+    onClose();
+  };
+
+  const showDropdown =
```
ROHIT-V-TW: blocking: Navigation URL `/clinical/${result.uuid}` may not match the router config.

The router in `App.tsx` registers the route as `<Route path=":patientUuid" .../>` — a relative path under the microfrontend's base. Other parts of the app use `BAHMNI_CLINICAL_PATH` or relative navigation. If the app is mounted at `/bahmni/clinical/`, this absolute path won't resolve to the correct route.

Verify the target — likely should be `navigate(result.uuid)` (relative) or use the `BAHMNI_CLINICAL_PATH` constant.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const [inputValue, setInputValue] = useState('');
+  const [submittedTerm, setSubmittedTerm] = useState('');
+
+  const { data: searchResults } = useQuery({
+    queryKey: ['patientSearch', submittedTerm],
+    queryFn: () => searchPatientByNameOrId(submittedTerm),
+    enabled: submittedTerm.trim().length > 0,
+  });
+
+  // Reset state when the search panel is closed
```
ROHIT-V-TW, line 35: suggestion: No loading state — user sees nothing between pressing Enter and results arriving.

`isLoading` from `useQuery` is available but not destructured. While the query is in-flight, `searchResults` is `undefined` so the dropdown stays hidden. Consider showing a spinner or "Searching..." message when `isLoading && submittedTerm` is truthy.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -0,0 +1,192 @@
...
+ */
+const PatientSearch: React.FC<PatientSearchProps> = ({ isOpen, onClose }) => {
+  const { t } = useTranslation();
+  const navigate = useNavigate();
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  const [inputValue, setInputValue] = useState('');
+  const [submittedTerm, setSubmittedTerm] = useState('');
+
+  const { data: searchResults } = useQuery({
```
ROHIT-V-TW: suggestion: No error handling — if the API call fails (network error, 401, 500), the user sees nothing.

Other form components in this codebase (InvestigationsForm, AllergiesForm) destructure `isError` from `useQuery` and show notifications. Consider adding error feedback here too.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -0,0 +1,192 @@
...
+  };
+
+  const handleResultClick = (result: PatientSearchResult) => {
+    navigate(`/clinical/${result.uuid}`);
+    onClose();
+  };
+
+  const showDropdown =
+    submittedTerm.trim().length > 0 && searchResults !== undefined;
+  const results = (
```
ROHIT-V-TW: suggestion: Stale dropdown — after a search, if the user types a new term but hasn't pressed Enter, the old results remain visible.

`showDropdown` depends on `submittedTerm`, not `inputValue`. The user could see results for "GAN123" while the input shows "GAN456". Consider clearing the dropdown when `inputValue` diverges from `submittedTerm`, or adding `inputValue === submittedTerm` to the `showDropdown` condition.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -0,0 +1,192 @@
...
+    (searchResults?.pageOfResults as PatientSearchResult[]) ?? []
+  ).filter(
+    (patient) =>
+      patient.identifier.toLowerCase() === submittedTerm.toLowerCase(),
+  );
+
```
ROHIT-V-TW, line 140: suggestion: Two concerns with the client-side filter:

1. **Unsafe type cast**: `searchResults?.pageOfResults as PatientSearchResult[]` bypasses the `PatientSearchResult[] | AppointmentSearchResult[]` union without a type guard.
2. **Silent result dropping**: The Lucene API may return valid matches that get filtered out here. If exact-ID-only is the intent, consider a server-side parameter rather than post-filtering.

The registration module's `usePatientSearch.ts` avoids both issues — worth aligning with that pattern.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -0,0 +1,192 @@
...
+          data-testid="patient-search-results"
+          role="listbox"
+          aria-label={t('SEARCH_PATIENT_ID_PLACEHOLDER')}
+        >
+          {results.length === 0 ? (
+            <li
+              className={styles.noResults}
+              data-testid="patient-search-no-results"
+              role="option"
+              aria-selected={false}
```
ROHIT-V-TW, line 185: suggestion: ARIA accessibility gap — `aria-selected={false}` is hardcoded on all options, and the search input lacks `aria-expanded`, `aria-controls`, and `aria-activedescendant`.

For a proper combobox/listbox pattern, track a `focusedIndex` state and set `aria-selected` dynamically. Arrow key navigation between results would also help keyboard-only users. Consider using Carbon's `ComboBox` which handles this automatically (consistent with InvestigationsForm and AllergiesForm).

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -91,7 +69,33 @@ const ConsultationPage: React.FC = () => {
   const { userPrivileges } = useUserPrivilege();
   const { addNotification } = useNotification();
   const [isActionAreaVisible, setIsActionAreaVisible] = useState(false);
+  const [isSearchOpen, setIsSearchOpen] = useState(false);
   const [searchParams] = useSearchParams();
+
+  const globalActions = [
```
ROHIT-V-TW: nit: `globalActions` is recreated every render now that it's inside the component body. The `onClose={() => setIsSearchOpen(false)}` prop is also a new reference each render, causing both `useEffect` hooks in `PatientSearch` to unmount/remount event listeners.

Consider wrapping in `useMemo`/`useCallback` — matches the existing pattern in this file (`episodeUuids`, `currentDashboard`, `sidebarItems` are all memoized).

apps/clinical/src/components/patientSearch/__tests__/PatientSearch.test.tsx
```diff
@@ -0,0 +1,326 @@
...
+        } as PatientSearchResultBundle,
+        isLoading: false,
+        isError: false,
+      } as any);
+
+      render(<PatientSearch {...defaultProps} />);
+      const input = screen.getByPlaceholderText('Search by Patient ID');
+      fireEvent.change(input, { target: { value: 'GAN123456' } });
+      fireEvent.keyDown(input, { key: 'Enter' });
+
```
ROHIT-V-TW, line 426: nit: This test ("has no accessibility violations with search results") calls `render()` a second time, creating a fresh instance without any submitted search term. The `axe` check runs against the empty state, not the state with results.

Use the `container` from the first `render` call (after `waitFor` confirms results are visible) for the `axe` check.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  const { data: searchResults } = useQuery({
+    queryKey: ['patientSearch', submittedTerm],
+    queryFn: () => searchPatientByNameOrId(submittedTerm),
+    enabled: submittedTerm.trim().length > 0,
+  });
+
+  // Reset state when the search panel is closed
+  useEffect(() => {
+    if (!isOpen) {
```
ROHIT-V-TW, line 37: nit: The codebase pattern is to extract query/filter logic into custom hooks under `apps/clinical/src/hooks/` (e.g., `useInvestigationsSearch`, `useMedicationSearch`). There's even a `usePatientSearch.ts` in the registration module. Consider extracting to `apps/clinical/src/hooks/usePatientSearch.ts` for consistency and reusability.

ROHIT-V-TW (changes requested):
## Review for BAH-4471

Good start on the patient search feature. Found a few issues to address before merge — primarily around navigation correctness, missing loading/error states, and accessibility.

Found 1 blocking issue, 5 suggestions, and 4 nits.

**Acceptance Criteria**: 7/9 met (navigation URL needs verification; no loading indicator while fetching)

### Issues not in diff

| File:Line | Type | Issue |
|-----------|------|-------|
| PatientSearch.module.scss | nit | SCSS file should be in a `styles/` subdirectory to match codebase convention (e.g., `patientSearch/styles/PatientSearch.module.scss`) |

apps/clinical/src/hooks/usePatientSearch.ts
```diff
@@ -0,0 +1,43 @@
...
+    queryKey: ['patientSearch', searchTerm],
+    queryFn: () => searchPatientByNameOrId(searchTerm),
+    enabled: searchTerm.trim().length > 0,
+  });
+
+  const results = (searchResults?.pageOfResults ?? [])
+    .filter(isPatientSearchResult)
+    .filter(
+      (patient) =>
+        patient.identifier.toLowerCase() === searchTerm.toLowerCase(),
```
ravinderkabli: suggestion: `results` is recomputed via `.filter()` on every render, creating a new array reference each time. The `useEffect(() => setFocusedIndex(-1), [results])` in `PatientSearch.tsx` will then fire on every render, not just when results actually change.

Wrap in `useMemo` to match the pattern in `useAllergenSearch.ts`:

```typescript
const results = useMemo(() =>
  (searchResults?.pageOfResults ?? [])
    .filter(isPatientSearchResult)
    .filter((patient) => patient.identifier.toLowerCase() === searchTerm.toLowerCase()),
  [searchResults, searchTerm]
);
```

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -0,0 +1,228 @@
...
+    usePatientSearch(submittedTerm);
+
+  const showDropdown =
+    submittedTerm.trim().length > 0 &&
+    inputValue === submittedTerm &&
+    !isLoading &&
+    !isError;
+
+  // Reset state when the search panel is closed
+  useEffect(() => {
```
ravinderkabli, line 36: suggestion: `showDropdown` uses `inputValue === submittedTerm` as a gate, but this reopens the last dropdown if the user edits and then backspaces back to the same text — stale results silently reappear without re-querying.

Clear `submittedTerm` on any input change:

```typescript
const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
  setInputValue(event.target.value);
  setSubmittedTerm('');  // reset so backspace doesn't replay old results
};
```

apps/clinical/src/components/patientSearch/PatientSearch.module.scss
```diff
@@ -0,0 +1,73 @@
+@use "@carbon/react/scss/spacing" as *;
+@use "@carbon/react/scss/colors" as *;
+@use "@carbon/react/scss/theme" as *;
+@use "@carbon/react/scss/type" as *;
+
+.patientSearchContainer {
+  position: fixed;
+  top: 0;
+  right: 6rem;
+  z-index: 9000;
```
ravinderkabli: suggestion: `right: 6rem` is a magic number derived from the current header icon row width — if the header layout changes this will silently misalign the search panel. CLAUDE.md says to avoid hardcoded spacing.

At minimum, extract to a named variable with a comment:
```scss
$header-actions-offset: 6rem; // offset to clear global action buttons in header

.patientSearchContainer {
  right: $header-actions-offset;
```
Also worth adding a comment above `z-index: 9000` explaining the z-index choice.

ravinderkabli (commented):
## Review for BAH-4471

Solid implementation with good ARIA support and a comprehensive test suite. Five suggestions to address before merge.

Found 0 blocking issues, 5 suggestions, and 3 nits.

**Acceptance Criteria**: 9/9 met

### Issues not in diff

| File:Line | Type | Issue |
|-----------|------|-------|
| apps/clinical/src/hooks/usePatientSearch.ts | suggestion | Missing unit tests for this hook — `PatientSearch.test.tsx` mocks it entirely, so `isPatientSearchResult` type guard and exact-match filter are untested. Per CLAUDE.md 90% coverage threshold, add `__tests__/usePatientSearch.test.ts` covering the filter logic, type guard, and enabled guard. |

ROHIT-V-TW (approved):
Looks good to me, will merge once I get a formal review from Ravinder

ravinderkabli (commented):
## PR #321 Code Review — BAH-4471: Search Functionality in New Clinical Module

Well-structured implementation. Clean separation into `usePatientSearch` hook + `PatientSearch` component, proper ARIA combobox semantics, thorough test coverage with axe accessibility checks, and both locale files updated. Submit-on-Enter with exact identifier matching aligns with the narrowed JIRA scope.

| Severity | Count |
|----------|-------|
| 🚫 BLOCKER | 0 |
| 🔴 CRITICAL | 0 |
| 🟠 MAJOR | 1 |
| 🟡 MINOR | 3 |
| 🔵 INFO | 2 |

**Recommendation: NEEDS DISCUSSION**

See inline comments for details on the items below.

---

### What's Done Well
1. **Clean separation of concerns** — hook for server state, component for UI, page for orchestration.
2. **Correct ARIA combobox pattern** — `role="combobox"`, `role="listbox"`, `aria-activedescendant`, `aria-expanded` / `aria-controls`.
3. **Thorough test coverage** — 437 lines of unit tests + axe accessibility checks + 197 lines of hook tests.
4. **Both locale files updated** — `locale_en.json` and `locale_es.json` with all 6 new keys.
5. **Smart reuse of existing API** — `searchPatientByNameOrId` from `@bahmni/services` with client-side exact-match filter.
6. **GlobalActions now translatable** — moved from module-level constant into `useMemo` with `t()` calls.

apps/clinical/src/components/patientSearch/styles/PatientSearch.module.scss
```diff
@@ -0,0 +1,75 @@
...
+  top: 100%;
+  right: 0;
+  width: 100%;
+  list-style: none;
+  margin: 0;
+  padding: 0;
+  background: $layer;
+  border: 1px solid $border-subtle-01;
+  border-top: none;
+  box-shadow: 0 4px 8px $shadow;
```
ravinderkabli: nit: Hardcoded pixel values in `box-shadow` + fragile header offset

`box-shadow: 0 4px 8px $shadow` uses literal pixel values. CLAUDE.md requires: "always use Carbon SCSS tokens — never hardcode colors, spacing, or font sizes." Carbon spacing tokens map: `$spacing-02` = 4px, `$spacing-03` = 8px.

Also, `$header-actions-offset: 6rem` (line 6) is a local variable coupled to the header's action button layout. If header actions change (e.g., adding a fourth button), this offset silently becomes wrong. Consider documenting the derivation (e.g., `// 3 action buttons × 2rem each`) or exporting the value from the header styles.

```scss
// box-shadow with Carbon tokens:
box-shadow: 0 $spacing-02 $spacing-03 $shadow;
```

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -230,6 +246,7 @@ const ConsultationPage: React.FC = () => {
 
   return (
     <ClinicalAppProvider episodeUuids={episodeUuids}>
+      <PatientSearch isOpen={isSearchOpen} onClose={handleSearchClose} />
```
ravinderkabli, line 251: nit: Search overlay rendered outside layout tree

`PatientSearch` is placed before `ActionAreaLayout` (as a sibling inside `ClinicalAppProvider`), relying entirely on `position: fixed; top: 0; right: 6rem; z-index: 9000` to appear in the header area. This couples a visual overlay concern to a data-provider boundary rather than the layout system.

The existing pattern for overlays (modals, notifications) is to use portal-based rendering or mount within the layout tree. If the header or `ActionAreaLayout` changes position/size, this component won't adapt — the `6rem` offset and `top: 0` are hardcoded assumptions about header geometry.

Consider rendering inside `ActionAreaLayout` (e.g., in the header slot) or using a portal. If fixed positioning is the right call here, a comment explaining why it must be outside the layout tree would help future maintainers.

apps/clinical/src/components/patientSearch/styles/PatientSearch.module.scss
ROHIT-V-TW: Addressed this

apps/clinical/src/pages/ConsultationPage.tsx
ROHIT-V-TW: Acknowledged — added a comment explaining the placement. It's rendered outside ActionAreaLayout because ActionAreaLayout doesn't expose a slot in the header area where we can inject arbitrary overlays (the headerWSideNav prop expects the Header component). Moving it inside the Header slot would require changes to the design-system's ActionAreaLayout/Header API to support an overlay slot. Fixed positioning outside the layout tree was the simplest approach for now. Open to moving to a portal or a header slot if ActionAreaLayout is extended to support it in the future.

ravinderkabli (approved):
Looks fine

PR #329, opened 2026-04-07
BAH-4521 | Fix. Browser date format issue

kousalyapt-tw:
This PR fixes a browser date format issue by implementing locale-based date format detection instead of using a hardcoded format string.


packages/bahmni-services/src/date/date.ts
```diff
@@ -155,6 +154,41 @@ function formatDateGeneric(
   return { formattedResult: format(dateToFormat, dateFormat) };
 }
```
PratitiP, line 156: Can we please precise the function description? 
I think we can remove some part from implementation as its self explanatory and repetitive

kousalyapt-tw: updated the function description

mohan-13 (approved):

mohan-13:
Duplicate exclusion has been added to date utils in design system to avoid dependency between bahmni-services package and design-system package

PR #330, opened 2026-04-08
BAH-4590 | Add Collapsible/Expandable Sections to Registration Form

hamsavarthiniR-Bahmni:
## Description

As a registration staff member, I want to collapse/expand non-primary form sections so that I can reduce visual clutter and focus on relevant sections, while keeping the first section always visible.


apps/registration/src/pages/PatientRegister/styles/index.module.scss
```diff
@@ -266,4 +266,42 @@
...
+  background: none;
+  border: none;
+  padding: 0;
+  cursor: pointer;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  text-align: left;
+
+  &:focus {
```
ravinderkabli: **blocking**: `outline: none` removes the keyboard focus indicator entirely — this violates WCAG 2.1 SC 2.4.7 (Focus Visible). Keyboard users tabbing to the toggle button will see no visual indication of focus.

Replace with a Carbon-token-based focus ring:
```scss
&:focus-visible {
  outline: 2px solid $focus;
  outline-offset: -2px;
}
```
Or better yet, use `Accordion`/`AccordionItem` from `@bahmni/design-system` which handles focus correctly out of the box.

apps/registration/src/pages/PatientRegister/PatientRegisterSection.tsx
```diff
@@ -59,16 +66,53 @@ const PatientRegisterSection: React.FC<PatientRegisterSectionProps> = ({
...
   }
 
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLButtonElement>) => {
+    if (e.key === 'Enter' || e.key === ' ') {
+      e.preventDefault();
+      onToggle();
+    }
+  };
+
   return (
```
ravinderkabli: **blocking**: This builds a custom collapsible mechanism (button + chevron + display:none + onKeyDown + aria-expanded) from scratch, but `@bahmni/design-system` already exports `Accordion`/`AccordionItem` wrappers that provide all of this out of the box — focus management, keyboard interaction, ARIA semantics, and animation.

CLAUDE.md: _"Don't create custom button, input, modal, or table components — Carbon provides these"_ and _"Don't import Carbon directly in feature code when a @bahmni/design-system wrapper exists."_

Consider using `<AccordionItem title={t(section.translationKey)} open={isExpanded} onHeadingClick={onToggle}>` — this also resolves the focus indicator issue.

apps/registration/src/pages/PatientRegister/PatientRegister.tsx
```diff
@@ -127,6 +153,30 @@ const PatientRegister = () => {
     );
 
     if (!isValid) {
```
ravinderkabli, line 155: **suggestion**: `validate()` is called on every ref here a second time — `validateAllSections` already called them above. This double-invocation fires error-state side effects twice per failed save (potential UI flicker or duplicate error messages).

Consider refactoring `validateAllSections` to return per-section validity results, then use that to drive the expand logic without re-invoking `validate()`.

apps/registration/src/pages/PatientRegister/PatientRegister.tsx
```diff
@@ -127,6 +153,30 @@ const PatientRegister = () => {
...
+        additionalIdentifiers: () =>
+          shouldShowAdditionalIdentifiers
+            ? (patientAdditionalIdentifiersRef.current?.validate() ?? true)
+            : true,
+        relationships: () =>
+          patientRelationshipsRef.current?.validate() ?? true,
+      };
+
+      const sectionsToExpand = new Set(expandedSections);
+      sections.forEach((section) => {
```
ravinderkabli, line 170: **suggestion**: The fallback `?? true` means unknown control types default to "no error" — if a new control type is added to the config but not to `controlValidators`, that section will never auto-expand even when `validateAllSections` fails.

Consider changing to `?? false` so unknown types default to "has error" (safer — the section expands rather than hiding the error).

apps/registration/src/pages/PatientRegister/__tests__/PatientRegister.test.tsx
```diff
@@ -1126,4 +1126,186 @@ describe('PatientRegister', () => {
...
+
+      await waitFor(() => {
+        // Verification is that component doesn't throw and state updates properly
+        expect(screen.getByTestId('patient-profile')).toBeInTheDocument();
+      });
+    });
+
+    it('should auto-expand sections with validation errors on save', async () => {
+      // Mock validation to fail
+      (validateAllSections as jest.Mock).mockReturnValue(false);
```
ravinderkabli, line 1242: **suggestion**: These auto-expand tests mock `validateAllSections` to return `false` but never mock any ref's `.validate()` to return `false`. Since refs are `createRef()` with `.current = null`, the validators always evaluate to `true` (no errors), so `sectionsToExpand` is always empty — the auto-expand logic is never actually exercised.

To fix: mock the config to include sections with specific control types, mock the corresponding ref's `.validate()` to return `false`, then assert the section content loses the `sectionContentCollapsed` class.

apps/registration/src/pages/PatientRegister/__tests__/PatientRegister.test.tsx
```diff
@@ -1126,4 +1126,186 @@ describe('PatientRegister', () => {
...
+      renderComponent();
+
+      // All section headers with translationKey should be rendered
+      const sectionHeaders = screen.getAllByTestId('section-header-tile');
+      expect(sectionHeaders.length).toBeGreaterThan(0);
+    });
+
+    it('should toggle section expanded state when clicking section header', async () => {
+      renderComponent();
+
```
ravinderkabli, line 1191: **suggestion**: This test clicks `section-header-tile` (the outer Tile div), but `onClick` is on the inner `<button data-testid="collapsible-toggle-button">`. Clicking the Tile won't trigger `toggleSection`. The assertion then checks `section-content` elements exist, which is always true.

Use `getAllByTestId('collapsible-toggle-button')` and assert on `aria-expanded` before/after the click.

ravinderkabli (changes requested):
## Review for BAH-4590

The collapsible section feature works end-to-end and the ARIA wiring is a good start. However there are two blocking accessibility/architecture issues that must be resolved before merge, plus four significant suggestions.

Found **2 blocking issue(s)**, **4 suggestion(s)**, and **3 nit(s)**.

**Acceptance Criteria**: 7/10 met — AC1-AC5 and AC7-AC8 appear satisfied; AC6 (section auto-expands and scrolls into view on failed save) is partially verified but the test coverage does not actually exercise the expand logic (see inline comment on line 1206).

---

### Additional issues (not pinned inline)

| # | Severity | Location | Summary |
|---|----------|----------|---------|
| 7 | MINOR | `PatientRegister.test.tsx` ~L1155 | **Weak multi-section collapse assertion** — test asserts the number of `.sectionContentCollapsed` elements but never checks *which* section collapsed; a swap bug would pass undetected. Assert on the specific section's content element instead. |
| 8 | MINOR | `PatientRegister.test.tsx` ~L1185 | **AC6 test only verifies first section** — the test config registers two sections with errors but only asserts `sectionsToExpand[0]`; the second section's expand behaviour is never verified. |
| 9 | MINOR | `PatientRegister.tsx` — `sections` declaration | **Missing `useCallback`/`useMemo`** — `handleSavePatient` and `toggleSection` are recreated on every render and passed as props; wrap them in `useCallback` to prevent unnecessary child re-renders. |
| 10 | INFO | `PatientRegisterSection.tsx` L3 | **Import ordering** — `React` import should come before the `@carbon` import to satisfy the configured `import/order` rule (builtin → external → internal). Run `yarn lint:fix` to auto-correct. |

apps/registration/src/pages/PatientRegister/PatientRegisterSection.tsx
hamsavarthiniR-Bahmni: done

apps/registration/src/pages/PatientRegister/styles/index.module.scss
hamsavarthiniR-Bahmni: done

apps/registration/src/pages/PatientRegister/PatientRegister.tsx
hamsavarthiniR-Bahmni: done

apps/registration/src/pages/PatientRegister/__tests__/PatientRegister.test.tsx
hamsavarthiniR-Bahmni: done

ravinderkabli (approved):

PR #335, opened 2026-04-13
BAH-4593, BAH-4594 | Refactor. Event-Driven Consultation Pad with Form Registry

rahu1ramesh:
**JIRA** → [BAH-4593](https://bahmni.atlassian.net/browse/BAH-4593)
**JIRA** → [BAH-4594](https://bahmni.atlassian.net/browse/BAH-4594)

### **Description**

Refactors the monolithic ConsultationPad component into a modular, event-driven architecture. The existing ~560-line component and its large test files are replaced with a FormRegistry-based system that declaratively registers form components alongside their validation, bundle creation, and state subscription logic. A new `startConsultation` event allows external triggers (e.g. the patient header action button) to open the consultation pad without direct component coupling.

### **Key Features**
  - Form Registry — each form (allergies, medications, vaccinations, observations, etc.) is registered with a config object describing its privilege, encounter types, reset/validate/hasData/subscribe/createBundleEntries behaviour.
  - Event-Driven Pad Opening — a startConsultation custom event with a dispatch helper and subscriber hook replaces the prop-drilling or direct state coupling previously used to open the consultation pad.
  - Observation Forms Panel — a dedicated ObservationFormsPanel component renders dynamically registered observation forms, replacing the hardcoded layout in ConsultationPad.
  - FormRenderer — a lightweight FormRenderer wrapper renders any registered form component.

[BAH-4593]: https://bahmni.atlassian.net/browse/BAH-4593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[BAH-4594]: https://bahmni.atlassian.net/browse/BAH-4594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/clinical/src/components/consultationPad/services.ts
PratitiP, line 1: Instead of creating separate service file here, can we move submitConsultation service function to existing consultationBundleService in the common services folder ?

rahu1ramesh: But this submit is only triggered by the consultationPad right? I don't see another use case for it now, we can move it to a common service if some other service uses it later.

apps/clinical/src/components/consultationPad/models.ts
PratitiP, line 1: In the clinical, We have Models, services in the separate folder. Good to follow the consistent structure. We already have a models file for consultationBundle

apps/clinical/src/events/startConsultation/event.ts
PratitiP, line 1: Earlier for consultationSave event we added a dispatch event function and the useSubscribe hook in the @bahmni/service -> event -> consultationEvents.ts
Should we add dispatchConsultationStart and useSubscribeConsultationStart there itself to keep them together for maintainability ?

PratitiP (approved):

apps/clinical/src/components/consultationPad/services.ts
PratitiP: Okay. so we need to refactor the existing structure for clinical

PR #336, opened 2026-04-14
BAH-4565 | Apply pagination for all display controls which have SortableDataTable (no Accordions) - DocumentsTable

hamsavarthiniR-Bahmni:
## Description

As a developer, I need pagination support in SortableDataTable so that widgets can implement pagination consistently.As part of this card apply the pagination for DocumentsTable alone.

Note: pageSize: controlled by “standard_config” parameter

PR #340, opened 2026-04-16
BAH-4605 | Apply pagination for all display controls which have SortableDataTable (no Accordions) - DiagnosesTable, ConditionsTable, PatientProgramsTable

hamsavarthiniR-Bahmni:
## Description

As a developer, I need pagination support in SortableDataTable so that widgets can implement pagination consistently.

Identified widgets include PatientProgramsTable, DiagnosesTable, ConditionsTable

Note: pageSize: controlled by “standard_config” parameter, default is 10

### DIFF
--- a/apps/clinical/src/components/dashboardSection/styles/DashboardSection.module.scss
+++ b/apps/clinical/src/components/dashboardSection/styles/DashboardSection.module.scss
@@ -16,8 +16,7 @@
 }
 
 .divider {
-  height: $spacing-07;
-  background-color: $layer-01;
+  height: 0;
 }
 
 .widgetLoading {
--- a/apps/clinical/src/components/dashboardSection/styles/DashboardSection.module.scss
+++ b/apps/clinical/src/components/dashboardSection/styles/DashboardSection.module.scss
@@ -16,7 +16,7 @@
 }
 
 .divider {
-  height: 0;
+  display: none;
 }
 
 .widgetLoading {
--- a/packages/bahmni-services/src/conditionService/conditionService.ts
+++ b/packages/bahmni-services/src/conditionService/conditionService.ts
@@ -2,7 +2,6 @@ import { Condition, Bundle } from 'fhir/r4';
 import { get } from '../api';
 import { PATIENT_CONDITION_RESOURCE_URL } from './constants';
 
-// TODO: Add Optional parameters for pagination and filtering
 /**
  * Fetches conditions for a given patient UUID from the FHIR R4 endpoint
  * @param patientUUID - The UUID of the patient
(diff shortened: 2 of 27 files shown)


packages/bahmni-widgets/src/conditions/ConditionsTable.tsx
```diff
@@ -1,39 +1,49 @@
...
-  const [conditions, setConditions] = useState<ConditionViewModel[]>([]);
+const ConditionsTable: React.FC<WidgetProps> = ({ config }) => {
+  // Number() safely handles non-numeric config values (NaN → falsy → fallback 10)
+  const configPageSize = Number(config?.pageSize) || 10;
   const patientUUID = usePatientUUID();
   const { t } = useTranslation();
   const { addNotification } = useNotification();
+
+  const [currentPage, setCurrentPage] = useState(1);
+  const [selectedPageSize, setSelectedPageSize] = useState(configPageSize);
```
ravinderkabli, line 30: suggestion: `serverTotal` derived-state anti-pattern — same pattern flagged in PR #336.

Syncing `data.total` into state via `useEffect` causes a double-render on each fetch and requires manual `setServerTotal(undefined)` resets in 3 places. This pattern repeats in `DiagnosesTable.tsx` and `PatientProgramsTable.tsx`.

Pass `data?.total` directly to `totalItems` and remove the state and sync effect:
```tsx
<SortableDataTable
  totalItems={data?.total}
  // ...
/>
```
For the page-size-change reset, use `placeholderData: (prev) => prev` in `useQuery` to keep previous page data visible during the transition, avoiding the flash-to-empty that motivated the explicit reset.

packages/bahmni-services/src/conditionService/conditionService.ts
```diff
@@ -28,3 +30,35 @@ export async function getConditions(patientUUID: string): Promise<Condition[]> {
...
+  const offset = (page - 1) * count;
+  const bundle = await get<Bundle>(
+    PATIENT_CONDITION_PAGE_URL(patientUUID, count, offset),
+  );
+  const conditions =
+    bundle.entry
+      ?.filter((entry) => entry.resource?.resourceType === 'Condition')
+      .map((entry) => entry.resource as Condition) ?? [];
+  return {
+    conditions,
```
ravinderkabli, line 61: suggestion: If the FHIR server omits `bundle.total`, this returns `conditions.length` (= current page size, e.g. 10) as the total. Since `total === pageSize`, `showPagination` in SortableDataTable evaluates to `false` — pagination controls permanently disappear and the user is stuck on page 1. The same fallback applies in `diagnosesService.ts` and `programService.ts`.

This was flagged in PR #336 as well. Two options:
1. Pass `bundle.total` directly (allowing `undefined`) and handle it in the widget: show pagination optimistically when `total === undefined && rows.length === pageSize`.
2. At minimum, add a comment: `// bundle.total must be present — pagination hidden if server omits it`.

ravinderkabli (commented):
## Review for BAH-4605

Good implementation — the separate `*_PAGE_URL` constants pattern is a clear improvement over PR #336. Found 2 suggestions.

Found 0 blocking issues, 2 suggestions, and 0 nits.

**Acceptance Criteria**: 6/7 met (PatientProgramsTable sort is per-page only due to server limitation — documented in comments).

apps/clinical/src/components/dashboardSection/styles/DashboardSection.module.scss
```diff
@@ -16,8 +16,7 @@
 }
 
 .divider {
-  height: $spacing-07;
-  background-color: $layer-01;
+  height: 0;
```
ravinderkabli: nit: `height: 0` makes the element invisible but leaves an empty `<div>` in the DOM and accessibility tree. If the intent is to remove the visual gap between widgets, `display: none` is more explicit (and skipped by screen readers)

ravinderkabli (commented):
One additional nit on the divider change.

packages/bahmni-widgets/src/conditions/ConditionsTable.tsx
hamsavarthiniR-Bahmni: done

packages/bahmni-services/src/conditionService/conditionService.ts
hamsavarthiniR-Bahmni: done

apps/clinical/src/components/dashboardSection/styles/DashboardSection.module.scss
hamsavarthiniR-Bahmni: done

ravinderkabli (approved):
Looks fine.

PR #342, opened 2026-04-17
BAH-4479 | Add the missing fields in the GET Document Reference API

Varun-beta:
>  [!NOTE]                                                                                                   
 >  Thank you for your contribution. Please find the details of this pull request below.                      
                                                                                                              
  JIRA → BAH-4479                                                                                             
                                                                                                              
  Description                                                                                                 
                                                                                                              
  Fix document references not returning in newest-first order due to the FHIR server not honouring            
  _sort=-date. Updated to _sort=-_lastUpdated which the server correctly supports.                            
                                                                                                              
  ---                                                                                                       
  Implementation Details                                                                                    
                                                                                                              
  - API Integration
    - Updated DocumentReference FHIR query param from _sort=-date to _sort=-_lastUpdated in                   
  documentReferenceService/constants.ts.                                                                      
                                                                                                            
  ---                                                                                                         
  Testing & Type Definitions                                                                                
                                                                                                            
  - Unit Tests
    - Updated BASE_URL constant in documentReferenceService.test.ts to reflect the new sort param.
                                                                                                              
  ---
  Reviewer(s)                                                                                                 
                                                                                                            
  @bahnew/developers                                                                                        
  Kindly review the proposed changes when convenient. Your feedback is appreciated.


packages/bahmni-services/src/documentReferenceService/constants.ts
```diff
@@ -4,7 +4,7 @@ export const PATIENT_DOCUMENT_REFERENCES_URL = (
   patientUuid: string,
   encounterUuids?: string[],
 ): string => {
-  const baseUrl = `${OPENMRS_FHIR_R4}/DocumentReference?patient=${patientUuid}&_sort=-date&_count=100`;
+  const baseUrl = `${OPENMRS_FHIR_R4}/DocumentReference?patient=${patientUuid}&_sort=-_lastUpdated&_count=100`;
```
ravinderkabli: blocking: This PR branch diverged from `main` after PR #336 (BAH-4565) landed. On `main`, `PATIENT_DOCUMENT_REFERENCES_URL` now accepts `count` and `offset` parameters for pagination. Merging as-is will:

1. Remove `count`/`offset` params, breaking `getDocumentReferencePage()` at runtime
2. Remove `getDocumentReferencePage` and `DocumentReferencePage` type entirely
3. Drop 8 pagination tests
4. Break `DocumentsTable.tsx` which imports `getDocumentReferencePage`

Please rebase on `main` and apply the `_sort=-_lastUpdated` fix within the pagination-aware signature.

packages/bahmni-design-system/src/molecules/imageTile/styles/ImageTile.module.scss
```diff
@@ -14,6 +14,7 @@
   display: flex;
   align-items: center;
   justify-content: center;
+  padding: $spacing-02;
```
ravinderkabli: suggestion: `padding: $spacing-02` is added unconditionally to `.thumbnailButton`. Since `.thumbnailImage` uses `width: 100%; height: 100%`, this subtly shrinks all rendered thumbnails (even when `hideThumbnail` is `false`).

Consider moving the padding inside `.hideThumbnail` so it only applies when the icon is shown:
```scss
.hideThumbnail {
  background-color: $background-brand;
  color: $background;
  padding: $spacing-02;
}
```
Same applies to `VideoTile.module.scss`.

packages/bahmni-widgets/src/forms/FormsTable.tsx
```diff
@@ -46,6 +46,7 @@ const FormsTable: React.FC<WidgetProps> = ({
   const [selectedRecord, setSelectedRecord] =
     useState<FormRecordViewModel | null>(null);
   const numberOfVisits = config?.numberOfVisits as number;
+  const hideThumbnail = !!config?.hideThumbnail;
```
ravinderkabli: suggestion: `config` is typed as `Record<string, unknown>`, so `config?.hideThumbnail` is `unknown`. The `!!` coercion means a string like `"false"` from JSON config would evaluate as `true`.

`Observations.tsx` correctly uses a typed `ObservationConfig` interface with `hideThumbnail?: boolean`. Consider defining an equivalent typed config interface for `FormsTable` to get the same type safety.

packages/bahmni-services/src/index.ts
```diff
@@ -306,3 +306,5 @@ export {
   type DocumentViewModel,
   type DocumentReference,
 } from './documentReferenceService';
+
+export { uploadDocument } from './documentUploadService';
```
ravinderkabli, line 324: suggestion: Every other service in this file exports its response/model types alongside functions (e.g., `type CreatePatientResponse`, `type DocumentViewModel`). Consumers of `uploadDocument` who need to type the return value can't import `DocumentUploadResponse` or `ProcessedFileData` from `@bahmni/services`.

```typescript
export { uploadDocument } from './documentUploadService';
export type { DocumentUploadResponse, ProcessedFileData } from './documentUploadService';
```

packages/bahmni-services/src/documentUploadService/utils.ts
```diff
@@ -0,0 +1,37 @@
+import { DEFAULT_FILE_TYPE } from './constants';
+import { ProcessedFileData } from './models';
+
+export async function processFileForUpload(
+  file: File,
+): Promise<ProcessedFileData> {
+  return new Promise((resolve, reject) => {
+    const reader = new FileReader();
+    reader.onload = () => {
+      const base64Content = (reader.result as string).split(',')[1];
```
ravinderkabli, line 10: nit: For a zero-byte file, `readAsDataURL` produces `"data:application/octet-stream;base64,"` — so `split(',')[1]` yields `''`. The upload proceeds with `content: ''`, likely failing at the OpenMRS layer with an unhelpful error.

Consider adding a guard: `if (!base64Content) { reject(new Error('File is empty or could not be read')); return; }`

packages/bahmni-services/src/documentUploadService/utils.ts
```diff
@@ -0,0 +1,37 @@
...
+      const fileType = file.type ? file.type.split('/')[0] : DEFAULT_FILE_TYPE;
+
+      resolve({
+        content: base64Content,
+        fileName,
+        fileType,
+        format,
+      });
+    };
+
```
ravinderkabli, line 34: nit: The `FileReader` error event contains the original `DOMException` (via `reader.error`) with specific error codes (`NotFoundError`, `NotReadableError`, `SecurityError`). Replacing it with a generic string loses diagnostic info.

```typescript
reader.onerror = () => {
  reject(reader.error ?? new Error('Failed to read file'));
};
```

packages/bahmni-widgets/src/observations/Observations.tsx
```diff
@@ -35,7 +36,12 @@ export const observationsQueryKeys = (
 
 const Observations: React.FC<WidgetProps> = ({ config }) => {
   const observationConfig = config as ObservationConfig;
-  const { conceptNames = [], conceptUuid = [] } = observationConfig;
+  const {
+    conceptNames = [],
+    conceptUuid = [],
```
ravinderkabli, line 41: nit: The rename to `hideThumbnailConfig` only to immediately `!!` it adds an unnecessary intermediate step. Consider using a default value in destructuring:

```typescript
const { conceptNames = [], conceptUuid = [], hideThumbnail = false } = observationConfig;
```

This also aligns more closely with how `FormsTable.tsx` handles it.

ravinderkabli (changes requested):
## Review for BAH-4479

Good progress on the upload service and hideThumbnail feature — both are well-structured and follow project patterns. However, the branch needs a rebase on `main` before merge due to a conflict with PR #336's pagination changes.

Found 1 blocking issue, 3 suggestions, and 5 nits.

**Acceptance Criteria**: 2/11 directly addressed (upload service, image/video tile styling). Remaining 9 AC scenarios (date display, timezone, localization, fallback) were already implemented on `main` and are not regressed.

### Issues not in diff

| File:Line | Type | Issue |
|-----------|------|-------|
| ObsByEncounter.test.tsx | nit | No test coverage for `hideThumbnail={true}` through `renderObservation` |
| ObservationItem.test.tsx | nit | No test verifying `hideThumbnail` prop is forwarded to ImageTile/VideoTile |
| ImageTile.test.tsx / VideoTile.test.tsx | nit | `hideThumbnail=true` tests don't assert `.hideThumbnail` CSS class on button, and don't test modal still opens |

packages/bahmni-services/src/documentReferenceService/constants.ts
Varun-beta: Fixed — rebased on `main` and applied the `_sort=-_lastUpdated` fix within the pagination-aware `getDocumentReferencePage` signature. Also updated the 3 test expectations in `documentReferenceService.test.ts` that were still asserting `_sort=-date`.

packages/bahmni-design-system/src/molecules/imageTile/styles/ImageTile.module.scss
Varun-beta: Fixed — removed `padding: $spacing-02` from `.thumbnailButton` and moved it inside `.hideThumbnail` in both `ImageTile.module.scss` and `VideoTile.module.scss`. Normal thumbnails no longer have the padding applied, preserving the full `width: 100%; height: 100%` image layout.

packages/bahmni-widgets/src/forms/FormsTable.tsx
Varun-beta: Fixed — added a `FormsTableConfig` interface with `numberOfVisits?: number` and `hideThumbnail?: boolean`, mirroring the `ObservationConfig` pattern in `Observations.tsx`. Now using `const { numberOfVisits, hideThumbnail = false } = (config ?? {}) as FormsTableConfig` — no more `unknown` coercion via `!!`.

packages/bahmni-services/src/index.ts
Varun-beta: Fixed — added `export type { DocumentUploadResponse, ProcessedFileData }` to both `documentUploadService/index.ts` and the top-level `src/index.ts`, following the same pattern as other services in the file.

packages/bahmni-services/src/documentUploadService/utils.ts
Varun-beta: Fixed — added a guard after the split: if `!base64Content`, we now `reject(new Error('File is empty or could not be read'))` immediately rather than proceeding with an empty string to OpenMRS.

packages/bahmni-services/src/documentUploadService/utils.ts
Varun-beta: Fixed — `reader.onerror` now uses `reject(reader.error ?? new Error('Failed to read file'))`, preserving the original `DOMException` (with its `NotFoundError`/`NotReadableError`/`SecurityError` codes) for better diagnostics.

packages/bahmni-widgets/src/observations/Observations.tsx
Varun-beta: Fixed — removed the intermediate `hideThumbnailConfig` rename and the `!!` coercion. Now using `hideThumbnail = false` directly in the destructuring, consistent with how `conceptNames` and `conceptUuid` defaults are handled in the same block.

ravinderkabli (approved):
Looks good

PR #346, opened 2026-04-20
BAH-4606 | Apply pagination for all display controls which have SortableDataTable (no Accordions) - AppointmentsTable

hamsavarthiniR-Bahmni:
## Description

As a developer, I need pagination support in SortableDataTable so that widgets can implement pagination consistently.

Identified widgets include AppointmentsTable

Note: pageSize: controlled by “standard_config” parameter, default is 10


packages/bahmni-widgets/src/appointments/hooks/usePaginatedAppointments.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (data) {
+      setServerTotal(data.total);
+    }
+  }, [data]);
+
+  useEffect(() => {
+    setCurrentPage(1);
+    setServerTotal(undefined);
+  }, [patientUUID]);
+
```
ravinderkabli, line 69: suggestion: This subscription checks `payload.updatedResources.conditions` to decide whether to refetch appointments — but conditions has nothing to do with appointments.

Every other widget subscribes to its domain flag: AllergiesTable → `.allergies`, MedicationsTable → `.medications`, DiagnosesTable → `.conditions`. Appointments have no corresponding field in `ConsultationSavedEventPayload.updatedResources`. As written, the appointments list refetches whenever a condition/diagnosis is saved, and never refetches for any appointment-related change.

Since appointments aren't created or modified through the consultation save flow, I'd suggest removing this subscription entirely. If auto-refresh on consultation save is needed in the future, an `appointments` key should first be added to `ConsultationSavedEventPayload`.

packages/bahmni-services/src/appointmentService/constants.ts
```diff
@@ -62,3 +62,21 @@ export const ALL_APPOINTMENT_SERVICES_URL =
 
 export const getDeleteAppointmentServiceUrl = (uuid: string): string =>
   `${OPENMRS_REST_V1}/appointmentService?uuid=${uuid}`;
+
+export const getUpcomingAppointmentsPageUrl = (
```
ravinderkabli, line 71: suggestion: These two new URL builders copy-paste the entire URL template from the existing `getAppointmentsUrl` — date operator, sort direction, timestamp logic, and `_count`. The only addition is `_getpagesoffset`. This means future changes to the base URL pattern need to be applied in three places.

Could we extend `getAppointmentsUrl` to accept an optional `offset` parameter and implement these as thin wrappers instead?

packages/bahmni-widgets/src/appointments/AppointmentsTable.tsx
```diff
@@ -32,9 +32,7 @@ const AppointmentsTable: React.FC<WidgetProps> = ({ config }) => {
 
   const configuredFields =
     (config?.fields as string[] | undefined) ?? DEFAULT_FIELDS;
-  const numberOfPastAppointments = config?.numberOfPastAppointments as
-    | number
-    | undefined;
+  const configPageSize = Number(config?.pageSize) || 10;
```
ravinderkabli: suggestion: The AC scenarios specify "25 per page" for both upcoming and past appointments, but the implementation defaults to `10`. The JIRA description note says "default is 10", which contradicts the scenario table.

Could you clarify with the PO whether the intended default is 10 or 25? If 25 was the intended default per AC, the fallback should be `|| 25`.

ravinderkabli (commented):
## Review for BAH-4606

Solid server-side pagination implementation with a clean, reusable hook. Found 3 suggestions and 1 nit to discuss before merge.

Found 0 blocking issues, 3 suggestions, and 1 nit.

**Acceptance Criteria**: 3/5 met (⚠️ default pageSize 10 vs AC-stated 25, tab navigation behavior needs PO clarification)

packages/bahmni-services/src/appointmentService/appointmentService.ts
```diff
@@ -104,3 +106,46 @@ export const getAllAppointmentServices = async (): Promise<
 export const deleteAppointmentService = async (uuid: string): Promise<void> => {
   await del(getDeleteAppointmentServiceUrl(uuid));
 };
+
+export interface AppointmentPage {
```
ravinderkabli: nit: `AppointmentPage` is a data shape — the module already has a `models.ts` for exactly this purpose (`AppointmentService`, `Speciality`, `Location`, etc.). Could this be moved there and re-exported?

ravinderkabli (changes requested):

packages/bahmni-widgets/src/appointments/AppointmentsTable.tsx
hamsavarthiniR-Bahmni: deafult is decided as 5 now

packages/bahmni-services/src/appointmentService/appointmentService.ts
hamsavarthiniR-Bahmni: done

packages/bahmni-widgets/src/appointments/AppointmentsTable.tsx
hamsavarthiniR-Bahmni: done

packages/bahmni-services/src/appointmentService/constants.ts
hamsavarthiniR-Bahmni: done

ravinderkabli (approved):
Looks fine

PR #348, opened 2026-04-21
BAH-4565 | Fix pagination footer visibility if totalItems is less or greater than the pageSize

hamsavarthiniR-Bahmni:


ravinderkabli (approved):
Looks good

PR #349, opened 2026-04-21
[BAH-4450] Fix observations horizontal scroll in split view

sivavithyaashree:
JIRA → [BAH-#4450](https://bahmni.atlassian.net/browse/BAH-4450?atlOrigin=eyJpIjoiY2YyMTJkNDQ2M2RjNDZlMGJlNmI2NmYxMzEwYmE5NGQiLCJwIjoiaiJ9)

Description

## Summary

- Adds horizontal scrolling support for the Observations display control in split-panel view, fixing a QA-reported issue where row content (concept name, value, recorded-by) was being truncated
- Applies `min-width` to the rows container inside each date accordion item (`#obs-by-encounter div[id^="accordion-item"] > div`) so the scroll is scoped per-date — the date header stays fixed while only the rows scroll, matching the expected UX
- Adds `padding-block-end` to observation accordion items so the horizontal scrollbar does not overlap the last row
- Approach mirrors the existing `min-width` treatment applied to wide data tables (`cds--data-table:has(td:nth-child(5))`)

## Root cause

The Observations widget renders rows using the `RowCell` component (a flex-based 30/50/20% layout) rather than a `cds--data-table`. Without a `min-width` on the row container, the RowCell adapts to the narrow panel width and the scroll range on `div[id^="accordion-item"]` was only ~16px (the gap overflow) — barely scrollable. Setting `min-width: 62.5rem` on the rows wrapper forces meaningful overflow, giving the accordion item's `overflow-x: auto` a proper scroll range.

## Test plan

- [x] Open a patient with Observations data in split-panel view
- [x] Verify each date accordion item scrolls horizontally and shows all columns (concept name, value, recorded-by)
- [x] Verify the date header (e.g. "04/16/2026") stays fixed while rows scroll
- [x] Verify no section-level scroll is added to the Observations control
- [x] Verify other controls (Allergies, Vitals) are unaffected
- [x] Verify behaviour in full-view (no split panel) is unchanged

Co-authored by: [Komal.sharma@thoughtworks.com](mailto:Komal.sharma@thoughtworks.com)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/apps/clinical/src/components/dashboardSection/styles/DashboardSection.module.scss
+++ b/apps/clinical/src/components/dashboardSection/styles/DashboardSection.module.scss
@@ -5,21 +5,23 @@
 
 .sectionWrapper {
   box-shadow: 0px 2px 6px 0px $shadow;
-  :global(.cds--data-table:has(td:nth-child(5))) td,
-  :global(.cds--data-table:has(td:nth-child(5))) th {
-    min-width: 110px;
-  }
-  :global(div[id^="accordion-item"]) {
-    padding-block-end: $spacing-05;
+  :global(html:has([data-separator])) & {
+    overflow-x: auto;
   }
 }
 
 :global(html:has([data-separator]) div[id^="accordion-item"]) {
   overflow-x: auto;
+  padding-block-end: $spacing-05;
 }
 
-:global(html:has([data-separator]) .cds--data-table:has(td:nth-child(5))) {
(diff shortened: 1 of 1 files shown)


sivavithyaashree:
@ROHIT-V-TW , @ravinderkabli Done with the fix for the QA observation regarding Basic Details control scroll. Please check and approve.

PR #354, opened 2026-04-21
[BAH-4450] Fix observations horizontal scroll in split view

komalsharma225:
Issue- In the 40-60 split view, RowCell-based display controls (Observations, History and
Examinations) were truncating content .

Root Cause: Two CSS issues in CollapsibleRowGroup.module.scss:

.accordion had overflow-x: hidden — blocking the scrollbar from appearing
.rows had width: 100% / max-width: 100% — content always shrank to fit the narrow container
Fix (in CollapsibleRowGroup.module.scss):

.accordion: Override overflow-x: hidden → overflow-x: auto in split mode to allow the
scrollbar
.rows: Set min-width: 62.5rem in split mode to force content wider than the narrow panel,
triggering horizontal scroll
Files Changed: packages/bahmni-design-system/src/molecules/collapsibleRowGroup/styles/Collapsi
bleRowGroup.module.scss


komalsharma225:
Sanity testing is done. Attaching the screenshots of the fix here:

@ROHIT-V-TW @ravinderkabli could you please validate this and approve .

1s

sivavithyaashree:
Hi @ravinderkabli , please review the changes to handle scrollbar inconsistencies in Basic Details Control

ravinderkabli (commented):
## Review for BAH-4450

Targeted CSS fix — the scroll mechanics (`overflow-x: auto` + `min-width` trigger) are correct. Two discussion points on architecture and coverage.

Found 0 blocking issues, 2 suggestions, and 0 nits.

**Acceptance Criteria**: 2/4 fully met, 2 need verification (AC #3 coverage gap, AC #4 global overflow change)

packages/bahmni-design-system/src/molecules/collapsibleRowGroup/styles/CollapsibleRowGroup.module.scss
```diff
@@ -4,6 +4,15 @@
 
 .container {
   width: 100%;
+  :global(html:has([data-separator]) #main-display-area) & {
```
ravinderkabli: suggestion: This selector couples a reusable design-system molecule to two layout-specific implementation details.

`:global(html:has([data-separator]) #main-display-area) &` reaches outside the component to detect:
- `[data-separator]` — an internal attribute of `react-resizable-panels` (not a public API, could change on library upgrade)
- `#main-display-area` — an ID from the template layer (`ActionAreaLayout`)

This is the first SCSS file in the merged codebase to use this pattern. No other design-system molecule reaches outside its scope to detect layout state.

Consider instead:
(a) Adding a `data-split-view` attribute to `ActionAreaLayout` when `isActionAreaVisible` is true — a stable, first-party signal
(b) Moving the scroll rules to the template/widget layer (e.g., `ActionAreaLayout` already manages split-view concerns)
(c) Adding a `scrollable?: boolean` prop to `CollapsibleRowGroup` so consumers opt in

packages/bahmni-design-system/src/molecules/collapsibleRowGroup/styles/CollapsibleRowGroup.module.scss
```diff
@@ -38,6 +47,10 @@
   width: 100%;
   max-width: 100%;
   box-sizing: border-box;
+  :global(html:has([data-separator]) #main-display-area) & {
+    min-width: 40rem;
```
ravinderkabli, line 51: suggestion: AC #3 requires horizontal scroll across *all* display control types — this fix only covers `CollapsibleRowGroup`-based controls (currently Observations via `ObsByEncounter`).

Controls not using `CollapsibleRowGroup`:
- Lab Results (`LabInvestigation` uses Carbon Accordion + SortableDataTable directly)
- Medications "All" tab (Carbon Accordion + SortableDataTable)
- Vitals, Allergies, Diagnoses, Conditions (SortableDataTable directly)

If the merged PR #336/341 already handles `SortableDataTable` controls via `DashboardSection.module.scss`, then this is a targeted complement. Can you verify in the browser that Lab Results and Medications "All" tab have correct horizontal scroll in the 40-60 split view?

sivavithyaashree:
Hi @ravinderkabli, thanks for the detailed review! Addressing both comments:

**CSS coupling (suggestion 1):**
Went with option (a) — added `data-split-view={isActionAreaVisible || undefined}` to the outer `action-area-layout` div in `ActionAreaLayout`. This is a stable, first-party signal we own, so it won't break on a library upgrade.

Replaced all occurrences of `html:has([data-separator])` with `[data-split-view]` across both `CollapsibleRowGroup.module.scss` and `DashboardSection.module.scss` (which also had the same pattern).

**Browser verification (suggestion 2):**
Verified in the browser — Lab Results and Medications "All" tab both have correct horizontal scroll in the 40-60 split view. The scroll rules in `DashboardSection.module.scss` cover the `SortableDataTable`-based controls, so this is a targeted complement as you noted.

ravinderkabli (approved):
Looks fine

ROHIT-V-TW (approved):
Looks good to me, but maybe the css classes we might have to take another look later, approving for now due to time-constraints, will take a tech-debt if required otherwise no changes are needed at all

PR #358, opened 2026-04-22
BAH-4471 | Refactor. Custom Dropdown to Carbon Design System's Combobox and also configurations

ROHIT-V-TW:
This PR is related to BAH-4471 and refactors the existing dropdown with CDS's combobox with no additional functional changes


apps/clinical/src/components/patientSearch/styles/PatientSearch.module.scss
```diff
@@ -18,58 +17,31 @@ $header-actions-offset: 6rem; // 3 action buttons × 2rem each — offset to cle
 
 .searchInput {
   width: 100%;
-}
```
PratitiP, line 20: Instead of overriding the carbon classname, can you try targeting the html element

apps/clinical/src/components/patientSearch/styles/PatientSearch.module.scss
```diff
@@ -18,58 +17,31 @@ $header-actions-offset: 6rem; // 3 action buttons × 2rem each — offset to cle
...
-  &:last-child {
-    border-bottom: none;
-  }
-
-  &:hover,
-  &:focus {
-    background: $layer-hover;
-    outline: none;
-  }
 }
```
PratitiP, line 38: heading-compact-01 have primary text color. No need to override again

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -65,163 +115,77 @@ const PatientSearch: React.FC<PatientSearchProps> = ({ isOpen, onClose }) => {
...
+  // Keyboard handler: Escape closes the panel, Enter submits search
   useEffect(() => {
     if (!isOpen) return;
 
     const handleKeyDown = (event: KeyboardEvent) => {
       if (event.key === 'Escape') {
         onClose();
+        return;
+      }
+      if (event.key === 'Enter' && inputValue.trim() && !submittedTerm) {
```
ravinderkabli: blocking: The `!submittedTerm` guard prevents re-searching after editing the input.

Once a search is submitted (`submittedTerm = "GAN"`), modifying the input to "GAN123" and pressing Enter does nothing — `!submittedTerm` is `false`, and `onInputChange` only clears `submittedTerm` when the input is fully emptied. Users get stuck with no results and no way to re-search without clearing entirely.

Fix the condition to allow re-searching when input differs from the last submitted term:
```ts
if (event.key === 'Enter' && inputValue.trim() && inputValue.trim() !== submittedTerm)
```

Also in `onInputChange`, clear `submittedTerm` when input changes (not just when empty):
```ts
if (input.trim() !== submittedTerm) setSubmittedTerm('');
```

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -65,163 +115,77 @@ const PatientSearch: React.FC<PatientSearchProps> = ({ isOpen, onClose }) => {
...
+      <ComboBox
+        id="patient-search-combobox"
+        data-testid="patient-search-combobox"
         placeholder={t('SEARCH_PATIENT_ID_PLACEHOLDER')}
-        value={inputValue}
-        onChange={handleChange}
-        onKeyDown={handleKeyDown}
-        onClear={handleClear}
-        autoFocus
+        items={showResults ? results : []}
```
ravinderkabli: blocking: The "No matching records" empty state was removed — AC #7 is broken.

When `results` is empty, Carbon's ComboBox renders an empty dropdown with no user feedback. The old code had an explicit "No matching records" div.

Other ComboBox usages in this codebase (MedicationsForm, InvestigationsForm, etc.) inject a disabled sentinel item for empty results — follow that established pattern here.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -65,163 +115,77 @@ const PatientSearch: React.FC<PatientSearchProps> = ({ isOpen, onClose }) => {
...
     <div
       ref={containerRef}
       className={styles.patientSearchContainer}
       data-testid="patient-search-container"
     >
-      <Search
-        id="patient-search-input"
-        testId="patient-search-input"
-        labelText={t('SEARCH_PATIENT_ID_PLACEHOLDER')}
+      <ComboBox
```
ravinderkabli, line 215: suggestion: The `<Loading>` spinner was removed with no replacement — users get no feedback during search.

The old code showed `<Loading description={t('SEARCHING')} />` while loading. Now `showResults` requires `!isLoading`, so while loading the ComboBox shows an empty items list and the UI appears to do nothing after pressing Enter.

Consider re-adding a loading state — e.g., a conditional spinner when `isLoading && submittedTerm`, or Carbon's `helperText` prop to show "Searching..." while loading.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -65,163 +115,77 @@ const PatientSearch: React.FC<PatientSearchProps> = ({ isOpen, onClose }) => {
...
+          setInputValue(input);
+          if (!input.trim()) {
+            setSubmittedTerm('');
+          }
+        }}
+        selectedItem={selectedPatient}
+        clearSelectedOnChange
+        shouldFilterItem={() => true}
+        autoAlign
+        aria-label={t('SEARCH_PATIENT_ID_PLACEHOLDER')}
```
ravinderkabli, line 237: suggestion: `autoFocus` was dropped — keyboard users lose focus-on-open.

The old `<Search>` had `autoFocus` so the input received focus immediately when the panel opened. Carbon ComboBox supports this via:
```tsx
inputProps={{ autoFocus: true }}
```

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -65,163 +115,77 @@ const PatientSearch: React.FC<PatientSearchProps> = ({ isOpen, onClose }) => {
...
         onClose();
+        return;
+      }
+      if (event.key === 'Enter' && inputValue.trim() && !submittedTerm) {
+        setSubmittedTerm(inputValue.trim());
       }
     };
 
     document.addEventListener('keydown', handleKeyDown);
     return () => {
```
ravinderkabli: suggestion: The global `document` keydown listener can conflict with Carbon's internal Enter handling.

Carbon's ComboBox handles Enter via downshift internally (selecting highlighted items). Since Carbon doesn't call `stopPropagation()`, both the internal handler and this global listener fire for the same Enter keypress. When results are showing and a user presses Enter on a highlighted item, both navigation (via Carbon's `onChange`) and `setSubmittedTerm` (via this handler) can fire.

Consider attaching the listener to `containerRef.current` instead of `document` to scope it more tightly.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -65,163 +115,77 @@ const PatientSearch: React.FC<PatientSearchProps> = ({ isOpen, onClose }) => {
...
   if (!isOpen) return null;
 
-  const handleResultClick = (resultUuid: string) => {
-    navigate(`../${resultUuid}`);
+  const handleChange = (
+    selectedItem: PatientSearchResult | null | undefined,
+  ) => {
+    if (!selectedItem?.uuid) return;
+    navigate(`../${selectedItem.uuid}`);
+    setSelectedPatient(selectedItem);
```
ravinderkabli: nit: `selectedPatient` state serves no functional purpose here.

`setSelectedPatient(selectedItem)` is called then immediately followed by `navigate()` and `onClose()`. The close triggers the reset effect which sets it back to `null`. The state is set and immediately discarded. `clearSelectedOnChange` also never gets a chance to fire.

Consider removing `selectedPatient`, `setSelectedPatient`, and `clearSelectedOnChange` entirely.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -65,163 +115,77 @@ const PatientSearch: React.FC<PatientSearchProps> = ({ isOpen, onClose }) => {
     };
   }, [isOpen, onClose]);
 
-  // Keyboard handler: Escape closes the panel
+  // Keyboard handler: Escape closes the panel, Enter submits search
   useEffect(() => {
     if (!isOpen) return;
 
     const handleKeyDown = (event: KeyboardEvent) => {
       if (event.key === 'Escape') {
```
ravinderkabli, line 154: nit: One Escape press closes the entire search panel, not just the dropdown.

Carbon's ComboBox closes its dropdown on Escape but doesn't call `stopPropagation()`, so the event bubbles to the document listener which calls `onClose()`. Standard combobox UX expects Escape to close the dropdown first, then a second Escape to close the panel. This may be intentional — just flagging.

apps/clinical/src/providers/clinicalConfig/schema.json
```diff
@@ -38,6 +38,32 @@
...
+      "additionalProperties": false,
+      "properties": {
+        "displayFields": {
+          "type": "array",
+          "description": "Fields to display in patient search dropdown results. Each entry maps to a PatientSearchResult field.",
+          "items": {
+            "type": "object",
+            "required": ["field"],
+            "additionalProperties": false,
+            "properties": {
```
ravinderkabli, line 114: nit: `"type": "string"` allows any value, but `getFieldValue` only handles `name`, `identifier`, `gender`, and `age`.

Consider constraining to `"enum": ["name", "identifier", "gender", "age"]` so misconfiguration is caught at validation time. Also narrow the TypeScript type in `PatientSearchDisplayField` to a union type.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
```diff
@@ -65,163 +115,77 @@ const PatientSearch: React.FC<PatientSearchProps> = ({ isOpen, onClose }) => {
...
+        onChange={({ selectedItem }) => handleChange(selectedItem)}
+        onInputChange={(input) => {
+          setInputValue(input);
+          if (!input.trim()) {
+            setSubmittedTerm('');
+          }
+        }}
+        selectedItem={selectedPatient}
+        clearSelectedOnChange
+        shouldFilterItem={() => true}
```
ravinderkabli: question: `shouldFilterItem={() => true}` is correct for server-side search (passes all items through, disabling client-side filtering). A brief inline comment would signal intent to future readers:
```tsx
shouldFilterItem={() => true} // Server-side search; disable client-side filtering
```

ravinderkabli (changes requested):
## Review for BAH-4471

Good refactor direction — moving to Carbon ComboBox is the right call. Found a few behavioral regressions to address before merge.

Found 2 blocking issues, 4 suggestions, and 4 nits.

**Acceptance Criteria**: 3/9 fully met (AC #7 missing, several others at risk due to behavioral regressions)

### Issues not in diff

| File:Line | Type | Issue |
|-----------|------|-------|
| PatientSearch.test.tsx | suggestion | No test for zero-results scenario — old "No matching records" test deleted with no replacement |
| PatientSearch.test.tsx | nit | Tab-reachability test deleted, losing explicit AC #1 coverage |

apps/clinical/src/components/patientSearch/PatientSearch.tsx
ROHIT-V-TW: Fixed. Changed the Enter condition to `inputValue.trim() !== submittedTerm` so editing the search term and pressing Enter triggers a new search. Also updated `onInputChange` to clear `submittedTerm` whenever input diverges from the last submitted term (not just when empty).

apps/clinical/src/components/patientSearch/PatientSearch.tsx
ROHIT-V-TW: Fixed. Added disabled sentinel items following the established pattern from MedicationsForm/InvestigationsForm/AllergiesForm. The `comboBoxItems` useMemo now returns:
- `t('SEARCHING')` disabled item while loading
- Error message disabled item on failure
- `t('NO_MATCHING_RECORDS')` disabled item when results are empty

AC #7 should be covered now.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
ROHIT-V-TW: Fixed. Addressed together with the empty state fix — the `comboBoxItems` useMemo returns a disabled `t('SEARCHING')` sentinel item while `isLoading` is true, so users see feedback in the dropdown during search.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
ROHIT-V-TW: Fixed. Added a `useEffect` that queries the input element inside the container and calls `.focus()` when `isOpen` becomes true. Carbon ComboBox doesn't have a direct `autoFocus` prop, so this approach is more reliable.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
ROHIT-V-TW: Fixed. Moved the keydown listener from `document` to `containerRef.current` so it's scoped to the search container and won't conflict with Carbon's internal Enter/downshift handling.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
ROHIT-V-TW: Fixed. Removed `selectedPatient`, `setSelectedPatient`, and `clearSelectedOnChange` entirely. `selectedItem` is now always `null`.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
ROHIT-V-TW: With the keydown listener now scoped to `containerRef.current` instead of `document`, Escape only fires when focus is within the search container. Since the container is just a combobox (no separate panel body), single-Escape-to-close feels like the right UX here — there's no intermediate dropdown-only state worth preserving.

apps/clinical/src/providers/clinicalConfig/schema.json
ROHIT-V-TW: Fixed. Added `"enum": ["name", "identifier", "gender", "age"]` to the JSON schema and narrowed the TypeScript type in `PatientSearchDisplayField` to `'name' | 'identifier' | 'gender' | 'age'`.

apps/clinical/src/components/patientSearch/PatientSearch.tsx
ROHIT-V-TW: Fixed. Added inline comment: `// Server-side search; disable client-side filtering`

apps/clinical/src/components/patientSearch/styles/PatientSearch.module.scss
ROHIT-V-TW: Hey Pratiti, thanks for bringing this up
Replaced :global(.cds--list-box__menu-item) and :global(.cds--list-box__menu-item__option) with li[role="option"] and > div.

Can u validate this once @PratitiP

apps/clinical/src/components/patientSearch/styles/PatientSearch.module.scss
ROHIT-V-TW: Got it, thanks Pratiti I've removed the override

ravinderkabli (approved):
Looks fine for the comments I gave.

PratitiP (approved):

PR #359, opened 2026-04-23
BAH-4519 & BAH-4638| Homepage migration from Angular to React

arshiyaTW2021:
## Migrate Bahmni Home Landing Page to React

**Jira:** [BAH-4519](https://bahmni.atlassian.net/browse/BAH-4519) | [BAH-4638](https://bahmni.atlassian.net/browse/BAH-4638)

### Summary

Replaces the legacy Angular home page with a React implementation aligned to the [Figma redesign](https://www.figma.com/design/w31Zjs4NHPCZSGs1xyBzF9/Home-Page-Redesign?node-id=0-1&p=f&m=dev). Users now land on a Carbon Design System-based home page after login, with full functional parity to the Angular version plus visual and accessibility improvements.

### What's included

**Home page shell (BAH-4519)**
- `HomePageHeader` — top bar, location selector dropdown, and user profile menu (overflow menu with Change Password / Logout)
- `LocationSelector` — dropdown of available login locations 
- `LocationProvider` / `LocationContext` — context-based location state shared across home page components
- `UserProfileMenu` — displays user greeting, handles logout  and change-password navigation
- Logout added to `userService` 
- `getAvailableLocations`, `getDefaultDateFormat`, `saveUserLocation` added to `userService`

**Module tiles grid (BAH-4638)**
- `AppTile` — clickable tile with label, icon, and arrow; privilege-gated via `useHasPrivilege` (tiles are hidden, not greyed out, when the user lacks the required privilege)
- `HomePageGrid` — fetches module config from `extension.json`, filters by extension point / privilege / online status, sorts by order field, renders tiles in a responsive Carbon Grid
- `moduleService` — fetches, filters, and sorts module extensions from `/bahmni_config/openmrs/apps/home/v2/extension.json`
- Cross-app navigation: SPA routes use React Router, legacy routes use `window.location`

**i18n**
- Home page namespace (`home`) with English and Spanish locale files under `distro/public/home/locales/` and `distro/public/locales/`
- i18n test infrastructure (`setupTests.i18n.ts`) for distro
- `initAppI18n` enhanced to preserve namespaces across module navigation

### Related changes

- **standard-config**: [`BAH-4638` branch](https://github.com/Bahmni/standard-config/tree/BAH-4638) — home page module configuration (`extension.json`) with tile definitions, icons, and ordering

[...]

PR #363, opened 2026-04-28
BAH-4519 | Fix. Update OpenMRS server-side session on location change + user profile icon size    

arshiyaTW2021:
**Description:**                                                                                                                    
   
  **1. Fix: Update OpenMRS server-side session on location change**                                                                   
                                                                                                                                
  When a user changes location from the home page header, the new Bahmni was only updating the client-side cookie and persisting  the preference to user properties but never updating the OpenMRS server-side session. This meant any encounters or  observations created after a location change would be attributed to the wrong location.                                         
                                                                                                                                
  - Added **updateSessionLocation(locationUuid)** to bahmni-services which posts { sessionLocation: uuid } to                         
  **/openmrs/ws/rest/v1/session** the same call the old Bahmni (openmrs-module-bahmniapps) made on location change
  - Wired it up in LocationProvider so it fires alongside the existing cookie update and saveUserLocation call on every location  change                                                                                                                          
  - Updated test mocks in LocationContext.test.tsx and LocationProvider.test.tsx to include the new function
                                                                                                                                  
  2. **Fix: User profile icon size in header**                                                                                      

  Changed the UserProfileMenu overflow menu icon size from sm to lg to correctly match the intended header design.


ravinderkabli:
## Code Review: PR #359 — BAH-4519 & BAH-4638 | Homepage migration from Angular to React

> PR #359 was merged to unblock QA. Posting review findings here for follow-up tracking.
> **PR #359**: https://github.com/Bahmni/bahmni-apps-frontend/pull/359
> **Reviewed**: 2026-04-28

---

### Summary
Several issues need follow-up.

| Severity | Count |
|----------|-------|
| 🔴 CRITICAL | 1 |
| 🟠 MAJOR | 9 |
| 🟡 MINOR | 9 |
| 🔵 INFO | 3 |

---

## 🔴 CRITICAL

### 1. `getAvailableLocations` silently swallows errors — caller can't detect failure

**File**: `packages/bahmni-services/src/userService/userService.ts` L85–94

`getAvailableLocations` catches all errors, logs them, and returns `[]`. The caller (`LocationProvider.initializeLocation`) has no way to distinguish a successful "no locations configured" response from a network failure. If the locations API fails, the user sees an empty dropdown with no error indication — the `catch` block in `initializeLocation` never fires because the error was swallowed upstream.

```typescript
export const getAvailableLocations = async (): Promise<UserLocation[]> => {
  try {
    const response = await get<LocationsResponse>(AVAILABLE_LOCATIONS_URL);
    return response.results ?? [];
  } catch (error) {
    console.error('Failed to fetch available locations:', error);
    return [];  // ← error swallowed; caller receives [] silently
  }
};
```

Contrast with `getCurrentUser` and `logout` which both propagate errors to their callers. Location selection is a core workflow — clinicians must have a valid location set.

**Fix**: Remove the try/catch (let the error propagate):
```typescript
export const getAvailableLocations = async (): Promise<UserLocation[]> => {
  const response = await get<LocationsResponse>(AVAILABLE_LOCATIONS_URL);
  return response.results ?? [];
};
```

[Code link](https://github.com/Bahmni/bahmni-apps-frontend/blob/fddb766993a949dd5eeabc03db8d60bae643d268/packages/bahmni-services/src/userService/userService.ts#L85-L94)

---

## 🟠 MAJOR

### 2. Missing translation keys `LOADING` and `NO_LOCATION_SELECTED` — raw key strings shown in UI

**File**: `distro/src/app/components/LocationSelector/LocationSelector.tsx` L31, L43

`LocationSelector` uses `t('LOADING')` and `t('NO_LOCATION_SELECTED')`, neither of which exists in any locale file. `UserProfileMenu` also uses `t('LOADING')`. Users see the raw string `LOADING` or `NO_LOCATION_SELECTED`. The AC for both tickets explicitly requires all text to be displayed in the configured language.

**Fix**: Add `HOME_LOADING` and `HOME_NO_LOCATION_SELECTED` to `distro/public/locales/locale_en.json` and `locale_es.json`, update component calls, and update tests to assert the translated string.

[Code link](https://github.com/Bahmni/bahmni-apps-frontend/blob/fddb766993a949dd5eeabc03db8d60bae643d268/distro/src/app/components/LocationSelector/LocationSelector.tsx#L31-L46)

---

### 3. `moduleService.ts` placed in `distro/src/services/` instead of `packages/bahmni-services/`

**File**: `distro/src/services/moduleService.ts`

All API/data services live in `packages/bahmni-services/src/`. Placing `moduleService.ts` at the distro level means it cannot be reused by sub-apps and breaks the established monorepo pattern the JIRA tech analysis describes.

**Fix**: Move to `packages/bahmni-services/src/moduleService/moduleService.ts` following the pattern of existing service packages. Export from `packages/bahmni-services/src/index.ts`.

[Code link](https://github.com/Bahmni/bahmni-apps-frontend/blob/fddb766993a949dd5eeabc03db8d60bae643d268/distro/src/services/moduleService.ts)

---

### 4. `HomePageGrid` uses `useState/useEffect` instead of TanStack Query

**File**: `distro/src/app/components/HomePageGrid/HomePageGrid.tsx` L5–35

Every other app in this monorepo that fetches data uses TanStack Query. The JIRA tech analysis explicitly lists "TanStack Query - caching pattern" as reusable infrastructure. The manual pattern provides no caching, no background refetch, no stale-while-revalidate, and no request deduplication.

**Fix**:
```typescript
const { data: modules = [], isLoading, isError } = useQuery({
  queryKey: ['home-modules'],
  queryFn: () => getVisibleModules('org.bahmni.home.dashboard'),
});
```

[Code link](https://github.com/Bahmni/bahmni-apps-frontend/blob/fddb766993a949dd5eeabc03db8d60bae643d268/distro/src/app/components/HomePageGrid/HomePageGrid.tsx#L5-L35)

---

### 5. `additionalProperties: true` in appointments and registration schemas — no documented reason

**Files**: `apps/appointments/src/providers/appointmentsConfig/schema.json` and `apps/registration/src/providers/registrationConfig/schema.json` L6

`additionalProperties: false` is a deliberate strictness setting. Both schemas were changed to `true` in this PR with no documented reason, silently allowing arbitrary extra keys to pass validation in two apps with no home-page functionality.

**Fix**: Revert both to `"additionalProperties": false`. If a new config property is genuinely needed, add it explicitly to the `properties` block.

---

### 6. `initAppI18n` promise has no `.catch()` — silent blank page on i18n init failure

**File**: `distro/src/main.tsx` L15–23

If `initAppI18n('home')` rejects, `createRoot().render()` is never called and the user sees a permanent blank page. Sub-apps wrap their i18n init in try/catch and render regardless — the distro shell lacks this resilience.

**Fix**:
```typescript
initAppI18n('home')
  .catch((err) => console.error('i18n init failed, rendering with fallback strings:', err))
  .finally(() => {
    const root = createRoot(document.getElementById('root') as HTMLElement);
    root.render(<StrictMode><BrowserRouter basename={PUBLIC_PATH ?? '/'}><App /></BrowserRouter></StrictMode>);
  });
```

[Code link](https://github.com/Bahmni/bahmni-apps-frontend/blob/fddb766993a949dd5eeabc03db8d60bae643d268/distro/src/main.tsx#L15-L23)

---

### 7. Missing `.noLocation` CSS class in `LocationSelector.module.scss`

**File**: `distro/src/app/components/LocationSelector/styles/LocationSelector.module.scss`

`LocationSelector.tsx` references `styles.noLocation` but the SCSS only defines `.dropdownWrapper`, `.loading`, and `.error`. With CSS Modules, `styles.noLocation` resolves to `undefined`, producing `className="undefined"` — the no-location state renders unstyled.

**Fix**: Add `.noLocation` to the SCSS file:
```scss
.noLocation {
  @include type-style("label-01");
  padding: $spacing-03 $spacing-04;
  color: var(--cds-text-secondary);
}
```

[Code link](https://github.com/Bahmni/bahmni-apps-frontend/blob/fddb766993a949dd5eeabc03db8d60bae643d268/distro/src/app/components/LocationSelector/styles/LocationSelector.module.scss)

---

### 8. Tests assert raw i18n keys — codifies broken behaviour

**File**: `distro/src/app/components/LocationSelector/__tests__/LocationSelector.test.tsx` L55–59, L71–76

Two tests assert `'LOADING'` and `'NO_LOCATION_SELECTED'` as expected text. These tests pass only because the locale keys don't exist — they perpetuate the bug rather than guard against it.

**Fix**: After adding locale keys (issue #2), update tests to assert the translated English string.

[Code link](https://github.com/Bahmni/bahmni-apps-frontend/blob/fddb766993a949dd5eeabc03db8d60bae643d268/distro/src/app/components/LocationSelector/__tests__/LocationSelector.test.tsx#L55-L76)

---

### 9. `UserLocation` type duplicated — divergent shapes with `display?` field

**File**: `distro/src/app/context/LocationContext.tsx` L3–6

`LocationContext.tsx` defines its own `UserLocation = { name: string; uuid: string }`. The canonical definition in `packages/bahmni-services/src/userService/models.ts` also has `display?: string`. TypeScript won't surface the mismatch and the `display` field used in the dropdown could be inadvertently dropped.

**Fix**: Remove the local definition and import `UserLocation` from `@bahmni/services`.

[Code link](https://github.com/Bahmni/bahmni-apps-frontend/blob/fddb766993a949dd5eeabc03db8d60bae643d268/distro/src/app/context/LocationContext.tsx#L3-L6)

---

### 10. `BAHMNI_USER_LOCATION_COOKIE` constant redefined locally — divergence risk

**File**: `distro/src/app/context/LocationProvider.tsx` L13

`packages/bahmni-services/src/userService/constants.ts` already exports this constant. `LocationProvider.tsx` redefines it locally with the same string — silent mismatch risk if the canonical value changes.

**Fix**: Export from the package's `index.ts` and import in `LocationProvider.tsx` from `@bahmni/services`.

[Code link](https://github.com/Bahmni/bahmni-apps-frontend/blob/fddb766993a949dd5eeabc03db8d60bae643d268/distro/src/app/context/LocationProvider.tsx#L13)

---

## 🟡 MINOR

### 11. `LocationProvider` at app root makes unnecessary API calls for non-home routes
**File**: `distro/src/app/app.tsx` L31–44 — wraps ALL routes, firing API calls on every app load regardless of route. Move to wrap only `<IndexPage />`.

### 12. `handleSetLocation` not wrapped in `useCallback`
**File**: `distro/src/app/context/LocationProvider.tsx` L55–97 — plain function recreated each render. Wrap in `useCallback([location, userUuid])`.

### 13. `isLoggingOut` not reset on successful logout — asymmetric state handling
**File**: `distro/src/app/components/UserProfileMenu/UserProfileMenu.tsx` L26–40 — error path resets `isLoggingOut`, success path does not. Use `try/finally`.

### 14. `await setLocation()` on a synchronous function — dead `catch` block
**File**: `distro/src/app/components/LocationSelector/LocationSelector.tsx` L13–27 — `handleSetLocation` is synchronous so the `catch` in `handleLocationChange` is dead code. Remove `async/await` or document intent.

### 15. `t` missing from `useCallback` deps in `loadModules`
**File**: `distro/src/app/components/HomePageGrid/HomePageGrid.tsx` L10–33 — `t` is used inside `useCallback([])` but missing from deps. Add `t` to deps.

### 16. Hardcoded Angular login/change-password URLs
**File**: `distro/src/app/components/UserProfileMenu/UserProfileMenu.tsx` L30, L57 — import from `@bahmni/services` constants instead of inlining strings.

### 17. `distro/public/home/locales/` (21 keys) appears unused — orphaned files
Not referenced in webpack config, 5 fewer keys than the bundled file. Remove or wire up via webpack config.

### 18. `Hi, {user.display}` greeting string not localized
**File**: `distro/src/app/components/UserProfileMenu/UserProfileMenu.tsx` L~67 — the `Hi, ` prefix is hardcoded English. Use `t('HOME_GREETING', { name: user.display })`.

### 19. `expect(baseElement).toBeTruthy()` — vacuous test assertion
**File**: `distro/src/app/__tests__/app.test.tsx` L24–29 — `baseElement` is always truthy. Replace with an assertion on actual rendered content.

---

## 🔵 INFO

### 20. `ICON_SIZE.X2` not in AppTile test mock — silently uses default
Mock has `ICON_SIZE: { LG: 'lg' }` but `AppTile` uses `ICON_SIZE.X2` — resolves to `undefined` in tests. Not a bug (Icon is mocked) but the mock is inaccurate.

### 21. Webpack dev proxy changed to catch-all function
Changed from `context: ['/bahmni_config', '/openmrs']` to a function proxying all paths not under `publicPath`. Verify this doesn't break other local dev proxy expectations.

### 22. `getVisibleModules` could pass user privileges for defense-in-depth
`HomePageGrid` never passes `userPrivileges` to `getVisibleModules`, relying entirely on `AppTile.useHasPrivilege`. Functionally correct but single-layered.

---

## Acceptance Criteria Check (BAH-4519 & BAH-4638)

| Criterion | Status | Notes |
|-----------|--------|-------|
| Recognisable home page with header and content area on login | ✅ | |
| Display name shown in header | ✅ | |
| Login location visible in header | ✅ | |
| Can choose from available locations, selection reflected in session | ✅ | |
| Logout ends session → login page | ✅ | |
| Change password → change password page | ✅ | |
| Keyboard navigation (location + profile icon) | ✅ | |
| User informed on logout error, session intact | ✅ | |
| All text in configured language | ⚠️ | `t('LOADING')` and `t('NO_LOCATION_SELECTED')` show raw keys; `Hi, ` prefix not localized |
| Fallback to English when locale not supported | ✅ | |
| Module grid: tile with name, icon, visual indicator; clicking navigates | ✅ | |
| Keyboard accessible tiles | ✅ | |
| Only configured modules shown, in implementer order | ✅ | |
| Clear error message when config can't be loaded | ✅ | |
| Tile visible/absent based on privilege | ✅ | |

---

## What's Done Well

1. **Comprehensive test coverage** — every component has unit tests including accessibility (jest-axe), keyboard navigation, error states, loading states, and privilege scenarios.
2. **Accessibility first** — `role="button"`, `tabIndex={0}`, `aria-label`, `aria-hidden` on decorative elements, and passing axe scans.
3. **Multi-namespace i18n** — the `initI18n` change (adding `isInitialized` guard) correctly solves the multi-app namespace problem.
4. **Carbon Design System usage** — consistent use of Carbon tokens and components throughout.
5. **Clean error recovery in `LocationProvider`** — cookie write failures correctly rollback location state and surface the error.

arshiyaTW2021:
Addressing comments from previous merged PR https://github.com/Bahmni/bahmni-apps-frontend/pull/359 and the current PR #363

1. **getAvailableLocations silently swallows errors — returns [] on API failure with no indication to caller.**

Fix: Removed try/catch from packages/bahmni-services/src/userService/userService.ts. Updated userService.test.ts to assert rejects.toThrow instead of checking for empty array return.

2. **t('LOADING') and t('NO_LOCATION_SELECTED') keys missing from all locale files**

Fix: Added HOME_LOADING, HOME_NO_LOCATION_SELECTED, and HOME_GREETING to distro/public/locales/locale_en.json and locale_es.json. Updated LocationSelector.tsx  and UserProfileMenu.tsx  to call t('HOME_LOADING') and t('HOME_NO_LOCATION_SELECTED').

3. **Tests assert 'LOADING' and 'NO_LOCATION_SELECTED' as expected text - they passed only because the locale keys didn't exist, perpetuating the bug.**

Fix: Updated LocationSelector.test.tsx 'Loading' and 'No location selected'. Updated UserProfileMenu.test.tsx 'Loading'. Removed the handles error from setLocation gracefully test (see item 14).

4. **moduleService.ts placed in distro/src/services/ instead of packages/bahmni-services/**

Moved to packages/bahmni-services/src/moduleService/moduleService.ts. 

5. **HomePageGrid uses useState/useEffect instead of TanStack Query.**

Fix: Replaced manual state in HomePageGrid.tsx with TanstackQuery

6. **additionalProperties: true at top level of apps/appointments and apps/registration config schemas**

Fix: Changed top-level "additionalProperties": true -> "additionalProperties": false in both schema.json files

7. **initAppI18n('home') promise has no .catch() - if i18n init fails the network call rejects, createRoot().render() is never called, and the user sees a permanent blank page.**

Replaced .then(...) with .catch(err => console.error(...)).finally(() => createRoot(...).render(...)) in distro/src/main.tsx..finally() guarantees the app always renders regardless of i18n failure. Users see components with fallback key strings rather than a blank page. Consistent with how sub-apps handle i18n init failures.

8. **LocationSelector.tsx references styles.noLocation but the SCSS file has no .noLocation class**

Fix: Added .noLocation file

9. **UserLocation type defined locally in LocationContext.tsx as { name: string; uuid: string } — missing display?: string from the canonical definition in packages/bahmni-services. TypeScript won't surface the mismatch.**

Removed local interface UserLocation from LocationContext.tsx. Added import { type UserLocation } from '@bahmni/services' and export type { UserLocation } so downstream files importing from ./LocationContext continue to work. Updated context/index.ts to re-export UserLocation from @bahmni/services directly. Added type UserLocation to the userService export block in packages/bahmni-services/src/index.ts.

10. **BAHMNI_USER_LOCATION_COOKIE = 'bahmni.user.location' redefined locally in LocationProvider.tsx — already exists in packages/bahmni-services/src/userService/constants.ts but not exported from the package index. Silent divergence risk if the canonical value changes.**


Exported BAHMNI_USER_LOCATION_COOKIE through packages/bahmni-services/src/userService/index.ts → packages/bahmni-services/src/index.ts. Removed local re-definition in LocationProvider.tsx; imported from @bahmni/services.

11. **LocationProvider wraps ALL routes in app.tsx (clinical, registration, appointments) - fires getAvailableLocations and getCurrentUser on every app load regardless of route.**

Removed <LocationProvider> from app.tsx. Added it to IndexPage.tsx wrapping only the home page content.

12. **handleSetLocation recreated on every render - not wrapped in useCallback, causing unnecessary re-renders of consumers via useMemo.**

Wrapped handleSetLocation in useCallback([location, userUuid]) in LocationProvider.tsx.

13. **isLoggingOut reset to false only in the error path of handleLogout - success path navigates away without resetting, leaving the button permanently disabled if navigation is prevented.**

Moved setIsLoggingOut(false) into a finally block in UserProfileMenu.tsx. Removed the setIsLoggingOut(false) call from the catch block.

14. **handleLocationChange is async and awaits setLocation which is a synchronous function — the catch block was dead code since handleSetLocation catches all errors internally and never throws to its caller.**

Made handleLocationChange synchronous in LocationSelector.tsx. Removed async/await, useState(isUpdating), and the catch block. Removed disabled={isUpdating} from the Dropdown. Removed the test case handles error from setLocation gracefully (it was testing the now-removed dead code path).

15. **t missing from useCallback deps array in loadModules in HomePageGrid — ESLint exhaustive-deps violation.**

Resolved by converting to TanStack Query (item 5) — useCallback and useEffect were removed entirely.

16. **Login and change-password URLs hardcoded as strings in UserProfileMenu.tsx — LOGIN_PATH already exists in packages/bahmni-services/src/api/constants.ts but was not exported.**

Added CHANGE_PASSWORD_PATH constant to packages/bahmni-services/src/api/constants.ts. Exported both LOGIN_PATH and CHANGE_PASSWORD_PATH from api/index.ts and packages/bahmni-services/src/index.ts. Replaced hardcoded strings in UserProfileMenu.tsx with the imported constants.

17. **distro/public/home/locales/ (with locale_en.json and locale_es.json) is orphaned - not referenced by webpack config, has 5 fewer keys than the active locale file at distro/public/locales/.**

Deleted distro/public/home/ directory entirely.

18. **Hi, {user.display} greeting is hardcoded English**

Added "HOME_GREETING": "Hi, {{name}}" (EN) and "HOME_GREETING": "Hola, {{name}}" (ES) to both locale files. Updated UserProfileMenu.tsx to use {t('HOME_GREETING', { name: user.display })}.

19. **expect(baseElement).toBeTruthy() in app.test.tsx -vacuous assertion; baseElement is always truthy, the test would pass even if the app threw during render.**

Replaced with expect(screen.getByRole('main')).toBeInTheDocument(). Added getVisibleModules: jest.fn().mockResolvedValue([]) to the mock (required after HomePageGrid now calls it via TanStack Query).

20. **ICON_SIZE.X2 not in AppTile test mock - ICON_SIZE: { LG: 'lg' } missing X2 which AppTile uses; resolves to undefined in tests.**

Fixed. Added X2: '2x' to the mock in AppTile.test.tsx.

apps/appointments/src/providers/appointmentsConfig/schema.json
```diff
@@ -3,7 +3,7 @@
   "title": "Appointments App Configuration Schema",
   "description": "Schema for global appointments app configuration",
   "type": "object",
-  "additionalProperties": true,
+  "additionalProperties": false,
```
sumaztwcode, line 6: nit: Changing `additionalProperties` from `true` to `false` means any Bahmni deployment with custom config properties not listed in the schema will fail validation. Worth confirming no existing deployments rely on extra properties here.

(Same applies to `registration/schema.json`.)

sumaztwcode (approved):
## Review for BAH-4519

Well-structured PR delivering the React migration of the Bahmni home page. Clean architecture — shared services extracted to `@bahmni/services`, `useQuery` migration, i18n with fallback resilience, and properly scoped providers. Tests thoroughly updated. No blocking issues.

Found 0 blocking issues, 0 suggestions, and 1 nit.

**Acceptance Criteria**: 10/10 met

### What's done well

1. **Clean service extraction** — `moduleService`, `updateSessionLocation`, `UserLocation` type moved to `@bahmni/services`
2. **Proper useQuery migration** — correct `enabled` guard, `queryKey` with privilege names
3. **i18n resilience** — `.catch().finally()` ensures app renders even if locale files fail
4. **Optimistic updates** — location changes update UI immediately with non-blocking warnings on failure
5. **Test quality** — `userEvent.setup()`, `jest-axe` a11y tests, proper `QueryClientProvider` wrapping

apps/appointments/src/providers/appointmentsConfig/schema.json
arshiyaTW2021: additionalProperties at top level of apps/appointments and apps/registration config schemas were turned to true for the same reason that any Bahmni deployment with custom config properties not listed in the schema will fail validation however in the following PR review https://github.com/Bahmni/bahmni-apps-frontend/pull/359,   it was suggested that this silently allows misconfigured keys through validation. Revert to false; if a new property is needed, add it explicitly to the properties block.

PR #364, opened 2026-04-28
BAH-4550 | Add. Dose Sequence Field And Config-Driven Columns In Immunization History

rahu1ramesh:
JIRA → [BAH-4550](https://bahmni.atlassian.net/browse/BAH-4550)

## Description

Extends the Immunization History feature with two enhancements: support for config-driven column visibility and expanded fields in the Administered/Not-Administered tabs, and a new optional Dose Sequence field in the immunization entry form that maps to the FHIR `protocolApplied.doseNumberPositiveInt` field on submission.

[BAH-4550]: https://bahmni.atlassian.net/browse/BAH-4550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/clinical/src/components/forms/immunizationHistory/components/SelectedImmunizationItem.tsx
```diff
@@ -322,6 +324,30 @@ const SelectedImmunizationItem: React.FC<SelectedImmunizationItemProps> = ({
...
+            <NumberInput
+              id={`immunization-dose-sequence-${id}`}
+              data-testid={`immunization-dose-sequence-${id}`}
+              label={t('IMMUNIZATION_HISTORY_DOSE_SEQUENCE')}
+              placeholder={t('IMMUNIZATION_HISTORY_DOSE_SEQUENCE_PLACEHOLDER')}
+              value={immunization.doseSequence ?? 0}
+              onChange={(_e, { value }) =>
+                updateDoseSequence(id, Number(value))
+              }
+              min={0}
```
mohan-13, line 338: min can be 1, i think the store validates it like that.

also why put default value, i guess carbon number input can be configured to not accept any input as well. pleaser check.

packages/bahmni-widgets/src/immunizationHistory/ImmunizationHistory.tsx
```diff
@@ -39,6 +45,26 @@ const ImmunizationHistory: React.FC<WidgetProps> = ({ config }) => {
   const encounterType = config?.encounterType as string;
   const startEncounterPrivilege = config?.startEncounterPrivilege as string;
 
+  const completedFields = config?.completedFields as string[] | undefined;
```
mohan-13: Just a thought on naming. can we name it administeredFields and notAdministeredFields which will be more implementation friendly, and goes well with the tabs.

packages/bahmni-widgets/src/immunizationHistory/__tests__/ImmunizationHistory.test.tsx
```diff
@@ -210,6 +218,50 @@ describe('ImmunizationHistory', () => {
     });
   });
 
+  it.each([
```
mohan-13, line 221: Seems a missing assertion for expanded fields ?

apps/clinical/src/components/forms/immunizationHistory/components/SelectedImmunizationItem.tsx
rahu1ramesh: Min if we set to one, will default to one always. 0 hear is considered null. `DoseSequence` is a positive integer always, zero here mean unset.

mohan-13: Please check the checks in store and fhir creator, it seems some places we are checking for undefined and zero.

Carbon supports `allowEmpty` which can be set, so that we dont need a default value and it will be empty unless user takes an action

packages/bahmni-widgets/src/immunizationHistory/__tests__/ImmunizationHistory.test.tsx
rahu1ramesh: I have added one for the expanded row as well, good catch.

apps/clinical/src/components/forms/immunizationHistory/components/SelectedImmunizationItem.tsx
rahu1ramesh: `allowEmpty` can be used, but from a user interaction perspective 0 is less confusing, to me min 1 feels like 1 is selected always.

mohan-13 (approved):

PR #365, opened 2026-04-28
BAH-4638| Fix. Tile rendering when there's limited privilege access.

arshiyaTW2021:


ravinderkabli (commented):
## Review for BAH-4638

Good progress on lifting privilege filtering from `AppTile` to the grid level — the architectural direction is sound, and the test coverage is thorough. Found a race condition with the provider's initial state that needs fixing.

Found 1 blocking issue, 2 suggestions, and 2 nits.

**Acceptance Criteria**: 4/4 in-scope ACs met (7, 8, 9, 6), but the race condition (issue #1) can show a spurious error on mount.

### Issues not in diff

| File:Line | Type | Issue |
|-----------|------|-------|
| HomePageGrid.tsx:89 | question | `onClose={() => setError(null)}` dismisses error but doesn't retry loading — user lands on "No modules available" with no recovery path |

distro/src/app/components/HomePageGrid/HomePageGrid.tsx
```diff
@@ -1,23 +1,36 @@
...
+    error: privilegeError,
+  } = useUserPrivilege();
   const [modules, setModules] = useState<Module[]>([]);
-  const [loading, setLoading] = useState(true);
+  const [loading, setLoading] = useState(false);
   const [error, setError] = useState<string | null>(null);
 
   const loadModules = useCallback(async () => {
+    if (privilegesLoading) return;
+    if (privilegeError || userPrivileges === null) {
```
ravinderkabli: blocking: Race condition — `userPrivileges === null` fires before provider has loaded.

`UserPrivilegeProvider` starts with `{ isLoading: false, userPrivileges: null }` (line 20 of `UserPrivilegeProvider.tsx`). Since child effects fire before parent effects in React, `HomePageGrid`'s `useEffect` sees null privileges with `isLoading=false` on the first cycle and enters this error path before any fetch has started.

Suggested fix — treat null-without-error as "still loading":
```tsx
if (privilegesLoading || (userPrivileges === null && !privilegeError)) return;
if (privilegeError) {
  setError(t('HOME_ERROR_FETCH_CONFIG'));
  return;
}
```

distro/src/app/components/HomePageGrid/HomePageGrid.tsx
```diff
@@ -1,23 +1,36 @@
...
 export const HomePageGrid: React.FC = () => {
   const { t } = useTranslation();
+  const {
+    userPrivileges,
+    isLoading: privilegesLoading,
+    error: privilegeError,
+  } = useUserPrivilege();
   const [modules, setModules] = useState<Module[]>([]);
-  const [loading, setLoading] = useState(true);
+  const [loading, setLoading] = useState(false);
```
ravinderkabli, line 17: suggestion: `useState(false)` causes a flash of "No modules available" on mount.

On first render, `privilegesLoading || loading` is `false || false`, so the skeleton is skipped and the component falls through to `modules.length === 0` → empty state. The previous `useState(true)` ensured the skeleton showed until data arrived.

Consider restoring `useState(true)`, or adding `userPrivileges === null` to the loading guard:
```tsx
if (privilegesLoading || loading || userPrivileges === null) {
  // show skeleton
}
```

distro/src/app/components/HomePageGrid/__tests__/HomePageGrid.test.tsx
```diff
@@ -119,4 +188,107 @@ describe('HomePageGrid', () => {
...
+    render(<HomePageGrid />);
+
+    await waitFor(() => {
+      expect(screen.getByTestId('app-tile-reports')).toBeInTheDocument();
+    });
+
+    expect(screen.getByText('Reports')).toBeInTheDocument();
+  });
+
+  it('shows error when privilege fetch fails', async () => {
```
ravinderkabli, line 233: suggestion: Missing test for the provider's actual initial state.

`UserPrivilegeProvider` starts with `{ isLoading: false, userPrivileges: null, error: null }`. No test covers this exact combination — the "privilege error" test below uses a non-null error. Adding a test for this state would catch the race condition described above.

distro/src/app/components/HomePageGrid/__tests__/HomePageGrid.test.tsx
```diff
@@ -119,4 +188,107 @@ describe('HomePageGrid', () => {
...
+    mockUseUserPrivilege.mockReturnValue({
+      userPrivileges: null,
+      setUserPrivileges: jest.fn(),
+      isLoading: false,
+      setIsLoading: jest.fn(),
+      error: 'Privilege fetch failed',
```
ravinderkabli: nit: Type mismatch — `error` should be `Error | null`, not a string.

`UserPrivilegeContextType.error` is typed `Error | null` (see `UserPrivilegeContext.tsx:9`). This should be `error: new Error('Privilege fetch failed')`.

distro/src/services/moduleService.ts
```diff
@@ -53,7 +53,7 @@ export const filterByPrivilege = (
   extensions: Module[],
   userPrivileges?: string[],
 ): Module[] => {
```
ravinderkabli, line 55: nit: The semantic change from `[]` → show all to `[]` → hide privileged is correct but non-obvious.

Consider an inline comment explaining the contract:
```ts
// undefined = skip privilege check (caller hasn't loaded privileges)
// []        = user has zero privileges (show only public tiles)
if (userPrivileges === undefined) {
```

distro/src/app/components/HomePageGrid/HomePageGrid.tsx
arshiyaTW2021: Fixed

distro/src/app/components/HomePageGrid/HomePageGrid.tsx
arshiyaTW2021: Done

distro/src/app/components/HomePageGrid/__tests__/HomePageGrid.test.tsx
arshiyaTW2021: Done

distro/src/app/components/HomePageGrid/__tests__/HomePageGrid.test.tsx
arshiyaTW2021: Done

distro/src/services/moduleService.ts
arshiyaTW2021: Done

ravinderkabli (approved):
Approved

PR #368, opened 2026-05-04
BAH-4664 | Fix. Send non-STAT order start date via timing.repeat.boundsPeriod

sivavithyaashree:
## Summary

- Previously, non-STAT medication orders were sending the order start date via `dosageInstruction.timing.event`, which is semantically incorrect — `event` is intended for fixed date/time occurrences, not a period start
- Non-STAT orders now send the start date via `dosageInstruction.timing.repeat.boundsPeriod.start`, which correctly represents the beginning of a dosing period
- STAT orders continue to use `dosageInstruction.timing.event` (unchanged), as they represent a single immediate occurrence

## What changed

**`medicationRequestResourceCreator.ts`**
- For non-STAT orders with a `startDate`, replaced `timing.event = [date.toISOString()]` with `timing.repeat.boundsPeriod = { start: date.toISOString() }`
- Used spread (`...timing.repeat`) to preserve any existing `repeat` fields (e.g. `duration`, `durationUnit`) when merging `boundsPeriod`

**`medicationRequestResourceCreator.test.ts`**
- Updated assertions to expect `timing.event` to be `undefined` for non-STAT orders
- Updated assertions to expect `timing.repeat.boundsPeriod.start` to carry the start date
- Added `boundsPeriod` assertions to existing duration-only test cases

## Test plan
- [x] Non-STAT order with start date: verify `dosageInstruction.timing.repeat.boundsPeriod.start` is set and `timing.event` is absent
- [x] Non-STAT order with start date + duration: verify both `boundsPeriod.start` and `repeat.duration`/`repeat.durationUnit` are present
- [x] STAT order: verify `dosageInstruction.timing.event` is still used and `timing.repeat` has no `boundsPeriod`
- [x] Run unit tests: `npx nx test clinical --testFile=apps/clinical/src/utils/fhir/__tests__/medicationRequestResourceCreator.test.ts`


apps/clinical/src/utils/fhir/medicationRequestResourceCreator.ts
```diff
@@ -142,16 +142,19 @@ const createTiming = (
       },
     };
   } else {
-    // Add event (start date) for non-STAT orders
     if (startDate) {
-      // Convert to ISO format with timezone
       const date = new Date(startDate);
-      timing.event = [date.toISOString()];
+      timing.repeat = {
```
sumaztwcode, line 147: question: Minor observation — at this point `timing.repeat` is still `undefined`, so the spread is a no-op. It works correctly (spreading `undefined` is safe in JS), but the spread only becomes meaningful in the duration block below (line 156), where it preserves `boundsPeriod` from this assignment.

Also worth noting: the two `if` blocks have an implicit ordering dependency — if someone reorders them in the future, `boundsPeriod` would be lost. A brief comment on the duration block could make this explicit. Not a blocker at all, just a future-proofing thought.

apps/clinical/src/utils/fhir/medicationRequestResourceCreator.ts
```diff
@@ -142,16 +142,19 @@ const createTiming = (
...
+        ...timing.repeat,
+        boundsPeriod: {
```
sumaztwcode, line 149: question: Non-STAT `boundsPeriod` has `start` but no `end`, while STAT orders set both. This is valid per FHIR R4 (open-ended periods are allowed), just flagging in case downstream consumers (e.g., Odoo) assume `end` is always present when `boundsPeriod` exists.

sumaztwcode (approved):
## Review for BAH-4664

Clean, focused PR that correctly fixes FHIR semantics — moving non-STAT order start dates from `timing.event` (meant for fixed date occurrences) to `timing.repeat.boundsPeriod.start` (correct for dosing period boundaries). Spread operator usage correctly preserves fields when both `boundsPeriod` and `duration`/`durationUnit` are set. Tests are thoroughly updated.

No blocking issues found. 3 informational observations posted as inline comments.

**Acceptance Criteria**: 2/2 frontend-verifiable criteria met (Odoo integration AC is backend scope)

apps/clinical/src/utils/fhir/medicationRequestResourceCreator.ts
sivavithyaashree: Good catch on the no-op spread, but I'd argue the symmetry is intentional. Both `if` blocks defensively use `...timing.repeat`, so the code is actually robust to reordering — if the duration block ran first, the startDate block's spread would preserve `duration`/`durationUnit`, and vice versa. Adding a comment to explain the dependency might obscure the symmetric pattern that prevents the dependency in the first place. Leaving as-is.

apps/clinical/src/utils/fhir/medicationRequestResourceCreator.ts
sivavithyaashree: The asymmetry is intentional and FHIR-correct: STAT orders have a known expiry (immediate dose with computed end time), so `start` + `end` is right. Non-STAT orders with a startDate represent an open-ended dosing period — `start` only is the correct representation per FHIR R4. When a duration is present on non-STAT, it lives in `timing.repeat.duration`/`durationUnit`, not as `boundsPeriod.end`. Adding an `end` would imply a hard cutoff that doesn't exist. The Odoo downstream assumption (if any) would be a backend contract concern rather than a frontend fix.

PR #370, opened 2026-05-05
BAH-4673 | Remove sorting on vital flow sheet display control

sivavithyaashree:
## Summary
- Removed sort options from all columns of the vital flow sheet display control
- `VitalFlowSheet` now passes `sortable={[]}` to `SortableDataTable`, which resolves every column to non-sortable via the component's `?? false` fallback
- Added a regression assertion in the existing happy-path test to lock in the behaviour

## JIRA
https://bahmni.atlassian.net/browse/BAH-4673

## How Tested
- [x] Unit tests passing (63/63 across 4 VitalFlowSheet suites)
- [x] Manual verification: sort icons no longer appear on any column header in the vital flow sheet UI

Co-authored-by: Anusuya Rajendran <anusuya.rajendran@thoughtworks.com>


sivavithyaashree:
Verified and tested the change for removing sort in vitals control. Attached screenshot for your reference
image
image

packages/bahmni-widgets/src/vitalFlowSheet/__tests__/VitalFlowSheet.test.tsx
```diff
@@ -29,6 +29,9 @@ jest.mock('react-router-dom', () => ({
   useParams: () => ({ patientUuid: 'test-patient-uuid' }),
 }));
 
+// eslint-disable-next-line no-var
+var mockCapturedSortable: unknown;
```
sumaztwcode, line 33: nit: `mockCapturedSortable` is module-scoped and not reset between test suites. `jest.clearAllMocks()` in `beforeEach` only clears mock functions, not plain variables. If a future test checks this without triggering a render first, it would see a stale value.

Consider adding a reset:
```ts
beforeEach(() => {
  mockCapturedSortable = undefined;
  jest.clearAllMocks();
});
```

packages/bahmni-widgets/src/vitalFlowSheet/VitalFlowSheet.tsx
```diff
@@ -292,6 +292,7 @@ const VitalFlowSheet: React.FC<VitalFlowSheetProps> = ({
       renderCell={renderCell}
       className={styles.vitalFlowSheetDataTable}
       dataTestId="vital-flow-sheet-table"
+      sortable={[]}
```
sumaztwcode: question: Inline `[]` creates a new array reference on every render. Currently `SortableDataTable` is not memoized so this has no impact, but if it ever gets wrapped in `React.memo`, this would break memoization. A module-level constant is zero-cost future-proofing:

```ts
const NO_SORTABLE_COLUMNS: { key: string; sortable: boolean }[] = [];
```

Not blocking — just a thought.

sumaztwcode (approved):
## Review for BAH-4673

Clean, minimal PR that correctly disables sorting on all vital flow sheet columns. The `sortable={[]}` mechanism is sound — works via `SortableDataTable`'s `?? false` fallback. No blocking issues.

Found 0 blocking issues, 0 suggestions, and 1 nit.

**Acceptance Criteria**: 1/1 met

sumaztwcode (changes requested):
check the custom command output that I shared

packages/bahmni-widgets/src/vitalFlowSheet/__tests__/VitalFlowSheet.test.tsx
sivavithyaashree: Good catch — added `mockCapturedSortable = undefined;` to the suite's `beforeEach` so a stale value from a prior test can't leak in if a future test doesn't trigger a render.

packages/bahmni-widgets/src/vitalFlowSheet/VitalFlowSheet.tsx
sivavithyaashree: Hoisted to a module-level `NO_SORTABLE_COLUMNS` constant — agreed it's zero-cost future-proofing if `SortableDataTable` ever gets memoized.

sumaztwcode (approved):

PR #383, opened 2026-05-07
BAH-4677 | Fix. Child Accordion Overflow In Patient Dashboard

sivavithyaashree:
## Summary
- Child accordions in Observations / History and Examinations overflowed horizontally past the patient dashboard container in normal (non-split) view
- Root cause: BAH-4450 (commit `cf82628c`) changed `.accordion` overflow globally from `hidden` → `visible` and added `.container .container { overflow-x: visible !important }` — both were intended for split-view only
- Fix: scoped both overflow rules to `[data-split-view] #main-display-area`, restoring `overflow-x: hidden` as the default in normal view while preserving the BAH-4450 split-view scrollbar fix

## Files Changed
- `packages/bahmni-design-system/src/molecules/collapsibleRowGroup/styles/CollapsibleRowGroup.module.scss` — CSS fix
- `packages/bahmni-widgets/src/observations/__tests__/__snapshots__/ObsByEncounter.test.tsx.snap` — CSS module hash update

## Test Plan
- [x] `nx lint @bahmni/design-system` — passing
- [x] `nx test @bahmni/design-system --testPathPattern=collapsibleRowGroup` — 6/6 passing
- [x] `nx test @bahmni/widgets --testPathPattern=observations` — 130/130 passing (1 snapshot updated for CSS hash)
- [x] Manual: verify child accordion no longer overflows in normal view (browser + tablet 7inch+)
- [x] Manual: verify split-view horizontal scroll still works

**JIRA:** https://bahmni.atlassian.net/browse/BAH-4677

Co-authored-by: Anusuya Rajendran <anusuya.rajendran@thoughtworks.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved horizontal scrolling and overflow handling for accordions and tables in split-view to prevent clipped or misaligned content.
  * Ensured nested accordion sections scroll independently for better usability in wide/split layouts.
  * Increased minimum content width for certain collapsible row layouts to preserve readable table and row rendering.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Bahmni/bahmni-apps-frontend/pull/383?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


sivavithyaashree:
Tested and verified changes in local. Attached screenshots for your reference: 
image
image

packages/bahmni-design-system/src/molecules/collapsibleRowGroup/styles/CollapsibleRowGroup.module.scss
```diff
@@ -8,21 +8,23 @@
...
-
 .accordion {
   display: flex;
   flex-direction: column;
   padding-inline: 0;
   border-block: none !important;
   width: 100%;
   max-width: 100%;
-  overflow: visible;
+  overflow-x: hidden;
```
sumaztwcode, line 16: nit: Per CSS spec, setting `overflow-x: hidden` without explicitly setting `overflow-y` causes the browser to compute `overflow-y` as `auto` (instead of the default `visible`). This means vertically-overflowing children (dropdowns, tooltips, absolute-positioned overlays) could be clipped in normal view.

Likely benign since accordion content flows in block direction with natural sizing, but worth a quick manual check. If clipping is observed, adding `overflow-y: visible` alongside would fix it.

sumaztwcode (commented):
## Review for BAH-4677

Clean, surgical CSS fix that correctly scopes the BAH-4450 overflow rules to split-view only. Precise root cause analysis, minimal change, consistent with established patterns, and thorough test plan.

No blocking issues found. 1 nit worth a quick manual check.

**Acceptance Criteria**: 3/3 met

### What's Done Well
- Precise root cause identification (commit cf82628c, exact rules)
- Minimal 7-line SCSS fix — no unnecessary refactoring
- Consistent with existing `:global([data-split-view] #main-display-area) &` pattern
- Thorough test plan covering lint, unit, snapshot, and manual testing

sumaztwcode (changes requested):
Verify the comment. Make sure you test all widgets on dashboard where accordion is used, as this is an important visible component on UI across forms

sivavithyaashree:
@sumaztwcode addressed both review items in commit `ed56f4bd`.

## 1. Inline nit on `.accordion { overflow-x: hidden }` ✅
Added `overflow-y: visible;` alongside, so the browser doesn't compute `overflow-y` as `auto`. This prevents any vertical clipping of dropdowns / tooltips / overlays.

## 2. CHANGES_REQUESTED — verify all widgets using accordion ✅
Audited the codebase — only the **Observations** widget consumes `CollapsibleRowGroup`. Tested it manually in both normal view and 40-60 split view, on the patient dashboard.

During manual testing in split view, I found a pre-existing regression from BAH-4450 that needed addressing:

### Regression found: per-child horizontal scrollbars in split view
Every Carbon `<div id="accordion-item-...">` (parent encounter AND each nested grouped-observation) was getting `overflow-x: auto` from this global rule in `DashboardSection.module.scss`:
```scss
:global([data-split-view] #main-display-area div[id^="accordion-item"]) {
  overflow-x: auto;
}
```
The rule was too broad — it matched nested accordion-items too, giving each child its own scrollbar. The cumulative effect made the dashboard look cluttered and the parent encounter's header appeared clipped because the inner accordion-items were independently scrolled.

### Fix
- Scoped the global rule to **only top-level accordion-items** by adding a sibling rule that resets nested accordion-items to `overflow-x: visible` (no `!important` needed; specificity wins naturally).
- Removed the now-redundant split-view overflow rules from `CollapsibleRowGroup.module.scss` (`.container` and `.accordion`) since `DashboardSection.module.scss` already handles the outer scroll at the accordion-item level.
- Bumped `.rows { min-width: 40rem → 75rem }` in split view so the info column has enough room to fully show "Recorded By [provider name]" without clipping.

### Final diff: 3 files, +12 / −18
- `apps/clinical/.../DashboardSection.module.scss` — 4 added (the new override for nested accordion-items)
- `packages/bahmni-design-system/.../CollapsibleRowGroup.module.scss` — net -8 (removed redundant rules, added `overflow-y: visible`, bumped min-width)
- `packages/bahmni-widgets/.../ObsByEncounter.test.tsx.snap` — CSS module hash update only

### Tested
- ✅ Normal view: child accordion no longer overflows the dashboard (original BAH-4677 fix)
- ✅ Split view: single outer scrollbar, no per-child scrollbars, info column fully readable at max-right scroll
- ✅ `nx lint @bahmni/design-system`, `nx test @bahmni/widgets`, `nx test @bahmni/clinical-app`

packages/bahmni-design-system/src/molecules/collapsibleRowGroup/styles/CollapsibleRowGroup.module.scss
```diff
@@ -48,7 +40,7 @@
   max-width: 100%;
   box-sizing: border-box;
   :global([data-split-view] #main-display-area) & {
-    min-width: 40rem;
```
sumaztwcode, line 51: why is min-width changed?

sivavithyaashree: At `40rem` (640px), the info column ends up ~125px wide (20% of rowCell after padding/gap). The "Recorded By [provider name]" text was getting clipped even when scrolled fully right, since the info column had `max-width: 20%` and couldn't grow to fit the text.

`75rem` was the smallest value at which "Recorded By Super Man" fully fits in the info column under the 40-60 split view that I tested. The closest "matching" constant in this area is `$table-min-width: 62.5rem` (used by the data-table widgets in `DashboardSection.module.scss`), but that wasn't quite wide enough for the longer recorded-by text in the observations widget.

Happy to revert to `40rem` and file the info-column sizing as a separate ticket if you'd rather keep BAH-4677's scope tight to just the original child-overflow fix + your nit.

apps/clinical/src/components/dashboardSection/styles/DashboardSection.module.scss
```diff
@@ -16,6 +16,10 @@ $table-cell-min-width: 6.875rem;
   overflow-x: auto;
 }
```
PratitiP: The id targeted here looks generic. Please make sure if its not creating side effect for other accordion-items in the split-view as there are lot of accordions used across the split-view

sivavithyaashree: Good catch — verified across every widget that uses Carbon Accordion in this repo:

| Widget | Accordion structure | Affected? |
|---|---|---|
| `Observations` (via `CollapsibleRowGroup`) | **Nested**: parent encounter accordion → child grouped-obs accordion | Yes (intentionally — this is what we're fixing) |
| `MedicationsTable` | Single-level: one `<Accordion>` with sibling `<AccordionItem>` per date | No |
| `LabInvestigation` | Single-level | No |
| `RadiologyInvestigationTable` | Single-level | No |
| `FormsTable` | Single-level | No |
| `GenericServiceRequestTable` | Single-level | No |

The selector `div[id^="accordion-item"] div[id^="accordion-item"]` requires an `accordion-item` div to be a descendant of another `accordion-item` div — i.e. true nesting. Single-level accordions (sibling items inside one `<Accordion>`) don't match. Also verified manually in the UI in both full view and 40-60 split view across these widgets — only Observations behavior changes, the others render unchanged.

PratitiP: Okay. Looks good for now. But if we are targeting only Observations display control, better to add in that widget styles itself. Why to control from dashboard ?

sivavithyaashree: Good point — moved the override into `Observations.module.scss` and scoped it under the `.observations` class, so it now only affects the Observations widget. Removed from `DashboardSection.module.scss`. Pushed in `1b0599a0`.

PratitiP (approved):

sivavithyaashree:
@PratitiP addressed your full feedback in `bfaabdc9`.

## Refactor summary

Moved all the split-view widget-specific styles out of `DashboardSection.module.scss` and into each widget's own SCSS, scoped under widget IDs. Removed the SCSS variables. Replaced `.cds--data-table` class overrides with element selectors.

### Changes per concern you raised:

**1. "Dashboard contains multiple tables. If we want to handle for any specific table, better to add styles in respective widgets"**
- Moved the `[accordion-item] overflow-x: auto` rule into each widget that uses Accordion: `MedicationsTable`, `LabInvestigation`, `RadiologyInvestigationTable`, `FormsTable`, `GenericServiceRequestTable`, plus `Observations` (already moved earlier).
- Moved the wide-table rules (`min-width`, `td/th min-width`, `td:first-child unset`) into the widgets that actually have 5+ column tables: `MedicationsTable` and `RadiologyInvestigationTable`. Other widgets are unaffected since the `:has(td:nth-child(5))` predicate gates the rules.

**2. "No need to define scss variables in module files"**
- Removed `$table-min-width: 62.5rem` and `$table-cell-min-width: 6.875rem` from `DashboardSection.module.scss`. The values are inlined in the widgets that use them (the values are layout-specific and don't map to Carbon spacing tokens, which top out at `~10rem`).

**3. "We follow carbon spacing tokens"**
- Where applicable (block padding, etc.) the rules keep using `$spacing-05`. The two table-sizing values don't map to Carbon tokens (1000px and 110px); these are inlined as-is.

**4. "Do not prefer overriding Carbon classNames. Give id on the parent elements and target needed html elements in shadow DOM"**
- Added `id` attribute on each widget's outer wrapper: `medications-table`, `lab-investigations`, `forms-table`, `generic-service-request-table` (Radiology and Observations already had IDs).
- Each widget's rules are now scoped under `#<widget-id>`, e.g., `#main-display-area #medications-table table:has(td:nth-child(5))`.
- Replaced `.cds--data-table` Carbon class references with the `table` element selector. The Carbon-specific `div[id^="accordion-item"]` pattern remains since it's how Carbon exposes accordion content areas and there's no widget-defined wrapper around them.

### Files touched (12)
- `apps/clinical/src/components/dashboardSection/styles/DashboardSection.module.scss` — removed vars and 3 rule blocks
- 5 widget TSX files: added `id` attribute on outer wrapper; `LabInvestigation.tsx` also wrapped its top-level `<Accordion>` in `<div id="lab-investigations">`
- 6 widget SCSS files: added the scoped rules
- 1 snapshot updated (`FormsTable.test.tsx.snap`) for the new `id` attribute

### Net behavior
Identical to before this commit. Verified `nx test @bahmni/widgets` (1094 tests, 16 snapshots), `nx test @bahmni/clinical-app` (1844 tests), and `nx lint` all passing. Also manually verified in `localhost:3000` patient dashboard — split view scrolling works on all affected widgets, normal view unchanged, no per-child inner scrollbars in Observations.

sumaztwcode (approved):

packages/bahmni-widgets/src/forms/styles/FormsTable.module.scss
```diff
@@ -3,6 +3,12 @@
 @use "@carbon/react/scss/colors" as *;
 @use "@carbon/react/scss/theme" as *;
```
PratitiP, line 5: Do we need [data-split-view] #main-display-area selectors as we are writing tests module scoped. I think forms-table id would be sufficient.
If tomorrow, FormsTable widgets is used with some other template which might not have #main-display-area, then Forms table should self contain these styles irrespective of data-split-view or main-display-area

PR #385, opened 2026-05-07
BAH 4665 | Match decision API integrated

sahupadma:
JIRA - https://bahmni.atlassian.net/browse/BAH-4665

Integrates the Consultation page with the backend match-decision API to determine whether an existing encounter should be resumed or a new one created.                    
                                                                                                                                                                                                                                                                                                 
                                                                                                                                                                               
  Changes                                                                                                                                                                      
                                                                                                                                                                             
  New useEncounterMatchDecision hook (packages/bahmni-widgets/src/encounterResume/)                                                                                            
  - Calls POST /openmrs/ws/rest/v1/bahmnicore/bahmniencounter/match-decision with patient, visit, provider, and location UUIDs
  - Derives UI-ready flags from the backend response:                                                                                                                          
    - canResume — true only on match_found; caller should resume the existing encounter                                                                                      
    - showEditButton — true for match_found or no_match; show the edit button with different behaviour                                                                         
    - shouldShowButton — false for no_active_visit or error; hide the button entirely                                                                                          
  - Exported from @bahmni/widgets alongside full TypeScript types                                                                                                              
                                                                                                                                                                               
  ConsultationPage integration                                                                                                                                                 
  - On page load (once per session via useRef guard), fires the match-decision API call with the current patient, active visit, practitioner, and resolved location          
  - Bootstraps selectedLocation from the locations list if not already set, and syncs activeVisit into the encounter details store                                             

                                                                                                                                                                             
  Tests                                                                                                                                                                        
                                                                                                                                                                             
  - 14 unit tests for useEncounterMatchDecision covering all status values (match_found, no_match, no_active_visit), missing parameters, correct payload shape (optional       
  providerUuid omitted when absent), and refetch behaviour                                                                                                                   
  - Updated ConsultationPage tests with mocks for useActivePractitioner, usePatientUUID, useActiveVisit, and useLocations


apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -35,11 +35,21 @@ import {
 interface ConsultationPadProps {
   encounterType: string;
   onClose: () => void;
+  /**
```
sumaztwcode: remove comments across the changes where code is self-explanatory

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -106,10 +116,19 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
   }, [resolvedEncounterType]);
 
   const { practitioner } = useActivePractitioner();
-  const { activeEncounter } = useEncounterSession({
+  const { activeEncounter: sessionEncounter } = useEncounterSession({
     practitioner,
     encounterTypeUUID: selectedEncounterType?.uuid,
   });
+
+  // When mode === 'new', always create a new encounter regardless of any existing session.
```
sumaztwcode: remove comments

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -84,6 +91,63 @@ const ConsultationPage: React.FC = () => {
       setIsActionAreaVisible(true);
     }, []),
   );
+
+  const patientUUID = usePatientUUID();
```
sumaztwcode, line 93: For now the edit consultation is triggered from dashboard display. The checks added here in ConsultationPage can move to a utils file? I will revisit this comment after going thru the changes.

packages/bahmni-widgets/src/encounterResume/useEncounterMatchDecision.ts
```diff
@@ -0,0 +1,171 @@
...
+}
+
+export interface UseEncounterMatchDecisionParams {
+  patientUuid: string | null | undefined;
+  visitUuid: string | null | undefined;
+  providerUuid?: string | null | undefined;
+  locationUuid: string | null | undefined;
+}
+
+export interface UseEncounterMatchDecisionResult {
```
sumaztwcode: Interface name also contains Use? Is that required?

packages/bahmni-widgets/src/encounterResume/useEncounterMatchDecision.ts
```diff
@@ -0,0 +1,171 @@
...
+  status: EncounterMatchStatus | null;
+  encounterUuid: string | null;
+  reason: string | null;
```
sumaztwcode, line 45: remove comments

packages/bahmni-widgets/src/encounterResume/useEncounterMatchDecision.ts
```diff
@@ -0,0 +1,171 @@
...
+async function fetchEncounterMatchDecision(
+  params: EncounterMatchDecisionRequest,
+): Promise<EncounterMatchDecisionResponse> {
+  return post<EncounterMatchDecisionResponse, EncounterMatchDecisionRequest>(
+    ENCOUNTER_MATCH_DECISION_URL,
+    params,
+  );
+}
+
+/**
```
sumaztwcode: remove comments - code below is self-explanatory

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -84,6 +91,63 @@ const ConsultationPage: React.FC = () => {
...
+
+  useEffect(() => {
+    if (
+      !matchDecisionFetchedRef.current &&
+      patientUUID &&
+      activeVisit?.id &&
+      practitioner?.uuid &&
+      selectedLocation?.uuid
+    ) {
+      matchDecisionFetchedRef.current = true;
```
sumaztwcode: blocking: The hook `useEncounterMatchDecision` (171 lines, 14 tests) derives `canResume`, `showEditButton`, `shouldShowButton` from the API response — but it's never called here. Instead this raw `post()` fires and the response is discarded (no `.then()`, only `.catch(() => {})`).

The hook's derived flags drive nothing, making it dead code in the application.

**Fix:** Replace this `useRef`/`useEffect`/`post()` block with:
```tsx
const { canResume, showEditButton } = useEncounterMatchDecision({
  patientUuid: patientUUID,
  visitUuid: activeVisit?.id,
  providerUuid: practitioner?.uuid,
  locationUuid: selectedLocation?.uuid,
});
```
The hook already handles the "disabled until all params present" guard, so `matchDecisionFetchedRef` is no longer needed.

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -84,6 +91,63 @@ const ConsultationPage: React.FC = () => {
...
+  const { activeVisit: fetchedActiveVisit } = useActiveVisit(patientUUID);
+
+  useEffect(() => {
+    setActiveVisit(fetchedActiveVisit ?? null);
+  }, [fetchedActiveVisit, setActiveVisit]);
+
+  const activeVisit = useEncounterDetailsStore((state) => state.activeVisit);
+
+  // Call match-decision API on page load - only once per session
+  const matchDecisionFetchedRef = useRef(false);
```
sumaztwcode: suggestion: `matchDecisionFetchedRef` doesn't reset when `patientUUID` changes. Since React Router reuses this component instance when navigating between patients, the API will only fire for the first patient — subsequent patients are skipped.

The `useEncounterMatchDecision` hook avoids this because its `queryKey` includes `patientUuid`, giving automatic per-patient invalidation.

**If keeping the ref approach**, add:
```tsx
useEffect(() => {
  matchDecisionFetchedRef.current = false;
}, [patientUUID]);
```

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -84,6 +91,63 @@ const ConsultationPage: React.FC = () => {
...
+  );
+  const setSelectedLocation = useEncounterDetailsStore(
+    (state) => state.setSelectedLocation,
+  );
+
+  const { locations } = useLocations();
+
+  useEffect(() => {
+    if (locations.length > 0 && !selectedLocation) {
+      setSelectedLocation(locations[0]);
```
sumaztwcode: nit: This location bootstrap (`setSelectedLocation(locations[0])`) and the `setActiveVisit` sync below duplicate logic that already exists in `EncounterDetails.tsx`. When both components are mounted, they race to write to the same Zustand store slices.

Consider extracting into a shared hook or removing the duplicate — the store will be populated by `EncounterDetails` when the pad opens.

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -84,6 +91,63 @@ const ConsultationPage: React.FC = () => {
...
+    }
+  }, [locations, selectedLocation, setSelectedLocation]);
+
+  const { activeVisit: fetchedActiveVisit } = useActiveVisit(patientUUID);
+
+  useEffect(() => {
```
sumaztwcode: nit: `setActiveVisit(fetchedActiveVisit ?? null)` fires on every `fetchedActiveVisit` reference change, even if the visit UUID is the same. Zustand notifies all subscribers on every `set` call. Consider adding an identity check:
```tsx
if (fetchedActiveVisit?.id !== activeVisit?.id) {
  setActiveVisit(fetchedActiveVisit ?? null);
}
```

apps/clinical/src/pages/__tests__/ConsultationPage.test.tsx
```diff
@@ -234,6 +248,20 @@ describe('ConsultationPage', () => {
...
+    (useActivePractitioner as jest.Mock).mockReturnValue({
+      practitioner: { uuid: 'provider-uuid', name: 'Dr. John' },
+    });
+
+    (usePatientUUID as jest.Mock).mockReturnValue('patient-uuid');
+
+    (useActiveVisit as jest.Mock).mockReturnValue({
+      activeVisit: { id: 'visit-uuid' },
+    });
+
```
sumaztwcode, line 276: nit: `post` is not mocked in the `@bahmni/services` mock block, so the `post()` call in the match-decision `useEffect` hits the real import. No test verifies the API call payload or the ref guard deduplication. Consider adding `post: jest.fn().mockResolvedValue({})` and an assertion.

packages/bahmni-widgets/src/encounterResume/useEncounterMatchDecision.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  const { data, isLoading, error, refetch } = useQuery<
+    EncounterMatchDecisionResponse | null,
+    Error
+  >({
+    queryKey: [
+      'encounterMatchDecision',
+      patientUuid,
+      visitUuid,
+      providerUuid,
```
sumaztwcode, line 78: question: Since this uses `useQuery` for a POST, the global retry policy may cause up to 3 POST calls on transient errors. The test file already sets `retry: false` — consider adding the same here for consistency:
```tsx
retry: false,
```

packages/bahmni-widgets/src/encounterResume/useEncounterMatchDecision.ts
```diff
@@ -0,0 +1,171 @@
...
+    ],
+    enabled: isEnabled,
+    queryFn: () => {
+      if (!patientUuid || !locationUuid) {
+        return Promise.reject(
+          new Error('patientUuid and locationUuid are required'),
+        );
+      }
+      return fetchEncounterMatchDecision({
+        patientUuid,
```
sumaztwcode, line 85: question: This guard is unreachable — `useQuery` only calls `queryFn` when `enabled` is `true`, which already requires `patientUuid && locationUuid`. Can be removed to reduce noise, or add a comment that it's a defensive backstop.

sumaztwcode (changes requested):
## Review for BAH-4665

Good hook design with clean TypeScript types and thorough test coverage (14 tests). The main concern is that `ConsultationPage` bypasses the hook and fires a raw `post()` that discards the response — the hook is effectively dead code. Also, the `useRef` guard won't reset when navigating between patients.

Found 1 blocking issue, 1 suggestion, and 3 nits.

**Acceptance Criteria**: 5/7 met, 2/7 partially met (hook not used in ConsultationPage; mode prop is follow-up scaffolding)

packages/bahmni-widgets/src/encounterResume/useEncounterMatchDecision.ts
sahupadma: done

packages/bahmni-widgets/src/encounterResume/useEncounterMatchDecision.ts
sahupadma: done

apps/clinical/src/pages/__tests__/ConsultationPage.test.tsx
sahupadma: added

apps/clinical/src/pages/ConsultationPage.tsx
sahupadma: done

apps/clinical/src/pages/ConsultationPage.tsx
sahupadma: done

apps/clinical/src/pages/ConsultationPage.tsx
sahupadma: done

apps/clinical/src/pages/ConsultationPage.tsx
sahupadma: done

packages/bahmni-widgets/src/encounterResume/useEncounterMatchDecision.ts
sahupadma: done

packages/bahmni-widgets/src/encounterResume/useEncounterMatchDecision.ts
sahupadma: removed

packages/bahmni-widgets/src/encounterResume/useEncounterMatchDecision.ts
sahupadma: changed

apps/clinical/src/pages/ConsultationPage.tsx
sahupadma: done

apps/clinical/src/components/consultationPad/index.tsx
sahupadma: removed

apps/clinical/src/components/consultationPad/index.tsx
sahupadma: removed

sahupadma:
Covered in BAH-4701: [Feature UI] Encounter Match Decision Mapper with Diagnostic Reasons
Closed
  thats why closing the PR

PR #394, opened 2026-05-11
BAH-4552 | Add. Resolve Transitive Vulnerability By Pinning fast-uri To 3.1.2

rahu1ramesh:
Addresses transitive security vulnerabilities by pinning fast-uri to 3.1.2 via Yarn resolutions and upgrading immutable from 3.8.1 to 4.3.8.

Key Features

  - Pins fast-uri@3.1.2 in resolutions to override vulnerable transitive dependency versions pulled in by downstream packages.
  - Upgrades immutable from 3.8.1 to 4.3.8 to resolve known vulnerabilities in the older version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,8 @@
   },
   "resolutions": {
     "path-to-regexp": "0.1.13",
-    "picomatch": "4.0.4"
+    "picomatch": "4.0.4",
+    "fast-uri": "3.1.2"
   },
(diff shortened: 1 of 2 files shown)


mohan-13 (approved):

mohan-13:
While this is an immediate fix, we need to review transitive dependencies and upgrade the root packages @rahu1ramesh

PR #398, opened 2026-05-12
BAH-4700: Refetch encounter session when consultation is saved

hamsavarthiniR-Bahmni:
## Summary

Fixes a bug where the "New Consultation" button does not update to "Edit Consultation" after a consultation is saved and the pad is closed.

## Root Cause

After BAH-4701 was merged, encounter session management was moved to `PatientHeader.tsx` (via `useEncounterSession`). The `consultationSaved` event subscription needed to live in PatientHeader so it can call the `refetch` function returned by `useEncounterSession`.

## Fix

**`PatientHeader.tsx`** — Subscribe to `consultationSaved` event via `useSubscribeConsultationSaved` and call `refetch()` to re-query the encounter session when a consultation is saved. The `refetch` function is now destructured from `useEncounterSession`.

This is the architecturally correct location for the refetch (post BAH-4701) since encounter session state lives in PatientHeader.

## Changes

- `apps/clinical/src/components/patientHeader/PatientHeader.tsx` — Add `useSubscribeConsultationSaved` to trigger `refetch` on consultation save
- `apps/clinical/src/components/patientHeader/__tests__/PatientHeader.test.tsx` — Add test verifying refetch is called on consultation saved event
- `apps/clinical/src/components/patientHeader/ConsultationActionButton.tsx` — Aligned with BAH-4701's props-based approach (no functional change)
- `apps/clinical/src/hooks/useEncounterSession.ts` — Updated to BAH-4701's implementation (matchReason, resolveEncounterMatchDecision)
- `apps/clinical/src/hooks/__tests__/useEncounterSession.test.ts` — Updated tests for new useEncounterSession implementation

## Jira

https://bahmni.atlassian.net/browse/BAH-4700


apps/clinical/src/components/patientHeader/ConsultationActionButton.tsx
```diff
@@ -20,13 +23,20 @@ const ConsultationActionButton: React.FC<ConsultationActionButtonProps> = ({
...
+
   const canAddEncounter = useHasPrivilege(
     CONSULTATION_PAD_PRIVILEGES.ENCOUNTER,
   );
 
+  useSubscribeConsultationSaved(() => {
+    if (canAddEncounter) {
+      refetch();
+    }
+  }, [canAddEncounter, refetch]);
```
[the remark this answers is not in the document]
Varun-beta: This is a false positive. Refetching on pad-close (without save) is unnecessary — cancel means nothing was saved, so the server state hasn't changed. The button state correctly stays the same as before opening the pad.

The only time the button needs to update is after a **successful save**, which is already covered by `useSubscribeConsultationSaved`. The `consultationSaved` event only fires after `submitConsultation` succeeds (in `consultationPad/index.tsx`), which is the right trigger.

Adding a `useRef`+`useEffect` for `isActionAreaVisible` transitions would just add complexity and a wasted API call on cancel.

apps/clinical/src/hooks/__tests__/useEncounterSession.test.ts
```diff
@@ -0,0 +1,263 @@
...
+    act(() => {
+      rerender({ practitioner: newPractitioner });
+    });
+
+    // State should be reset immediately before the new fetch completes
+    await waitFor(() => {
+      expect(result.current.hasActiveSession).toBe(false);
+      expect(result.current.activeEncounter).toBeNull();
+      expect(result.current.isPractitionerMatch).toBe(false);
+    });
```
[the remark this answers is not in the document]
hamsavarthiniR-Bahmni: Fixed in commit 0b6a5be — split the `waitFor` block into:
1. Synchronous assertions (`expect(result.current.hasActiveSession).toBe(false)` etc.) immediately after `act(() => rerender(...))` to verify the state resets synchronously before the delayed fetch resolves
2. A separate `await waitFor(() => expect(result.current.isLoading).toBe(false))` to let the async fetch lifecycle settle

This ensures the test validates the *immediate* reset behaviour, not just eventual consistency.

apps/clinical/src/components/patientHeader/ConsultationActionButton.tsx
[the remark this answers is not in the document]
hamsavarthiniR-Bahmni: Confirmed — no code change needed here. As @Varun-beta explained and CodeRabbit acknowledged, this suggestion is a false positive. The `consultationSaved` event fires only after a successful `submitConsultation`, making it the correct and only required trigger. Adding a `useRef`+`useEffect` to also refetch on every pad-close/cancel would introduce an unnecessary API call when server state hasn't changed. The current implementation is intentional and correct. Please resolve this thread.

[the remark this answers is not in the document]
hamsavarthiniR-Bahmni: Bot confirming: this thread is a **false positive** — the CodeRabbit suggestion has been retracted and both reviewers confirmed no code change is needed. The current implementation using `useSubscribeConsultationSaved` is correct and intentional. All CI checks are passing. Please resolve this thread so the PR can proceed to merge.

apps/clinical/src/components/patientHeader/PatientHeader.tsx
```diff
@@ -28,10 +29,15 @@ const PatientHeader: React.FC<PatientHeaderProps> = ({
...
-  const { matchReason, editActiveEncounter, isLoading } = useEncounterSession({
-    practitioner,
-    encounterTypeUUID: CONSULTATION_ENCOUNTER_TYPE_UUID,
-  });
+  const { matchReason, editActiveEncounter, isLoading, refetch } =
+    useEncounterSession({
+      practitioner,
+      encounterTypeUUID: CONSULTATION_ENCOUNTER_TYPE_UUID,
+    });
+
```
sumaztwcode, line 42: blocking: Every other `useSubscribeConsultationSaved` subscriber in the codebase guards on `payload.patientUUID === patientUUID` before acting (`MedicationsTable`, `MedicationsForm`, `LabInvestigation`, `AdministeredTab`, `VaccinationForm`). This callback calls `refetch()` unconditionally, so a consultation saved for *any* patient triggers a FHIR refetch here.

Suggested fix:
```tsx
useSubscribeConsultationSaved((payload) => {
  if (payload.patientUUID === patientUUID) {
    refetch();
  }
}, [patientUUID, refetch]);
```
`usePatientUUID` is available from `@bahmni/widgets` (already imported in this file).

apps/clinical/src/components/patientHeader/PatientHeader.tsx
```diff
@@ -28,10 +29,15 @@ const PatientHeader: React.FC<PatientHeaderProps> = ({
...
+  useSubscribeConsultationSaved(() => {
+    refetch();
```
sumaztwcode: suggestion: `refetch` is recreated on every render inside `useEncounterSession` (no `useCallback`), so `[refetch]` in deps causes the subscription to tear down and re-register on every render.

Since `useSubscribeConsultationSaved` uses a `useRef`-based design internally, you can safely pass `[]` (or just `[patientUUID]` if the guard above is added) — the ref ensures the latest callback is always invoked.

apps/clinical/src/components/patientHeader/__tests__/PatientHeader.test.tsx
```diff
@@ -197,4 +197,41 @@ describe('PatientHeader Component', () => {
       expect(header).not.toHaveAttribute('data-can-edit-encounter');
     });
   });
+
+  describe('Consultation saved refetch', () => {
```
sumaztwcode, line 201: suggestion: The happy-path test is good, but there's no negative test case. Once the `patientUUID` guard is added, consider adding:
```ts
test('does not call refetch for a different patient UUID', () => {
  // ... same setup ...
  act(() => {
    window.dispatchEvent(new CustomEvent(CONSULTATION_SAVED_EVENT, {
      detail: { patientUUID: 'other-patient-uuid', ... }
    }));
  });
  expect(mockRefetch).not.toHaveBeenCalled();
});
```

apps/clinical/src/hooks/__tests__/__mocks__/encounterSessionMocks.ts
```diff
@@ -0,0 +1,36 @@
+import { Provider } from '@bahmni/services';
```
sumaztwcode: nit: This file is not imported by any test in this PR. The `PatientHeader` test inlines its own mock values instead. Either import these constants in the test for consistency, or defer adding this file until the tests that consume it arrive.

sumaztwcode (changes requested):
## Review for BAH-4700

Clean, focused bug fix — architecturally correct placement (PatientHeader owns encounter session state post BAH-4701). One critical gap to address before merge.

Found 1 blocking issue, 2 suggestions, and 1 nit.

**Acceptance Criteria**: 1/1 partially met (mechanism correct but missing patient-UUID guard)

### What's Done Well
- Architecturally correct placement — subscription must live where `useEncounterSession` is called
- Uses the established `useSubscribeConsultationSaved` pattern
- Clean, minimal 127-line diff focused on the exact bug
- Good PR description explaining root cause and placement rationale

apps/clinical/src/components/patientHeader/PatientHeader.tsx
hamsavarthiniR-Bahmni: Fixed in commit ee9aba3 — Added the `patientUUID` guard as suggested. `usePatientUUID()` is now called in `PatientHeader` and the callback checks `payload.patientUUID === patientUUID` before calling `refetch()`, matching the pattern used in `MedicationsTable`, `MedicationsForm`, and other subscribers. A negative test case was also added in the test file to verify refetch is not triggered for a different patient's saved consultation.

apps/clinical/src/components/patientHeader/PatientHeader.tsx
hamsavarthiniR-Bahmni: Fixed in commit ee9aba3 — Removed `refetch` from the deps array and replaced it with `[patientUUID]`. Since `useSubscribeConsultationSaved` uses a `useRef`-based design internally (the latest callback is always invoked via the ref), `refetch` does not need to be in deps. Using `[patientUUID]` is correct — the subscription only re-registers if the patient UUID changes, which avoids the unnecessary teardown/re-register on every render caused by `refetch` being recreated each render.

apps/clinical/src/components/patientHeader/__tests__/PatientHeader.test.tsx
hamsavarthiniR-Bahmni: Fixed in commit 26a6a19 — Added the negative test case: `'does not call refetch when consultation saved event is dispatched for a different patient'`. It dispatches `CONSULTATION_SAVED_EVENT` with `patientUUID: 'other-patient-uuid'` (the mocked `usePatientUUID` returns `'patient-uuid'`) and asserts `expect(mockRefetch).not.toHaveBeenCalled()`.

apps/clinical/src/hooks/__tests__/__mocks__/encounterSessionMocks.ts
hamsavarthiniR-Bahmni: Fixed in commit 5625682 — Removed `encounterSessionMocks.ts` from the branch. As noted, it wasn't imported by any test in this PR (PatientHeader.test.tsx inlines its own mock values). It can be re-added when tests that actually consume it are added.

sumaztwcode (approved):

PR #405, opened 2026-05-13
Feature/BAH-4693 Display patient name in Clinical Dashboard breadcrumb

Kanchi-Oswal-twk:
**Summary**                                                                                                                                       
                                                                                                                                                
  Display the patient's actual name in the Clinical Dashboard breadcrumb instead of the generic "Current Patient" text, so clinicians can       
  confirm they are viewing the correct patient's record at a glance.
                                                                                                                                                
  **What Changed**                                                                                                                                

  -packages/bahmni-services/src/utils/utils.ts — No new utility added; uses the existing          
  capitalize utility  
  - apps/clinical/src/pages/ConsultationPage.tsx — Fetch patient data via useQuery with the same cache key as PatientDetails widget (no extra   
  API call), display patient name in breadcrumb with fallback                                                                                   
  - apps/clinical/src/pages/__tests__/ConsultationPage.test.tsx — Added 4 tests covering patient name display, loading fallback, error fallback,
   and unavailable fallback                                                                                                                     
                                                                                                                                              
  **How It Works**                                                                                                                                  
                                                                                                                                              
  - Patient UUID is read from the URL via usePatientUUID()                                                                                      
  - Patient data is fetched via getFormattedPatientById — React Query deduplicates this with the PatientDetails widget so no additional API call
   is made                                                                                                                                      
  - Breadcrumb shows capitalize(patient.fullName) when available, falls back to t('CURRENT_PATIENT') during loading, error, or when name is   
  unavailable                                                                                                                                   
  - breadcrumbItems wrapped in useMemo consistent with globalActions and sidebarItems in the same file                                        
                                                                                                                                                
  **Acceptance Criteria**                                                                                                                         
                                                                                                                                                
  - Patient name displayed in breadcrumb (e.g., John Doe)                                                                                       
  - Shows First name / Middle name / Last name as recorded during registration
  - Fallback to "Current Patient" while loading                                                                                                 
  - Fallback to "Current Patient" on error or null name                                                                                       
  - Updates on patient switch (React Query refetches on UUID change)                                                                            
  - Localization supported — uses existing CURRENT_PATIENT translation key (EN: "Current Patient", ES: "Paciente Actual")                       
  - Tab responsive — no layout changes, uses existing breadcrumb rendering  
 

**Patient Name in the breadcrumb:**

<img width="2958" height="1170" alt="image" src="https://github.com/user-attachments/assets/ad56c870-2588-4ba5-93cc-9beaf1264c66" />

<img width="3024" height="940" alt="image" src="https://github.com/user-attachments/assets/cadf1a02-b935-4e0f-97d6-02ea87ea421d" />


**When Patient information is not loaded yet :**

<img width="2978" height="700" alt="image" src="https://github.com/user-attachments/assets/5a4376e7-dc16-46cb-8163-c075149fffaf" />



**Translation for the “Current Patient” :** 
<img width="1356" height="594" alt="image" src="https://github.com/user-attachments/assets/e9e1db0c-229f-40dd-8b8c-894003c5423e" />



 


      


## Summary by CodeRabbit

* **New Features**
  * Patient names now display in the breadcrumb navigation on the consultation page, automatically formatted with proper capitalization. Falls back to the default label when patient information is unavailable.

* **Tests**
  * Enhanced test coverage for breadcrumb rendering across various patient data states.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/Bahmni/bahmni-apps-frontend/pull/405)

<!-- review_stack_entry_end -->

### DIFF
--- a/apps/clinical/src/pages/ConsultationPage.tsx
+++ b/apps/clinical/src/pages/ConsultationPage.tsx
@@ -11,11 +11,14 @@ import {
   BAHMNI_HOME_PATH,
   getConfig,
   generateId,
+  getFormattedPatientById,
+  capitalizeFirstLetter,
 } from '@bahmni/services';
 import {
   ProgramDetails,
   useNotification,
   useUserPrivilege,
+  usePatientUUID,
 } from '@bahmni/widgets';
 import { useQuery } from '@tanstack/react-query';
 import React, {
@@ -121,15 +124,31 @@ const ConsultationPage: React.FC = () => {
   );
   const viewingForm = useObservationFormsStore((state) => state.viewingForm);
 
-  const breadcrumbItems = [
-    { id: 'home', label: 'Home', href: BAHMNI_HOME_PATH },
-    {
-      id: 'clinical',
-      label: 'Clinical',
-      href: BAHMNI_CLINICAL_PATH,
-    },
-    { id: 'current', label: t('CURRENT_PATIENT'), isCurrentPage: true },
-  ];
+  const patientUUID = usePatientUUID();
+  const { data: patient } = useQuery({
+    queryKey: ['patient', patientUUID],
+    queryFn: () => getFormattedPatientById(patientUUID!),
+    enabled: !!patientUUID,
+  });
+
+  const breadcrumbItems = useMemo(
+    () => [
+      { id: 'home', label: 'Home', href: BAHMNI_HOME_PATH },
+      { id: 'clinical', label: 'Clinical', href: BAHMNI_CLINICAL_PATH },
+      {
+        id: 'current',
+        label: patient?.fullName
+          ? patient.fullName
+              .trim()
+              .split(/\s+/)
+              .map(capitalizeFirstLetter)
+              .join(' ')
+          : t('CURRENT_PATIENT'),
+        isCurrentPage: true,
+      },
+    ],
+    [patient?.fullName, t],
+  );
 
   const episodeUuids = useMemo(() => {
     const episodeUuid = searchParams.get(EPISODE_UUID_SEARCH_PARAMS_KEY);
--- a/apps/clinical/src/pages/ConsultationPage.tsx
+++ b/apps/clinical/src/pages/ConsultationPage.tsx
@@ -12,7 +12,7 @@ import {
   getConfig,
   generateId,
   getFormattedPatientById,
-  capitalizeFirstLetter,
+  capitalize,
 } from '@bahmni/services';
 import {
   ProgramDetails,
@@ -138,11 +138,7 @@ const ConsultationPage: React.FC = () => {
       {
         id: 'current',
         label: patient?.fullName
-          ? patient.fullName
-              .trim()
-              .split(/\s+/)
-              .map(capitalizeFirstLetter)
-              .join(' ')
+          ? capitalize(patient.fullName)
           : t('CURRENT_PATIENT'),
         isCurrentPage: true,
       },
--- a/apps/clinical/src/pages/__tests__/ConsultationPage.test.tsx
+++ b/apps/clinical/src/pages/__tests__/ConsultationPage.test.tsx
@@ -402,7 +402,7 @@ describe('ConsultationPage', () => {
     it('should show "Current Patient" fallback in breadcrumb while patient data is loading', async () => {
       (usePatientUUID as jest.Mock).mockReturnValue('test-patient-uuid');
       (getFormattedPatientById as jest.Mock).mockReturnValue(
-        new Promise(() => { }),
+        new Promise(() => {}),
       );
 
       renderWithProvider();
--- a/apps/clinical/src/pages/__tests__/ConsultationPage.test.tsx
+++ b/apps/clinical/src/pages/__tests__/ConsultationPage.test.tsx
@@ -1,8 +1,9 @@
-import { getConfig } from '@bahmni/services';
+import { getConfig, getFormattedPatientById } from '@bahmni/services';
 import {
   useHasPrivilege,
   useNotification,
   useUserPrivilege,
+  usePatientUUID,
 } from '@bahmni/widgets';
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
 import { render, screen, waitFor, fireEvent } from '@testing-library/react';
@@ -128,6 +129,7 @@ jest.mock('@bahmni/design-system', () => ({
       activeSideNavItemId,
       globalActions,
       onSideNavItemClick,
+      breadcrumbItems,
     }) => (
       <div data-testid="mocked-header-component">
         {globalActions?.map(
@@ -164,6 +166,18 @@ jest.mock('@bahmni/design-system', () => ({
         <div data-testid="active-sidenav-item">
           {activeSideNavItemId ?? 'none'}
         </div>
+        {breadcrumbItems?.map(
+          (item: {
+            id: string;
+            label: string;
+            href?: string;
+            isCurrentPage?: boolean;
+          }) => (
+            <div key={item.id} data-testid={`breadcrumb-item-${item.id}`}>
+              {item.label}
+            </div>
+          ),
+        )}
       </div>
     ),
   ),
@@ -174,11 +188,13 @@ jest.mock('@bahmni/widgets', () => ({
   useUserPrivilege: jest.fn(),
   useHasPrivilege: jest.fn(),
   useNotification: jest.fn(),
+  usePatientUUID: jest.fn(),
 }));
 
 jest.mock('@bahmni/services', () => ({
   ...jest.requireActual('@bahmni/services'),
   getConfig: jest.fn(),
+  getFormattedPatientById: jest.fn(),
 }));
 
 const mockClinicalConfig = {
@@ -260,6 +276,9 @@ describe('ConsultationPage', () => {
     });
 
     (getConfig as jest.Mock).mockResolvedValue(mockDashboardConfig);
+
+    (usePatientUUID as jest.Mock).mockReturnValue(null);
+    (getFormattedPatientById as jest.Mock).mockResolvedValue(null);
   });
 
   afterEach(() => {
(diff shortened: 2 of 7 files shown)


packages/bahmni-services/src/patientService/patientService.ts
```diff
@@ -76,14 +76,17 @@ const extractAddressExtensions = (
  * @returns A formatted name string
  * @returns null if no name is provided
  */
+export const capitalizeFirstLetter = (str: string): string =>
```
sumaztwcode: nit: `capitalizeFirstLetter` duplicates the existing `capitalize` utility in `utils/utils.ts`. The key difference is that `capitalize` lowercases first then capitalizes (breaking "McDonald" → "Mcdonald"), while this preserves existing casing — which is better for names.

If the case-preserving behavior is intentional, keep it but move it to `utils/utils.ts` alongside the other string utilities, and add a JSDoc explaining why it differs from `capitalize`.

Also: the registration module inlines `split(' ').map(capitalizeFirstLetter).join(' ')` twice — consider a shared helper.

apps/registration/src/hooks/usePatientDetails.ts
```diff
@@ -105,7 +106,10 @@ export const usePatientDetails = ({ patientUuid }: UsePatientDetailsProps) => {
         patientUuid: patientDetails.patient.uuid,
         patientIdentifier:
           patientDetails.patient.identifiers[0]?.identifier ?? '',
-        patientName: patientDetails.patient.person.display ?? '',
+        patientName: (patientDetails.patient.person.display ?? '')
```
sumaztwcode: nit: `person.display` can contain consecutive spaces. `split(' ')` produces empty tokens (`['John', '', 'Doe']`) and the rejoin preserves the double space.

Consider: `.trim().split(/\s+/).map(capitalizeFirstLetter).join(' ')` to normalize whitespace.

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -120,6 +121,11 @@ const ConsultationPage: React.FC = () => {
     [handleSearchOpen, t],
   );
   const viewingForm = useObservationFormsStore((state) => state.viewingForm);
+  const {
+    patient,
+    loading: patientLoading,
+    error: patientError,
+  } = usePatient();
 
   const breadcrumbItems = [
```
sumaztwcode: nit: `breadcrumbItems` is recreated every render. In this same file, `globalActions` and `sidebarItems` are wrapped in `useMemo`. Consider:
```typescript
const breadcrumbItems = useMemo(() => [
  { id: 'home', label: 'Home', href: BAHMNI_HOME_PATH },
  { id: 'clinical', label: 'Clinical', href: BAHMNI_CLINICAL_PATH },
  { id: 'current', label: patient?.fullName ?? t('CURRENT_PATIENT'), isCurrentPage: true },
], [patient?.fullName, t]);
```

apps/clinical/src/pages/__tests__/ConsultationPage.test.tsx
```diff
@@ -298,6 +330,64 @@ describe('ConsultationPage', () => {
...
+        loading: false,
+        error: null,
+        refetch: jest.fn(),
+      });
+
+      renderWithProvider();
+
+      await waitFor(() => {
+        expect(screen.queryByTestId('carbon-loading')).not.toBeInTheDocument();
+      });
```
sumaztwcode, line 377: nit: The breadcrumb fallback tests cover `patient: null, loading: false` and `patient: { fullName }`, but not:
- `loading: true, patient: null` → should show "Current Patient"
- `error: new Error(...), patient: null` → should show "Current Patient"

The AC explicitly requires fallback during loading and on API failure. Adding these two cases would make the coverage match the AC.

packages/bahmni-widgets/src/patientDetails/PatientDetails.tsx
```diff
@@ -1,14 +1,36 @@
...
+}
+
+const PatientDetails: React.FC<PatientDetailsProps> = ({
+  patient: patientProp,
+  loading: loadingProp,
+  error: errorProp,
+}) => {
   const { t } = useTranslation();
-  const { patient, loading, error } = usePatient();
+  const hasExternalData = patientProp !== undefined;
```
sumaztwcode: nit: `hasExternalData = patientProp !== undefined` — the entire behavioral switch (internal fetch vs parent-owned data) relies on the `undefined` vs `null` distinction. The type `patient?: FormattedPatientData | null` doesn't make this clear.

Consider a JSDoc note: "Pass `patient={null}` to signal parent owns the data (disables internal fetch). Omit `patient` entirely for self-fetching mode."

sumaztwcode (approved):
## Review for BAH-4693

Clean implementation — full patient name in breadcrumb with proper fallback, and smart duplicate-fetch prevention via the `usePatient({ enabled })` pattern. Looks good to merge with a few nits.

Found 0 blocking issues, 0 suggestions, and 5 nits.

**Acceptance Criteria**: 6/6 met

### INFO

| File | Type | Note |
|------|------|------|
| PR description | question | Description says "exclude middle name" but code correctly includes all given names (`name.given?.map(...)` joins all). Code is correct per AC — update the description to avoid a future developer "fixing" it. |
| PatientHeader.tsx | question | `PatientHeader` is now a pure pass-through for `patient/loading/error`. Works fine, but if other pages use `<PatientHeader>` without passing these props, they'll trigger a separate `usePatient()` fetch. A context could make the contract more explicit. |
| usePatient.ts:33-55 | question | No cleanup for in-flight fetch when `enabled` toggles to `false`. In current flow this race doesn't trigger since `enabled` is constant, but if the hook is reused with dynamic toggling, consider adding a cancelled-flag cleanup. |

packages/bahmni-services/src/utils/utils.ts
```diff
@@ -74,6 +74,19 @@ export const generateUUID = (): string => {
...
 
+/**
+ * Capitalizes only the first letter of a string, leaving the rest unchanged.
+ * Unlike `capitalize`, this does NOT lowercase the remaining characters first,
+ * which preserves mixed-case names (e.g., "McDonald" stays "McDonald").
+ * Use this for patient names where casing must be preserved.
+ * @param str - The string to capitalize
+ * @returns The string with only its first character uppercased
+ */
+export function capitalizeFirstLetter(str: string): string {
```
rahu1ramesh:
```
p::first-letter {
    text-transform: capitalize;
}
```

is a simpler alternative that you can consider.

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -121,15 +124,27 @@ const ConsultationPage: React.FC = () => {
...
+
+  const breadcrumbItems = useMemo(
+    () => [
+      { id: 'home', label: 'Home', href: BAHMNI_HOME_PATH },
+      { id: 'clinical', label: 'Clinical', href: BAHMNI_CLINICAL_PATH },
+      {
+        id: 'current',
+        label: patient?.fullName
+          ? capitalize(patient.fullName)
+          : t('CURRENT_PATIENT'),
```
rahu1ramesh, line 142: The patient name will always be there for a valid patient uuid, I feel the fall back is not needed, and if patient uuid is wrong, then the page should say that instead, where "CURRENT_PATIENT" might not be right, good to confirm this with the UX.

packages/bahmni-services/src/utils/utils.ts
Kanchi-Oswal-twk: @rahu1ramesh , considered the alternative but this will only capitalize the first letter of first word remaining words will be lower case. So considered the existing Utility function "Capitalize"

apps/clinical/src/pages/ConsultationPage.tsx
ravinderkabli: When a patient is loaded in the app, their full name is built by combining the given name and family name from the FHIR data. However, there are  
  situations where this name might not be available — for example, the patient data might still be loading, or the patient's record in OpenMRS might
   simply not have a name filled in. Because of this, patient.fullName can be null in perfectly valid scenarios. The code handles this gracefully by
   checking if the name exists before displaying it — if it does, it shows the actual patient name in the breadcrumb; if not, it falls back to the  
  generic "Current Patient" label so the UI never breaks or shows something empty.    

It is good to have "Current Patient" as a fallback

narmadhaTW: @rahu1ramesh are we good ? Could u pls resolve the conversation
.

rahu1ramesh: @ravinderkabli I am not sure what your source here is, but to the best of my knowledge name is mandatory. https://talk.openmrs.org/t/requirement-to-hide-patient-name/23135/2

And I don't see how "Current Patient" is gracefully handling this, feels like a miss to me. When loading a skeleton (loading text) should be shown while in case of an invalid patient the page itself should redirect to 404 or something similar (confirm with UX).

"Current Patient" was a shortcut we took to get started with, no point in retaining this.

@narmadhaTW openmrs model requires name [(first +/ middle +/ last) atleast] there is no need for a fallback in my view. Please check with Angshu / Mohan and decide before proceeding.

rahu1ramesh: + @narmadhaTW I have added a comment, won't block the PR but will leave this to your discretion whether or not to proceed, but I am not resolving this conversation until I am convinced.

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -121,15 +124,27 @@ const ConsultationPage: React.FC = () => {
...
-    { id: 'home', label: 'Home', href: BAHMNI_HOME_PATH },
-    {
-      id: 'clinical',
-      label: 'Clinical',
-      href: BAHMNI_CLINICAL_PATH,
-    },
-    { id: 'current', label: t('CURRENT_PATIENT'), isCurrentPage: true },
-  ];
+  const patientUUID = usePatientUUID();
+  const { data: patient } = useQuery({
```
rahu1ramesh, line 128: We are already making this call for the patient details widget, can we instead think of a patient details provider that we can use within the page?

@angshu @PratitiP @mohan-13

apps/clinical/src/pages/ConsultationPage.tsx
ravinderkabli: Yes, the name(First Name, Last Name) are mandatory from frontend. But the same check is not present in openmrs.
I am able to create a patient without a last name/family name. Adding details on the payload below:

https://docker.standard.mybahmni.in/openmrs/ws/fhir2/R4/Patient

{
  "resourceType": "Patient",
  "identifier": [
    {
      "use": "official",
      "value": "ABC200099",
      "type": {
        "coding": [
          {
            "code": "d3153eb0-5e07-11ef-8f7c-0242ac120002"
          }
        ],
        "text": "Patient Identifier"
      },
      "extension": [
        {
          "url": "http://fhir.openmrs.org/ext/patient/identifier#location",
          "valueReference": {
            "reference": "Location/b5da9afd-b29a-4cbf-91c9-ccf2aa5f799e"
          }
        }
      ]
    }
  ],
  "name": [
    {
      "given": [
        "Sammy"
      ]
    }
  ],
  "gender": "male",
  "birthDate": "1981-05"
}

Like I said above, 
The OpenMRS REST API (/ws/rest/v1/person) permits familyName to be null, which means concatenating givenName + " " + familyName directly in the breadcrumb can produce "null Doe", "John ", or a completely blank label depending on what the API returns or whether the call fails entirely.     
Displaying "Current Patient" as a fallback is the safer choice here — gives the clinician a clear signal that a  patient is loaded even when the name isn't resolvable, and handles both the loading state and the null/error state consistently under one behaviour.

Hope this helps.

apps/clinical/src/pages/ConsultationPage.tsx
ravinderkabli: @rahu1ramesh - Yeah, that's a fair point! Just to clarify though — React Query will deduplicate the fetch since we're using the same ['patient', patientUUID] key, so no extra network call is happening here.                                                                                                  

That said, we don't have a PatientProvider yet, so building one would be new work that can be taken in separate JIRA improvement card rather than scope-creeping this PR . Also, It would also need refactoring to PatientDetails widget to consume from context instead of fetching on its own.    

Please create a JIRA card and share it with product team.

narmadhaTW (approved):

PR #406, opened 2026-05-13
BAH 4701 | Encounter Match Decision Mapper with Diagnostic Reasons

sahupadma:
JIRA → https://bahmni.atlassian.net/browse/BAH-4701                                                                                                                                                           

  Description                                                                                                                                                                  
   
  Adds a structured encounter match decision mapper that explains why an encounter does or does not match the current clinician session. Previously useEncounterSession        
  returned a binary match/no-match result; now it returns a MatchReasonCode[] (MATCHED, SESSION_EXPIRED, PROVIDER_MISMATCH, LOCATION_MISMATCH, NO_ACTIVE_VISIT,              
  NO_ACTIVE_ENCOUNTER) that downstream widgets can use to display contextual messages and gate edit actions. No new API endpoints — the decision is derived entirely from      
  existing FHIR search results.                                                                                                                                              

  ---
  Key Features
              
  - New resolveEncounterMatchDecision() service function returns a structured EncounterMatchDecision with matched, encounter, and reasons[].
  - useEncounterSession now accepts locationUUID (sourced from the login session cookie) and exposes matchReason: MatchReasonCode[] alongside the existing editActiveEncounter 
  boolean.                                                                                                                                                                     
  - PatientHeader preloads the encounter session decision on page load and surfaces it as data-match-reason / data-can-edit-encounter DOM attributes, so downstream widget     
  consumers (Allergies, Conditions, Medications) can read the decision without waiting for ConsultationPad to open.                                                            
  - Location mismatch is detected client-side by inspecting encounter.location[].location.reference from the FHIR response — no backend changes required.                    
                                                                                                                                                                               
  ---                                                                                                                                                                        
  Implementation Details                                                                                                                                                       
                                                                                                                                                                               
  - UI Components
  PatientHeader.tsx — calls useEncounterSession on mount and writes data-match-reason / data-can-edit-encounter to the header element. No visible UI change; this is a data    
  contract for widget consumers.                                                                                                                                               
  - State Management
  useEncounterSession gains a new matchReason: MatchReasonCode[] state field. Existing hasActiveSession, editActiveEncounter, and isPractitionerMatch fields are preserved     
  unchanged for backward compatibility.                                                                                                                                        
  - Custom Hooks
  useEncounterSession — updated to call resolveEncounterMatchDecision instead of findActiveEncounterInSession. Uses isOwnInSessionEncounter(decision) helper to derive         
  sessionExists, keeping the interpretation logic co-located with the mapper.                                                                                                  
  - API Integration
  Three parallel FHIR GET /Encounter searches run concurrently via Promise.all:                                                                                                
    - own provider + session window → detects MATCHED / LOCATION_MISMATCH                                                                                                      
    - own provider + all time → detects SESSION_EXPIRED                                                                                                                        
    - any provider + session window → detects PROVIDER_MISMATCH                                                                                                                
                                                                                                                                                                               
  findActiveEncounterInSession() is retained unchanged for backward compatibility with existing consumers.                                                                                                          
  - Accessibility                                                                                                                                                              
  No user-facing UI changes. Existing toHaveNoViolations assertion in PatientHeader.test.tsx continues to pass.                                                              
 

## Summary by CodeRabbit

* **New Features**
  * Added localized encounter-match reason messages (English & Spanish) and surfaced match reason codes for encounter outcomes.
  * Patient header now exposes consultation match reason and edit-capability indicators to the UI.

* **Refactor**
  * Encounter-session flow reorganized to produce explicit match reasons and resumability; consultation action receives session/edit/loading state from its parent.

* **Tests**
  * Expanded unit tests covering encounter-matching decisions, hook behavior, and header/button data attributes.

<!-- review_stack_entry_start -->

[![Review Change Stack](https://storage.googleapis.com/coderabbit_public_assets/review-stack-in-coderabbit-ui.svg)](https://app.coderabbit.ai/change-stack/Bahmni/bahmni-apps-frontend/pull/406)

<!-- review_stack_entry_end -->


packages/bahmni-services/src/index.ts
```diff
@@ -221,7 +221,15 @@ export {
   type Provider,
   type Person,
 } from './providerService';
-export { findActiveEncounterInSession } from './encounterSessionService';
+export {
+  findActiveEncounterInSession,
+  resolveEncounterMatchDecision,
+  isOwnInSessionEncounter,
```
sumaztwcode: What does this variable mean? isOwnInSessionEncounter?

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
```diff
@@ -0,0 +1,212 @@
+import { Encounter } from 'fhir/r4';
+import { getActiveVisit } from '../encounterService';
+import {
+  searchEncounters,
+  getEncounterSessionDuration,
+} from './encounterSessionService';
+
+export type MatchReasonCode =
```
sumaztwcode: you may end up using these types and constant variables below when you have edit scenarios as condition checks will vary. it is better to place these in constants file

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
```diff
@@ -0,0 +1,212 @@
...
+
+function checkLocationMatch(
+  encounter: Encounter,
+  loginLocationUUID: string,
+): boolean {
+  const encounterLocations = encounter.location ?? [];
+  return encounterLocations.some(
+    (loc) => getReferenceId(loc.location?.reference) === loginLocationUUID,
+  );
+}
```
sumaztwcode, line 32: blocking: When `getUserLoginLocation()` throws in `useEncounterSession`, `locationUUID` defaults to `''`. Here, `checkLocationMatch` compares `getReferenceId(loc.location?.reference) === ''`, which can never be true — so any in-session own encounter that **should** match gets classified as `LOCATION_MISMATCH` instead of `MATCHED`.

A clinician whose location cookie is temporarily unavailable (cleared, SSO re-auth) sees "You started an encounter at a different location" when the real issue is a missing cookie.

Suggested fix — skip location check when location is unknown:
```typescript
function checkLocationMatch(
  encounter: Encounter,
  loginLocationUUID?: string,
): boolean {
  if (!loginLocationUUID) return true; // unknown → skip check
  const encounterLocations = encounter.location ?? [];
  return encounterLocations.some(
    (loc) => getReferenceId(loc.location?.reference) === loginLocationUUID,
  );
}
```
Also add a mapper test for `locationUUID=''` to document the intended behavior.

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
```diff
@@ -0,0 +1,212 @@
...
+export function isOwnInSessionEncounter(
+  decision: EncounterMatchDecision,
+): boolean {
+  return (
+    decision.matched ||
+    (decision.reasons.includes('LOCATION_MISMATCH') &&
+      !decision.reasons.includes('SESSION_EXPIRED') &&
+      !decision.reasons.includes('PROVIDER_MISMATCH'))
+  );
+}
```
sumaztwcode, line 193: suggestion: `isOwnInSessionEncounter` returns `true` for `LOCATION_MISMATCH`-only decisions, which sets `editActiveEncounter=true` in the hook.

But the i18n string `ENCOUNTER_MATCH_REASON_LOCATION_MISMATCH` says: **"this will create a new one"** — contradicting `editActiveEncounter=true`.

The code says "can edit" while the message says "will create new." When a downstream widget uses `data-can-edit-encounter="true"` to enable editing, the user sees a message about creating a new encounter but is actually editing an existing one.

Please reconcile:
- If LOCATION_MISMATCH should allow editing → change i18n to "resuming the existing consultation"
- If LOCATION_MISMATCH should NOT allow editing → change `isOwnInSessionEncounter` to return `false`

apps/clinical/src/components/patientHeader/PatientHeader.tsx
```diff
@@ -19,12 +23,22 @@ const PatientHeader: React.FC<PatientHeaderProps> = ({
   isActionAreaVisible,
 }) => {
   const { t } = useTranslation();
+  const { practitioner } = useActivePractitioner();
+
+  const { matchReason, editActiveEncounter } = useEncounterSession({
```
sumaztwcode: suggestion: `ConsultationActionButton` (child rendered at line 44) already calls `useActivePractitioner()` + `useEncounterSession()` independently. This adds a second invocation in the parent — both hooks fire `resolveEncounterMatchDecision` independently, resulting in **6 FHIR searches** (3×2) on every mount instead of 3.

The two independent hook instances also have separate state, so they can briefly show different loading/resolved states.

Consider:
1. Call `useEncounterSession` only here and pass `matchReason`/`editActiveEncounter` as props to `ConsultationActionButton`
2. Or share state via React Context

apps/clinical/src/hooks/useEncounterSession.ts
```diff
@@ -1,61 +1,61 @@
...
-      return;
-    }
-
-    // Get practitioner UUID for session filtering
-    const practitionerUUID = practitioner?.uuid;
+  let locationUUID = '';
+  try {
+    locationUUID = getUserLoginLocation().uuid;
+  } catch {
+    // location cookie unavailable
```
sumaztwcode: suggestion: `getUserLoginLocation()` is called in the hook body (outside `useEffect`), so it re-runs on every render. The result is captured by `fetchSessionState`'s closure but `locationUUID` is not in the `useEffect` dependency array (suppressed by `eslint-disable`).

If the login location changes between renders, the effect won't re-fire — using a stale value.

Moving this inside `fetchSessionState()` would:
1. Only run when the async fetch triggers (not every render)
2. Always capture a fresh value per-fetch
3. Not need to be in the dep array

apps/clinical/src/components/patientHeader/PatientHeader.tsx
```diff
@@ -19,12 +23,22 @@ const PatientHeader: React.FC<PatientHeaderProps> = ({
...
 
   return (
     <div
       aria-label={t('PATIENT_HEADER_LABEL')}
       className={styles.header}
       data-testid="patient-header"
+      data-match-reason={
+        matchReason.length > 0 ? matchReason.join(',') : undefined
+      }
+      data-can-edit-encounter={editActiveEncounter}
```
sumaztwcode: nit: `editActiveEncounter` is boolean, but React renders `false` as the string `"false"` on DOM attributes. Any consumer doing `if (element.dataset.canEditEncounter)` gets `true` because `"false"` is truthy in JS.

Consider: `data-can-edit-encounter={editActiveEncounter ? 'true' : undefined}` — omit when false, so consumers check attribute presence.

apps/clinical/src/hooks/useEncounterSession.ts
```diff
@@ -64,54 +64,48 @@ export function useEncounterSession(
     setError(null);
 
     try {
-      // Find active encounter for current patient and practitioner
-      const activeEncounter = await findActiveEncounterInSession(
+      const decision = await resolveEncounterMatchDecision(
         patientUUID,
         practitionerUUID,
-        undefined,
+        locationUUID,
```
sumaztwcode: nit: No abort/cleanup for the `resolveEncounterMatchDecision` promise. When deps change rapidly (navigating between patients), the stale promise resolves and calls setState, potentially overwriting the new patient's state.

Consider adding an ignore flag:
```typescript
useEffect(() => {
  let ignore = false;
  const run = async () => {
    // ... existing logic
    if (!ignore) { /* setState */ }
  };
  run();
  return () => { ignore = true; };
}, [...]);
```

sumaztwcode (changes requested):
## Review for BAH-4701

Well-structured encounter match decision mapper with excellent test coverage (750+ lines). The architecture cleanly separates service logic from hook and UI layers. A few issues to address before merge.

Found 1 blocking issue, 3 suggestions, and 3 nits.

**Acceptance Criteria**: 7/9 met (location mismatch behavior has a bug + semantic contradiction)

### What's Done Well
1. Clean separation: mapper → hook → UI data attributes
2. Thorough tests covering all reason codes, combinations, error handling
3. Backward compat: `findActiveEncounterInSession` retained unchanged
4. `MatchReasonCode[]` as array allows reason combinations — good modeling
5. Consistent i18n in both English and Spanish

### Issues not in diff

| File:Line | Type | Issue |
|-----------|------|-------|
| encounterMatchDecisionMapper.ts:64 | nit | `getActiveVisit` and `getEncounterSessionDuration` are independent but awaited serially — could run in parallel via `Promise.all` to save one network round-trip |

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
```diff
@@ -0,0 +1,212 @@
...
+        encounter: null,
+        reasons: ['NO_ACTIVE_ENCOUNTER'],
+      };
+    }
+
+    const sessionDuration = await getEncounterSessionDuration();
+    const sessionStartTime = new Date(Date.now() - sessionDuration * 60 * 1000);
+    const lastUpdatedParam = `ge${sessionStartTime.toISOString()}`;
+
+    const [inSessionOwn, allTimeOwn, inSessionAny] = await Promise.all([
```
sumaztwcode: from here onwards, try to make the logic more understandable by using meaningful variables - e.g currentPractitionerRecentEncounters/ otherProvidersRecentEncounters/ recentEncountersInVisit/ practitionerEncountersAllTime

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    // PROVIDER_MISMATCH = found in-session for any provider but NOT this practitioner
+    const otherProviderEncounters = inSessionAnyInVisit.filter(
+      (e) => !inSessionOwnIds.has(e.id),
+    );
+
+    const reasons: MatchReasonCode[] = [];
+    let primaryEncounter: Encounter | null = null;
+
+    if (inSessionOwnInVisit.length > 1) {
```
sumaztwcode: We should be able to explain the logic flow as lot of checks are in place here. Say, consider this ->
  1. Active visit? → NO → return NO_ACTIVE_VISIT (which you are doing initially)
  2. NO_ACTIVE_ENCOUNTER 
  3. Current practitioner recent encounter? → YES → check location → MATCHED or LOCATION_MISMATCH (post this, all other mismatch scenarios)                              
  4. Multiple encounters? → return MULTIPLE_ENCOUNTERS_FOUND                                     
  5. Other providers active? → PROVIDER_MISMATCH                                                                                
  6. Practitioner had old encounter? → SESSION_EXPIRED                                                                                                                                                                  
  7. Add LOCATION_MISMATCH if applicable

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
```diff
@@ -0,0 +1,212 @@
...
+  );
+}
+
+export async function resolveEncounterMatchDecision(
+  patientUUID: string,
+  practitionerUUID: string,
+  locationUUID: string,
+  encounterTypeUUID?: string,
+): Promise<EncounterMatchDecision> {
+  try {
```
sumaztwcode, line 49: This is a simplified version - please compare against current implementation. I am retaining the comments, but these can be removed later.
    try {
      // 1. Check active visit exists
      const activeVisit = await getActiveVisit(patientUUID);
      if (!activeVisit?.id) {
        return { matched: false, encounter: null, reasons: ['NO_ACTIVE_VISIT'] };
      }

      // 2. Get session window
      const sessionDuration = await getEncounterSessionDuration();
      const sessionStartTime = new Date(Date.now() - sessionDuration * 60 * 1000);
      const recentUpdatedParam = `ge${sessionStartTime.toISOString()}`;

      // 3. Search encounters
      const [recentEncounters, allPractitionerEncounters] = await Promise.all([
        searchEncounters({
          patient: patientUUID,
          _tag: 'encounter',
          _lastUpdated: recentUpdatedParam,
          type: encounterTypeUUID,
        }),
        searchEncounters({
          patient: patientUUID,
          _tag: 'encounter',
          participant: practitionerUUID,
          type: encounterTypeUUID,
        }),
      ]);

      // 4. Filter by current visit
      const recentEncountersInVisit = filterEncountersByVisit(recentEncounters, activeVisit.id);
      const practitionerEncountersAllTime = filterEncountersByVisit(allPractitionerEncounters, activeVisit.id);

      // 5. Check if any encounters exist in this visit
      if (recentEncountersInVisit.length === 0 && practitionerEncountersAllTime.length === 0) {
        return { matched: false, encounter: null, reasons: ['NO_ACTIVE_ENCOUNTER'] };
      }

      // 6. Categorize encounters
      const currentPractitionerRecentEncounters = recentEncountersInVisit.filter(
        (e) => getReferenceId(e.participant?.[0]?.individual?.reference) === practitionerUUID
      );
      const otherProvidersRecentEncounters = recentEncountersInVisit.filter(
        (e) => getReferenceId(e.participant?.[0]?.individual?.reference) !== practitionerUUID
      );

      // 7. Check for multiple encounters by same practitioner
      if (currentPractitionerRecentEncounters.length > 1) {
        return {
          matched: false,
          encounter: currentPractitionerRecentEncounters[0],
          reasons: ['MULTIPLE_ENCOUNTERS_FOUND'],
        };
      }

      // 8. Check for matching encounter by this practitioner
      if (currentPractitionerRecentEncounters.length === 1) {
        const encounter = currentPractitionerRecentEncounters[0];
        if (checkLocationMatch(encounter, locationUUID)) {
          return { matched: true, encounter, reasons: ['MATCHED'] };
        }
        return { matched: false, encounter, reasons: ['LOCATION_MISMATCH'] };
      }

      // 9. No encounter by this practitioner in session - check other conditions
      const reasons: MatchReasonCode[] = [];
      let primaryEncounter: Encounter | null = null;

      if (otherProvidersRecentEncounters.length > 0) {
        // Other providers are actively working on this patient
        reasons.push('PROVIDER_MISMATCH');
        primaryEncounter = otherProvidersRecentEncounters[0];
      } else if (practitionerEncountersAllTime.length > 0) {
        // Practitioner had an encounter but it's outside session window
        reasons.push('SESSION_EXPIRED');
        primaryEncounter = practitionerEncountersAllTime[0];
      }

      // 10. Check location for fallback encounters
      if (primaryEncounter && !checkLocationMatch(primaryEncounter, locationUUID)) {
        reasons.push('LOCATION_MISMATCH');
      }

      return { matched: false, encounter: primaryEncounter, reasons };
    } catch (error) {
      console.error(
        'Error in resolveEncounterMatchDecision:',
        error instanceof Error ? error.message : error,
      );
      return { matched: false, encounter: null, reasons: ['NO_ACTIVE_ENCOUNTER'] };
    }
  }

apps/clinical/src/hooks/useEncounterSession.ts
```diff
@@ -1,61 +1,61 @@
...
-    const practitionerUUID = practitioner?.uuid;
+  let locationUUID = '';
+  try {
+    locationUUID = getUserLoginLocation().uuid;
+  } catch {
+    // location cookie unavailable
+  }
 
-    if (!practitionerUUID || !encounterTypeUUID) {
+  const fetchSessionState = async () => {
```
sumaztwcode: changes introduced in this, need to be tested thoroughly - for places where this API is used earlier, to ensure that no breaking changes exist

apps/clinical/src/hooks/useEncounterSession.ts
```diff
@@ -64,54 +64,48 @@ export function useEncounterSession(
...
         practitionerUUID,
-        undefined,
+        locationUUID,
         encounterTypeUUID,
       );
-      const sessionExists = activeEncounter !== null;
 
-      // Since server filters by practitioner, if we get an encounter, it belongs to current practitioner
-      const practitionerMatches = sessionExists;
+      const sessionExists = isOwnInSessionEncounter(decision);
```
sumaztwcode: Name meaningfully even if variable name is long

sumaztwcode (changes requested):

apps/clinical/src/hooks/useEncounterSession.ts
sahupadma: renamed

apps/clinical/src/hooks/useEncounterSession.ts
sahupadma: yes tested

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
sahupadma: done

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
sahupadma: done

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
sahupadma: done

apps/clinical/src/hooks/useEncounterSession.ts
sahupadma: created ignored flag

apps/clinical/src/components/patientHeader/PatientHeader.tsx
sahupadma: done

apps/clinical/src/hooks/useEncounterSession.ts
sahupadma: changed

apps/clinical/src/components/patientHeader/PatientHeader.tsx
sahupadma: done point 1

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
sahupadma: changed i18n

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
sahupadma: done

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
sahupadma: moved to constant file

packages/bahmni-services/src/index.ts
sahupadma: Returns true when the decision represents an encounter owned by the current
   practitioner that is still within the session window — meaning the clinician
 can resume/edit it rather than starting a new consultation.
 LOCATION_MISMATCH alone qualifies because the encounter belongs to this
practitioner in this session; only the login location has changed.  i have renamed it to canResumeOwnInSessionEncounter

sumaztwcode (approved):

PR #411, opened 2026-05-18
BAH-4654 | Add. Edit Medications feature for the Medications Widget

ROHIT-V-TW:
## Summary

Implements the ability for clinicians to edit active and scheduled medications directly from the patient dashboard within an active visit session. Follows the same architectural pattern established by Edit Allergies (PR #421).

### Edit Flow
- **Row-level edit** — per-row edit action in the overflow menu (`Actions` component); visible only for medications belonging to the active encounter
- **Edit form** — opens ConsultationPad scoped to medications only (`editOnly: 'medication'`); pre-populates the MedicationRequestForm with parsed FHIR MedicationRequest data
- **Done button gating** — disabled until user actually modifies a field; re-enables on change; **re-disables on revert** (snapshot-based comparison replaces one-way `isModified` flag)
- **Save routing** — MATCHED encounter → PUT to existing; SESSION_EXPIRED / LOCATION_MISMATCH / PROVIDER_MISMATCH → creates new encounter with `priorPrescription` reference

### Event & Data Flow
1. User clicks "Edit" (row) → `actionHandlers.ts` dispatches `startConsultation` custom event with `editMedications`, `editOnly: 'medication'`, `editEncounterUuid`
2. `ConsultationPad` receives the event → fetches the encounter via `getEncounterByUuid()` with `AbortController` → seeds `MedicationRequestStore.setPendingFhirEdits()`
3. `MedicationRequestForm` detects `pendingFhirEdits` → parses them via `parseFhirToMedicationInputEntry()` in a `useMemo` → loads into store via `loadMedicationsForEdit()` in a single `useEffect`
4. `SelectedMedicationRequestItem` renders pre-populated fields; `fhirResourceId` guards prevent mount-time effects from marking entries as modified
5. On "Done" → `consultationBundleService` builds FHIR bundle with `priorPrescription` reference → POST/PUT to server

### Action Buttons Logic
- `MedicationsTable` reads `config.actions` from dashboard JSON config — actions array is config-driven, not hardcoded
- `editAction = actions.find(a => a.type === 'edit')` — presence of edit action in config enables the feature
- `canEdit = !!editAction && hasPrivilege(userPrivileges, editAction.requiredPrivilege)` — privilege-gated
- `editableMedications` — filtered from `activeAndScheduledMedications` by encounter UUID match: only medications from the active encounter show edit actions
- **Icon vs text rendering** — actions with mapped icons (e.g., `edit` → pencil icon) render as `IconButton`; actions without icons (e.g., `administer`) render as text `Button`, preserving existing behavior for the Vaccinations widget

### Architecture
- **Props over store** — `MedicationsTable` receives `canEditOrCreate` and `activeEncounterUuid` as props from `DashboardSection` (following PR #421 pattern); no direct store consumption in widget layer since module-level singletons are duplicated across separately-bundled packages
- **Encounter session store** — reverted `globalThis` mutation; store stays as module-level singleton; data flows via props across bundles
- **Reusable MedicationsTable** — the same `treatment` widget type is used for both Medications and Vaccinations sections; edit behavior is entirely config-driven via `actions` array — no code changes needed to add new action types in the future
- **Shared constants** — `MEDICATIONS_INPUT_CONTROL_KEY` moved to `@bahmni/services` to eliminate implicit string coupling between `apps/clinical` and `@bahmni/widgets`
- **Named service function** — `getEncounterByUuid()` extracted to `@bahmni/services/encounterService` with `AbortController` support for true request cancellation on unmount
- **Layer inversion fix** — `DURATION_UNIT_OPTIONS` moved from `components/forms/` to `constants/medications.ts`
- **Snapshot-based change detection** — replaced `isModified` one-way flag with `originalEditSnapshots` Map + `hasMedicationChanged()` field-by-field comparison (dosage, dosageUnit, frequency, route, duration, durationUnit, instruction, isSTAT, isPRN, startDate, dispenseQuantity, dispenseUnit, note); Done button now correctly re-disables when user reverts all changes
- **fhirResourceId guard** — dispense quantity auto-calc effect skips FHIR-loaded edit entries, preventing false `isModified` on mount
- **useMemo + single useEffect** — replaced two separate useEffects in MedicationRequestForm with `useMemo` for parsing + single `useEffect` for loading, removing refs and double-effect complexity

### FHIR Parsing
- New `parseFhirToMedicationInputEntry()` — converts FHIR MedicationRequest back to `MedicationInputEntry` form model for editing
- Handles: STAT, PRN, contained medications, dose forms, dispense quantities/units, notes, start dates from `boundsPeriod` / `authoredOn` / `extension`
- Resolves concepts (dosageUnit, frequency, route, instruction, durationUnit, dispenseUnit) against `MedicationConfig` by matching UUID or name

### Bug Fixes (Latest Round)
- **All tab date sorting** — date groups were unsorted because `new Date(formattedString)` returned `NaN` for strings like "Today"; fixed to sort by raw `orderDate` ISO string from group items
- **Start date edit detection** — `hasMedicationChanged()` was missing `startDate` comparison, so date-only edits didn't enable the Done button
- **Start date revert detection** — used `toDateString()` instead of `getTime()` for comparison, since FHIR dates have time components but the DatePicker returns midnight; same-day re-selection now correctly detects "no change"
- **Administer action icon** — removed `<Edit />` fallback for unknown action types; actions without mapped icons (like `administer`) now render as text `Button` matching the deployed standard behavior
- **Duplicate jest.mock** — consolidated two `jest.mock('@bahmni/services')` calls in `ImmunizationForm.integration.test.tsx` that caused the second to silently override the first, leading to test timeouts

### Locale
- Added: `MEDICATIONS_EDIT_FORM_TITLE`, `MEDICATIONS_EDIT_ACTION`, `MEDICATIONS_ACTIONS_MENU_LABEL` (EN + ES)

## Related PRs
| Repo | PR | Purpose |
|------|----|---------|
| bahmni-module-fhir2-addl-extension | [#71](https://github.com/Bahmni/bahmni-module-fhir2-addl-extension/pull/71) | `priorPrescription` → `REVISE` action translation in FHIR2 module |
| standard-config | [#119](https://github.com/Bahmni/standard-config/pull/119) | Dashboard config: adds `actions: [{type: "edit"}]` to treatment control |
| clinic-config | [#253](https://github.com/Bahmni/clinic-config/pull/253) | Same config changes for clinic deployment |

## Test Plan
- [x] Verify row-level edit action appears only for medications in the active encounter
- [x] Verify edit form pre-populates with correct medication data (dosage, frequency, route, duration, instructions, notes)
- [x] Verify "Done" button is disabled until a field is changed
- [x] Verify "Done" button re-disables when all changes are reverted to original values
- [x] Verify start date change to future enables Done; re-selecting original date disables it
- [x] Verify STAT medications preserve their original start date on edit
- [x] Verify save creates new MedicationRequest with `priorPrescription` reference
- [x] Verify edit works across encounter session states (MATCHED, SESSION_EXPIRED, LOCATION_MISMATCH)
- [x] Verify no edit button appears without `Edit Orders` privilege
- [x] Verify Vaccinations section "Administer" action renders as text button (not Edit icon)
- [x] Verify All tab date groups are sorted descending (most recent first)


packages/bahmni-widgets/src/medications/MedicationsTable.tsx
```diff
@@ -232,6 +242,116 @@ const MedicationsTable: React.FC<WidgetProps> = ({
...
+      const el = document.querySelector(HEADER_SELECTOR);
+      if (el) {
+        bodyObserver.disconnect();
+        readAttributes();
+        const attrObserver = observeHeader(el);
+        // Store for cleanup — overwrite returned cleanup
+        cleanupRef.current = () => attrObserver.disconnect();
+      }
+    });
+    bodyObserver.observe(document.body, { childList: true, subtree: true });
```
sumaztwcode: blocking: `bodyObserver.observe(document.body, { childList: true, subtree: true })` fires on **every DOM mutation site-wide** until the header appears. If the header never mounts, this observer lives for the component's lifetime.

Additionally, `cleanupRef` is a plain object (not `useRef`), and its `.current` is overwritten asynchronously inside the `bodyObserver` callback — this races with component unmount, potentially leaking the `attrObserver`.

The codebase already has `consultationSaved` (app→widget via `@bahmni/services`) as the established event pattern. Consider adding a `consultationSessionChanged` event instead of DOM attribute polling. If DOM reading is needed, at minimum: use `useRef` for cleanup, add a timeout/mutation-count cap on the bodyObserver, and disconnect it explicitly after finding the header.

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -108,10 +118,36 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
...
     practitioner,
     encounterTypeUUID: selectedEncounterType?.uuid,
   });
+
+  const editEncounterUuid = encounterSessionStartContext.editEncounterUuid as
+    | string
+    | undefined;
+  const [editEncounter, setEditEncounter] = useState<Encounter | null>(null);
+  const [editEncounterLoading, setEditEncounterLoading] = useState(false);
+  useEffect(() => {
```
sumaztwcode, line 173: blocking: This `useEffect` fetches encounter data but has no `AbortController` or cleanup. If the user opens edit mode and immediately cancels (closes ConsultationPad) while the fetch is in-flight, `setEditEncounter`, `setEditEncounterLoading`, and `addNotification` fire on an unmounted component.

Suggested fix: add an `AbortController`, pass its signal to the fetch, and abort in the cleanup return.

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
```diff
@@ -116,17 +116,10 @@ export async function resolveEncounterMatchDecision(
...
-        reasons: ['MULTIPLE_ENCOUNTERS_FOUND'],
-      };
-    }
-
-    // 8. One recent encounter by this practitioner → MATCHED or LOCATION_MISMATCH
-    if (currentPractitionerRecentEncounters.length === 1) {
+    // 7–8. One or more recent encounters by this practitioner → MATCHED or LOCATION_MISMATCH
+    //       When multiple exist, pick the most recent (index 0) — they all belong
+    //       to this practitioner within the session window.
+    if (currentPractitionerRecentEncounters.length >= 1) {
```
sumaztwcode, line 121: blocking: Changed from `> 1` (MULTIPLE_ENCOUNTERS_FOUND, matched: false) to `>= 1` (silently picks `[0]`). Multiple encounters within one session window for one practitioner is anomalous — the old code surfaced this as an explicit signal. The new code silently picks an encounter that may not be the most recent (depends on FHIR search sort order, which isn't asserted here).

This is a behavioral change beyond Edit Medications scope — it affects all encounter session matching.

Suggested fix: at minimum log a warning when `length > 1`, verify/assert the array is sorted by recency, and document why MULTIPLE_ENCOUNTERS_FOUND was removed.

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -226,22 +262,52 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
...
-  const enablePrimaryButton = useMemo(
-    () =>
-      hasError ||
-      !isEncounterDetailsFormReady ||
-      isSubmitting ||
-      !hasConsultationData,
-    [hasError, isEncounterDetailsFormReady, isSubmitting, hasConsultationData],
-  );
+  const isEditMode = !!editOnlyKey;
+  const editChangesExist = useMedicationStore((state) => {
```
sumaztwcode: suggestion: This inline selector re-implements the identical 13-field comparison that already exists in `medicationsStore.ts` as `hasMedicationChanged()` + `hasEditChanges()`. The store method is never called — it's dead code. Future field additions need updating in two places.

Suggested fix: replace with `useMedicationStore((state) => !isEditMode || state.hasEditChanges())`.

apps/clinical/src/components/forms/medications/SelectedMedicationItem.tsx
```diff
@@ -194,6 +203,7 @@ const SelectedMedicationItem: React.FC<SelectedMedicationItemProps> =
         if (isSTAT) {
           updateStartDate(id, getTodayDate());
         }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
sumaztwcode: suggestion: Both `eslint-disable` comments in this file (here and on the `setDefaultInstruction` effect) suppress a real missing dependency — `medicationInputEntry.fhirResourceId` is read inside the effect body as a guard but not listed in the deps array.

If `fhirResourceId` ever changes after mount, the guard evaluates a stale value. Adding it to the deps array removes the need for the suppression and makes the intent explicit.

apps/clinical/src/utils/fhir/medicationRequestParser.ts
```diff
@@ -0,0 +1,200 @@
...
+  return {
+    id: resourceId,
+    fhirResourceId: fhirMedRequest.id ?? undefined,
+    medication,
+    display: displayName,
+    dosage,
+    dosageUnit,
+    frequency,
+    instruction,
+    route,
```
sumaztwcode, line 95: suggestion: `dispenseUnit` resolves against `medicationConfig.doseUnits`, but `MedicationConfig` has a separate `dispensingUnits` field. If the system is configured with different concept lists for dosing vs. dispensing (valid in OpenMRS), dispense units not also in dose units silently resolve to null.

Suggested fix: resolve against `medicationConfig.dispensingUnits ?? medicationConfig.doseUnits`. Update `mockConfig` in tests to use non-empty `dispensingUnits`.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
```diff
@@ -232,6 +242,116 @@ const MedicationsTable: React.FC<WidgetProps> = ({
...
+
+    const cleanupRef = { current: () => bodyObserver.disconnect() };
+    return () => cleanupRef.current();
+  }, []);
+
+  const { canEditEncounter, activeEncounterUuid, activePractitionerUuid } =
+    editContext;
+
+  const editableMedications = useMemo(() => {
```
sumaztwcode, line 245: suggestion: This filters `activeAndScheduledMedications` by encounter UUID and practitioner UUID — authorization/ownership logic that belongs in the app layer, not a shared display widget. The widget is re-deriving the same rule that PatientHeader already resolved.

Consider passing editable medication IDs or a callback from the app layer instead of computing ownership inside the widget.

apps/clinical/src/stores/medicationsStore.ts
```diff
@@ -32,8 +35,40 @@ export interface MedicationState {
...
+    current.route?.uuid !== original.route?.uuid ||
+    current.duration !== original.duration ||
+    current.durationUnit?.code !== original.durationUnit?.code ||
+    current.instruction?.uuid !== original.instruction?.uuid ||
+    current.isPRN !== original.isPRN ||
+    current.isSTAT !== original.isSTAT ||
+    current.dispenseQuantity !== original.dispenseQuantity ||
+    current.dispenseUnit?.uuid !== original.dispenseUnit?.uuid ||
+    (current.note ?? '') !== (original.note ?? '') ||
+    current.startDate?.toDateString() !== original.startDate?.toDateString()
```
sumaztwcode: suggestion: `toDateString()` is locale/timezone-sensitive and ignores time components. For STAT medications, `parseStartDate()` returns `new Date()` (current timestamp), so re-parsing on edit produces a different time — but `toDateString()` hides the difference. The same issue exists in the duplicate inline comparison in `consultationPad/index.tsx`.

Suggested fix: use `date-fns` `isSameDay()` (already available in the project) or compare via `getTime()` for exact equality.

apps/clinical/src/utils/fhir/medicationRequestParser.ts
```diff
@@ -0,0 +1,200 @@
...
+  concepts: Concept[] | undefined,
+): Concept | null {
+  if (!name || !concepts) return null;
+  return concepts.find((c) => c.name === name) ?? null;
+}
+
+function resolveFrequencyByUuid(
+  uuid: string | undefined,
+  frequencies: MedicationConfig['frequencies'] | undefined,
+): MedicationConfig['frequencies'][number] | null {
```
sumaztwcode, line 152: nit: For STAT medications, the stored `boundsPeriod.start` is discarded and replaced with `new Date()`. On round-trip edit, this means the original start date is lost. Combined with the `toDateString()` comparison in `hasMedicationChanged`, this can produce false "no changes" or false "has changes" depending on timing.

Suggested fix: for edit parsing, prefer the stored `boundsPeriod.start` if present; only default to `new Date()` when no start date exists.

packages/bahmni-widgets/src/medications/components/actionHandlers.ts
```diff
@@ -12,3 +12,24 @@ export const handleAction = (
       }),
     );
 };
+
+export const handleEditAction = (
+  editMedications: MedicationRequest[],
+  encounterType: string,
+): void => {
```
sumaztwcode: nit: `editEncounterUuid` is derived from `editMedications[0]` only. For `handleEditAll`, medications are pre-filtered to the same encounter, so this is safe today. But if `encounter` is null on the FHIR resource, `editEncounterUuid` becomes `undefined`, and ConsultationPad silently falls back to `sessionEncounter`.

Consider adding a guard or documenting the assumption.

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
```diff
@@ -116,17 +116,10 @@ export async function resolveEncounterMatchDecision(
...
-      return {
-        matched: false,
-        encounter: currentPractitionerRecentEncounters[0],
-        reasons: ['MULTIPLE_ENCOUNTERS_FOUND'],
-      };
-    }
-
-    // 8. One recent encounter by this practitioner → MATCHED or LOCATION_MISMATCH
-    if (currentPractitionerRecentEncounters.length === 1) {
+    // 7–8. One or more recent encounters by this practitioner → MATCHED or LOCATION_MISMATCH
```
sumaztwcode: nit: Comment says `// 7-8.` — old step numbering from when there were two conditions. Now there's one. Renumber to `// 7.`.

sumaztwcode (changes requested):
## Review for BAH-4654

Solid feature — the FHIR medication parser is well-tested and the edit flow integrates cleanly with existing patterns. A few critical items need addressing before merge.

Found 3 blocking issues, 7 suggestions, and 4 nits.

**Acceptance Criteria**: 2/3 fully met, 1 partially met (AC1: practitioner filter not tested; behavioral change in encounter matching)

### What's Done Well
- Comprehensive FHIR parser (`medicationRequestParser.ts`) with 13 test cases
- Clean `priorPrescription` linkage for order supersession
- Idiomatic Zustand store extension with dedup guard
- Non-invasive integration with existing consultation flow

### Issues not in diff

| File:Line | Type | Issue |
|-----------|------|-------|
| `apps/clinical/public/locales/locale_es.json` | nit | Three new i18n keys (`MEDICATIONS_EDIT_ACTION`, `MEDICATIONS_EDIT_ALL`, `MEDICATIONS_EDIT_FORM_TITLE`) missing from Spanish locale |
| `MedicationsForm.tsx` | suggestion | `removeMedication` still enabled in edit mode — allows implicit Stop via the Edit flow (out-of-scope per AC). Disable close button when `isEditMode && fhirResourceId` |
| `MedicationsTable.test.tsx` | suggestion | `setPatientHeaderAttribute` helper never sets `data-active-practitioner-uuid` — every string `endsWith('')` is true, silently bypassing practitioner filter in all tests. Add practitioner UUID param and a mismatch test case |

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ROHIT-V-TW: Fixed. Removed the body-wide MutationObserver entirely. Now using `useSubscribeConsultationSaved` (the established event pattern) to re-read attributes after save. The initial `useEffect` only attaches an attribute observer if the header is already in DOM — no fallback observer needed since the `consultationSaved` event handles the post-save re-read.

apps/clinical/src/components/consultationPad/index.tsx
ROHIT-V-TW: Fixed. Added a `cancelled` flag with cleanup (`return () => { cancelled = true; }`). All state setters (`setEditEncounter`, `setEditEncounterLoading`, `addNotification`) are guarded with `if (!cancelled)` to prevent updates on unmounted components.

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
ROHIT-V-TW: Fixed. Added `console.warn` when `length > 1` to surface the anomaly. Renumbered comment to `// 7.`. The FHIR search uses `_lastUpdated` so results are sorted by recency — index `[0]` is the most recent. `MULTIPLE_ENCOUNTERS_FOUND` was removed from the type union since it's no longer produced.

apps/clinical/src/components/consultationPad/index.tsx
ROHIT-V-TW: Fixed. Replaced the inline 13-field comparison with `useMedicationStore.getState().hasEditChanges()`. The component subscribes to `selectedMedications` for reactivity, and `hasEditChanges()` is called at render time to get the current value. Single source of truth now.

apps/clinical/src/components/forms/medications/SelectedMedicationItem.tsx
ROHIT-V-TW: Fixed. Added `medicationInputEntry.fhirResourceId` to both dependency arrays and removed both `eslint-disable` comments. Since `fhirResourceId` is stable for a given entry, this won't cause extra effect runs.

apps/clinical/src/utils/fhir/medicationRequestParser.ts
ROHIT-V-TW: Fixed. Now resolves against `medicationConfig.dispensingUnits ?? medicationConfig.doseUnits` — uses the dedicated dispensing units list when configured, falls back to dose units otherwise.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ROHIT-V-TW: Acknowledged. This is a valid architectural concern — the widget layer currently computes ownership/editability internally by reading DOM attributes. Extracting a shared `useEncounterEditContext()` hook or passing editable IDs from the app layer is the right long-term approach. Documented as a follow-up refactoring item — not addressed in this PR to keep scope focused on the edit feature itself.

apps/clinical/src/stores/medicationsStore.ts
ROHIT-V-TW: Fixed. Replaced `toDateString()` with `isSameDay()` from date-fns (already a project dependency) via a new `areDatesEqual()` helper. This handles timezone and time-component differences correctly.

apps/clinical/src/utils/fhir/medicationRequestParser.ts
ROHIT-V-TW: Fixed. `parseStartDate()` now prefers the stored `boundsPeriod.start` when present (for both STAT and non-STAT). Only falls back to `new Date()` when no start date exists in the FHIR resource. This preserves the original start date on round-trip edit.

packages/bahmni-widgets/src/medications/components/actionHandlers.ts
ROHIT-V-TW: Fixed. Added a guard — if `editEncounterUuid` resolves to undefined (no encounter reference on the FHIR resource), a `console.warn` documents the fallback to session encounter. Also added a code comment explaining the assumption that all medications in a batch edit belong to the same encounter.

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
ROHIT-V-TW: Fixed. Renumbered to `// 7.`.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
```diff
@@ -232,6 +242,100 @@ const MedicationsTable: React.FC<WidgetProps> = ({
...
+
+    const observer = new MutationObserver(readAttributes);
+    observer.observe(header, {
+      attributes: true,
+      attributeFilter: OBSERVED_ATTRS,
+    });
+    return () => observer.disconnect();
+  }, []);
+
+  // Re-read attributes when consultation is saved (header may update after refetch)
```
sumaztwcode: remove the comment

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
```diff
@@ -232,6 +242,100 @@ const MedicationsTable: React.FC<WidgetProps> = ({
     return [...activeMedications, ...scheduledMedications];
   }, [allMedications]);
 
+  const [editContext, setEditContext] = useState(() => {
+    const header = document.querySelector('[data-testid="patient-header"]');
```
sumaztwcode: I am skeptical about using above construct. Also usage of new MutationObserver(readAttributes) below. check for alternative approaches.

apps/clinical/src/components/forms/medications/MedicationsForm.tsx
```diff
@@ -43,239 +45,272 @@ import styles from './styles/MedicationsForm.module.scss';
...
-  useEffect(() => {
-    if (existingMedicationsError) {
-      addNotification({
-        title: t('ERROR_DEFAULT_TITLE'),
-        message: existingMedicationsError.message,
-        type: 'error',
-      });
-    }
-  }, [existingMedicationsError, addNotification, t]);
+    const prevEditMedicationsRef = useRef(editMedications);
```
sumaztwcode: can usage of multiple useEffects be avoided, this as well as next? Check this and verify if this still satisfies the flow -                                                                                                                
                                                                                                                                       
  ENTRY: editMedications + medicationConfig both ready
     ↓
  useMemo: Parse editMedications → MedicationInputEntry[]
     ↓                                                                                                                                   
  useEffect: When parsed entries change → loadMedicationsForEdit(entries)
     ↓                                                                                                                                   
  Render with UI data                                                                                                                  
                                                                                                                                         
  Removes: Refs, double-effect, manual staleness tracking

packages/bahmni-widgets/src/medications/components/actionHandlers.ts
```diff
@@ -17,9 +17,17 @@ export const handleEditAction = (
...
 ): void => {
-  // Extract the encounter UUID from the medication being edited
+  // Extract the encounter UUID from the first medication being edited.
+  // All medications in a batch edit belong to the same encounter (pre-filtered).
   const encounterRef = editMedications[0]?.encounter?.reference;
-  const editEncounterUuid = encounterRef?.split('/').pop();
+  const editEncounterUuid = encounterRef?.split('/').pop() ?? undefined;
+
+  if (!editEncounterUuid) {
+    // eslint-disable-next-line no-console
```
sumaztwcode: remove this console.warn, I understand it is added to cater to above comment. But no need to capture this as a warning, you can respond to some of the system generated comments based on our project practices

apps/clinical/src/stores/medicationsStore.ts
```diff
@@ -35,6 +36,12 @@ export interface MedicationState {
   reset: () => void;
   getState: () => MedicationState;
 }
+function areDatesEqual(a: Date | undefined, b: Date | undefined): boolean {
```
sumaztwcode: most of the common date related functions are in common file - there was refactoring done earlier. Ensure to use available method from utils if available.

ravinderkabli (commented):
Inline review comments on specific code locations for the issues identified in the overall PR review.

apps/clinical/src/components/forms/medications/MedicationsForm.tsx
```diff
@@ -43,239 +45,272 @@ import styles from './styles/MedicationsForm.module.scss';
...
+        {medicationConfig &&
+          selectedMedications &&
+          selectedMedications.length > 0 && (
+            <BoxWHeader
+              title={t('MEDICATIONS_ADDED_MEDICATIONS')}
+              className={styles.medicationsBox}
+            >
+              {selectedMedications.map((medication) => (
+                <SelectedItem
+                  onClose={() => removeMedication(medication.id)}
```
ravinderkabli: **🟡 Issue 9: `removeMedication` enabled in edit mode has no server effect**

Removing a medication here drops the entry from the local store but does not generate a FHIR `MedicationRequest` with `status: stopped` or `status: cancelled` in the bundle. The bundle submits without that entry — the medication stays active server-side, silently inconsistent with what the user sees.

**Fix (two options):**
1. **Disable** the close/remove button in edit mode — e.g. `onClose={isEditMode ? undefined : () => removeMedication(medication.id)}` — with a tooltip explaining stop is out of scope.
2. **Handle it explicitly** — in `consultationBundleService`, for any entry present in `originalEditSnapshots` but absent from `selectedMedications` at submission time, generate a superseding `MedicationRequest` with `status: stopped`.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ravinderkabli: Reading application state off a DOM attribute via `document.querySelector` bypasses React's component model entirely.                             
  This creates invisible, untestable coupling — if `PatientHeader` ever renames its `data-testid` or attribute names,                               
  this silently breaks with no compile-time error.                                                                                                  
                                                                                                                                                    
  This entire `editContext` state block (lines 245–304) should be replaced with a subscription to `encounterSessionStore`                           
  (introduced in PR #421). `canEditOrCreate`, `activeEncounterUuid`, and `activePractitionerUuid` are already published                             
  there — one `useStore` call replaces this and the `MutationObserver` below.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
```diff
@@ -232,6 +242,100 @@ const MedicationsTable: React.FC<WidgetProps> = ({
...
+          header?.getAttribute('data-active-practitioner-uuid') ?? '',
+      });
+    };
+
+    readAttributes();
+
+    const header = document.querySelector(HEADER_SELECTOR);
+    if (!header) return;
+
+    const observer = new MutationObserver(readAttributes);
```
ravinderkabli: Three problems with this observer:                                                                                                                
                                                                                                                                                    
  1. **Fires on every DOM mutation site-wide** — watching all DOM changes means this callback runs on every React                                   
     re-render across the entire app until the patient header is found.                                                                             
  2. **Leak if header never mounts** — if `document.querySelector(HEADER_SELECTOR)` on line 279 returns null,                                       
     the observer is never created and `disconnect()` is never called, but there is no fallback cleanup.                                            
  3. **No re-observation after remount** — if the patient header unmounts and remounts (e.g. navigation),                                           
     the `attrObserver` does not re-attach.                                                                                                         
                                                                                                                                                    
  Fix: Delete this entire `useEffect` block. Subscribe to `encounterSessionStore` from `@bahmni/services` instead.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
```diff
@@ -232,6 +242,100 @@ const MedicationsTable: React.FC<WidgetProps> = ({
...
+    observer.observe(header, {
+      attributes: true,
+      attributeFilter: OBSERVED_ATTRS,
+    });
+    return () => observer.disconnect();
+  }, []);
+
+  // Re-read attributes when consultation is saved (header may update after refetch)
+  useSubscribeConsultationSaved(() => {
+    const header = document.querySelector('[data-testid="patient-header"]');
```
ravinderkabli: Same DOM-attribute-reading anti-pattern as above — reading application state from a DOM node inside an event                                      
  subscription. If the header is not in the DOM at the time a consultation is saved, all three values silently                                      
  fall back to their defaults (`false`, `''`, `''`), potentially hiding the edit button incorrectly.                                                
                                                                                                                                                    
  Fix: Remove this callback entirely once `encounterSessionStore` is used — the store will already hold the                                         
  latest values reactively.

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -108,10 +118,45 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
...
+  const editEncounterUuid = encounterSessionStartContext.editEncounterUuid as
+    | string
+    | undefined;
+  const [editEncounter, setEditEncounter] = useState<Encounter | null>(null);
+  const [editEncounterLoading, setEditEncounterLoading] = useState(false);
+  useEffect(() => {
+    if (!editEncounterUuid) return;
+    let cancelled = false;
+    setEditEncounterLoading(true);
+    get<Encounter>(`/openmrs/ws/fhir2/R4/Encounter/${editEncounterUuid}`)
```
ravinderkabli: The `let cancelled = false` flag on line 133 correctly guards state updates after unmount, but it does not cancel                                 
  the underlying HTTP request — `get<Encounter>(...)` continues running to completion even after the component                                      
  unmounts. If the user navigates away and back quickly, multiple in-flight requests can pile up and the last one                                   
  to resolve wins, regardless of order.

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
```diff
@@ -116,17 +116,15 @@ export async function resolveEncounterMatchDecision(
...
-        encounter: currentPractitionerRecentEncounters[0],
-        reasons: ['MULTIPLE_ENCOUNTERS_FOUND'],
-      };
-    }
-
-    // 8. One recent encounter by this practitioner → MATCHED or LOCATION_MISMATCH
-    if (currentPractitionerRecentEncounters.length === 1) {
+    // 7. Recent encounters by this practitioner → MATCHED or LOCATION_MISMATCH
+    //    When multiple exist, pick the first (assumed most recent by FHIR _lastUpdated sort).
+    if (currentPractitionerRecentEncounters.length >= 1) {
```
ravinderkabli, line 121: This changes the condition from `=== 1` (exactly one encounter) to `>= 1` (one or more), silently absorbing the                                   
  multiple-encounters case that the old code surfaced as an explicit `MULTIPLE_ENCOUNTERS_FOUND` error state.                                       
                                                                                                                                                    
  Two problems:                                                                                                                                     
  1. **Non-deterministic** — FHIR search sort order is not asserted anywhere in this codebase. Picking `[0]` is not                                 
     guaranteed to be the most recent encounter.                                                                                                    
  2. **Wrong scope** — this change affects all encounter session matching across the entire app, not just medication                                
     editing. Any caller that was handling `MULTIPLE_ENCOUNTERS_FOUND` explicitly is now silently bypassed.                                         
                                                                                                                                                    
  Fix: Revert this change in full. If the multiple-encounters case needs addressing, open a dedicated Jira ticket                                   
  and handle it in a separate PR with deterministic sorting and full test coverage.

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
```diff
@@ -116,17 +116,15 @@ export async function resolveEncounterMatchDecision(
...
+      if (currentPractitionerRecentEncounters.length > 1) {
+        // eslint-disable-next-line no-console
+        console.warn(
```
ravinderkabli, line 124: A `console.warn` is not a substitute for the explicit `MULTIPLE_ENCOUNTERS_FOUND` error state that was here before.                               
  The warn will be invisible in production to any user or downstream caller — there is no way for the UI to react to                                
  it or surface a meaningful message to the clinician. This is a silent data integrity risk.

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -226,22 +271,34 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
...
-      hasError ||
-      !isEncounterDetailsFormReady ||
-      isSubmitting ||
-      !hasConsultationData,
-    [hasError, isEncounterDetailsFormReady, isSubmitting, hasConsultationData],
-  );
+  const isEditMode = !!editOnlyKey;
+  // Subscribe to selectedMedications to trigger re-renders on field changes;
+  // hasEditChanges() reads from the store via get() at call time.
+  useMedicationStore((state) => state.selectedMedications);
```
ravinderkabli: Calling `useMedicationStore.getState()` directly inside the render function is a React anti-pattern. `.getState()`                                
  is a snapshot read that bypasses React's scheduler — in concurrent mode it can return a stale value that is                                       
  inconsistent with what React is currently rendering.                                                                                              
                                                                                                                                                    
  The subscription on line 277 (`useMedicationStore((state) => state.selectedMedications)`) is added as a workaround                                
  to force re-renders, but this is fragile — it only re-evaluates when `selectedMedications` changes, not when                                      
  `originalEditSnapshots` changes (e.g. on `loadMedicationsForEdit`).

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
```diff
@@ -232,6 +242,100 @@ const MedicationsTable: React.FC<WidgetProps> = ({
...
+        header?.getAttribute('data-active-encounter-uuid') ?? '',
+      activePractitionerUuid:
+        header?.getAttribute('data-active-practitioner-uuid') ?? '',
+    });
+  }, []);
+
+  const { canEditEncounter, activeEncounterUuid, activePractitionerUuid } =
+    editContext;
+
+  const editableMedications = useMemo(() => {
```
ravinderkabli, line 245: `editableMedications` filters by active encounter UUID and practitioner UUID to decide which rows show an edit                                    
  button. Encounter ownership and practitioner authorization are app-layer concerns — a display widget should only                                  
  render what it is told to render, not make authorization decisions.                                                                               
                                                                                                                                                    
  This also makes the widget non-reusable in any context where the ownership model differs (e.g. a supervisor                                       
  editing another provider's medications).                                                                                                          

You can either Move this `useMemo` derivation into `MedicationsForm.tsx` or a dedicated selector in the app layer.                                          
  Pass the result to the widget as a prop — e.g. `editableMedicationIds: Set<string>` — and let the widget                                          
  use it purely for conditional rendering.

apps/clinical/src/stores/medicationsStore.ts
```diff
@@ -32,8 +36,46 @@ export interface MedicationState {
...
+  originalEditSnapshots: new Map(),
+
+  hasEditChanges: () => {
+    const { selectedMedications, originalEditSnapshots } = get();
+    if (selectedMedications.some((m) => !m.fhirResourceId)) return true;
+    return selectedMedications.some((m) => {
+      const original = originalEditSnapshots.get(m.id);
+      if (!original) return true;
+      return hasMedicationChanged(m, original);
+    });
```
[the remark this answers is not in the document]
ROHIT-V-TW: Fixed. `hasEditChanges()` now tracks deletions via `originalEditIds`. On load, we capture all edit entry IDs. When checking for changes, we verify if any original ID is missing from the current selection:

```ts
const currentIds = new Set(selectedMedicationRequests.map((m) => m.id));
if (originalEditIds.some((id) => !currentIds.has(id))) return true;
```

Additionally, the approach was refactored per review feedback — instead of snapshot comparison across 13 fields, we now use an `isModified` flag per entry that gets set by `withEditFlag()` whenever any update method is called on an edited item.

packages/bahmni-widgets/src/medications/__tests__/Actions.test.tsx
```diff
@@ -46,87 +46,55 @@ describe('Actions', () => {
...
         medication={fhirMedicationRequestMock}
+        hiddenActionTypes={['administer']}
       />,
     );
 
-    const button = screen.getByTestId('medication-action-administer-button');
-    expect(button).toHaveProperty('disabled', expectDisabled);
+    const menu = screen.getByTestId('medication-actions-menu-test-med-id');
+    expect(menu).toBeInTheDocument();
   });
```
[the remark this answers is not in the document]
ROHIT-V-TW: Acknowledged. The test now correctly asserts the menu trigger exists. The deeper assertion (clicking menu and verifying hidden actions don't appear) is a valid improvement — will address as a follow-up test enhancement.

packages/bahmni-widgets/src/medications/__tests__/MedicationsTable.test.tsx
```diff
@@ -770,4 +770,135 @@ describe('MedicationsTable', () => {
...
+    it('shows actions overflow menu when location mismatch but can-edit-encounter is true', () => {
+      setupWithActiveMeds();
+      setPatientHeaderAttribute('LOCATION_MISMATCH', 'true');
+
+      render(<MedicationsTable config={editConfig} />);
+
+      expect(
+        screen.getByTestId('medication-actions-menu-1'),
+      ).toBeInTheDocument();
+    });
```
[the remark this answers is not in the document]
ROHIT-V-TW: Acknowledged. The gating logic has been significantly refactored — MutationObserver/DOM queries replaced with `useEncounterSessionStore()`. The test assertions should be updated to verify the new store-based gating. Will address as a follow-up.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
```diff
@@ -318,7 +428,32 @@ const MedicationsTable: React.FC<WidgetProps> = ({
...
+            testId="medications-edit-all-button"
+          >
+            <Icon
+              id="edit-all-medication-icon"
+              name="fa-pen-to-square"
+              size={ICON_SIZE.SM}
+            />
+          </IconButton>
+        </div>
+      )}
```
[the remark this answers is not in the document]
ROHIT-V-TW: Fixed. The "Edit All" button is now gated with `canEdit`:

```tsx
{editAction && canEdit && (
  <div className={styles.widgetEditActions}>
    ...
  </div>
)}
```

Users without the required privilege will not see the edit entry point at all.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ROHIT-V-TW: Fixed. Comment removed.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ROHIT-V-TW: Fixed. Completely removed the MutationObserver and all `document.querySelector` usage. Replaced with a `encounterSessionStore` — a lightweight pub/sub store on `globalThis` that PatientHeader writes to and MedicationsTable reads via `useEncounterSessionStore()` hook. This follows the same reactive pattern as the Edit Allergies/Conditions PR.

apps/clinical/src/components/forms/medications/MedicationsForm.tsx
ROHIT-V-TW: Refactored. The old `MedicationsForm.tsx` has been replaced by the new `MedicationRequestForm.tsx` (from BAH-4721 merge). The edit flow now uses two focused useEffects:

1. `setAttributes` on mount (existing, from main)
2. Parse pending FHIR edits when `medicationConfig` loads (new, runs once via `editLoadedRef` guard)

These serve distinct lifecycle purposes and can't be meaningfully combined — `setAttributes` runs immediately on mount while the FHIR parsing depends on async config loading. The ref guard ensures it only runs once.

packages/bahmni-widgets/src/medications/components/actionHandlers.ts
ROHIT-V-TW: Fixed. Removed the `console.warn`. The fallback to session encounter when `editEncounterUuid` is undefined is now handled silently — ConsultationPad already falls back gracefully via the `editEncounterUuid ? editEncounter : sessionEncounter` logic.

apps/clinical/src/stores/medicationsStore.ts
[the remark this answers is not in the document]
ROHIT-V-TW: Fixed. hasEditChanges() now tracks deletions via originalEditIds. On edit load, we capture all entry IDs. When checking for changes, we verify if any original ID is missing from current selection. Also refactored to use isModified flag per entry instead of 13-field snapshot comparison.

packages/bahmni-widgets/src/medications/__tests__/Actions.test.tsx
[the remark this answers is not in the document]
ROHIT-V-TW: Acknowledged. Will improve test assertions to click the menu and verify hidden actions do not appear — as a follow-up test enhancement.

packages/bahmni-widgets/src/medications/__tests__/MedicationsTable.test.tsx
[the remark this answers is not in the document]
ROHIT-V-TW: Acknowledged. The gating logic has been significantly refactored — MutationObserver/DOM queries replaced with useEncounterSessionStore(). The test assertions should be updated to verify the new store-based gating. Will address as a follow-up.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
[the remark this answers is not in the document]
ROHIT-V-TW: Fixed. The "Edit All" button is now gated with canEdit: {editAction && canEdit && (...)}. Users without the required privilege will not see the edit entry point.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ROHIT-V-TW: Fixed. The entire MutationObserver/DOM query block (including the comment) has been removed and replaced with useEncounterSessionStore() — a Zustand-like pub/sub store on globalThis.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ROHIT-V-TW: Fixed. Completely removed the MutationObserver and all document.querySelector usage. Replaced with encounterSessionStore — a lightweight pub/sub store on globalThis that PatientHeader writes to and MedicationsTable reads via useEncounterSessionStore() hook. This follows the same reactive pattern as the Edit Allergies/Conditions PR (#421).

apps/clinical/src/components/forms/medications/MedicationsForm.tsx
ROHIT-V-TW: Refactored. The old MedicationsForm.tsx has been replaced by MedicationRequestForm.tsx (from BAH-4721 merge). The edit flow now uses two focused useEffects: (1) setAttributes on mount (existing, from main) and (2) parse pending FHIR edits when medicationConfig loads (new, runs once via editLoadedRef guard). These serve distinct lifecycle purposes — setAttributes runs immediately on mount while FHIR parsing depends on async config loading.

packages/bahmni-widgets/src/medications/components/actionHandlers.ts
ROHIT-V-TW: Fixed. Removed the console.warn. The fallback to session encounter when editEncounterUuid is undefined is handled silently — ConsultationPad already falls back gracefully via the editEncounterUuid ? editEncounter : sessionEncounter logic.

ROHIT-V-TW:
### Replies to outstanding review comments

**@coderabbitai — Track deletions in hasEditChanges():**
Fixed. hasEditChanges() now tracks deletions via originalEditIds. On edit load, we capture all entry IDs. When checking for changes, we verify if any original ID is missing from current selection. Also refactored to use isModified flag per entry instead of 13-field snapshot comparison.

**@coderabbitai — Hidden-action test assertion:**
Acknowledged. Will improve test assertions to click the menu and verify hidden actions do not appear — as a follow-up test enhancement.

**@coderabbitai — Encounter-gating tests false positives:**
Acknowledged. The gating logic has been significantly refactored — MutationObserver/DOM queries replaced with useEncounterSessionStore(). Test assertions will be updated as a follow-up.

**@coderabbitai — Hide Edit All without privilege:**
Fixed. The Edit All button is now gated with canEdit: editAction && canEdit && (...). Users without the required privilege will not see the edit entry point.

**@sumaztwcode — remove the comment:**
Fixed. The entire MutationObserver/DOM query block has been removed and replaced with useEncounterSessionStore().

**@sumaztwcode — skeptical about MutationObserver:**
Fixed. Completely removed the MutationObserver and all document.querySelector usage. Replaced with encounterSessionStore — a lightweight pub/sub store on globalThis that PatientHeader writes to and MedicationsTable reads via useEncounterSessionStore() hook. This follows the same reactive pattern as the Edit Allergies/Conditions PR (#421).

**@sumaztwcode — can usage of multiple useEffects be avoided:**
Refactored. The old MedicationsForm.tsx has been replaced by MedicationRequestForm.tsx (from BAH-4721 merge). The edit flow now uses two focused useEffects: (1) setAttributes on mount and (2) parse pending FHIR edits when medicationConfig loads (runs once via editLoadedRef guard). These serve distinct lifecycle purposes and cannot be combined.

**@sumaztwcode — remove console.warn:**
Fixed. Removed the console.warn from actionHandlers. The fallback to session encounter is handled silently by ConsultationPad.

apps/clinical/src/stores/medicationsStore.ts
ROHIT-V-TW: Acknowledged. The old medicationsStore.ts with custom date comparison has been removed. The new architecture uses isModified flag per entry instead of date-based snapshot comparison, so the date utility concern no longer applies.

apps/clinical/src/components/forms/medications/MedicationsForm.tsx
ROHIT-V-TW: Valid concern. For this PR, remove is disabled in edit mode. The $stop operation for removed medications will be handled as a follow-up once the backend FHIR2 module fully supports the priorPrescription → stop flow.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ROHIT-V-TW: Fixed. Completely removed document.querySelector and all DOM attribute reading. Replaced with encounterSessionStore — a lightweight pub/sub store on globalThis that PatientHeader writes to and MedicationsTable reads via useEncounterSessionStore() hook.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ROHIT-V-TW: Fixed. Deleted the entire MutationObserver useEffect block. Now subscribing to encounterSessionStore from @bahmni/services instead — no DOM observation, no leak, no re-attach needed.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ROHIT-V-TW: Fixed. Removed this callback entirely. The encounterSessionStore holds the latest values reactively — no need to re-read DOM attributes on consultation save.

apps/clinical/src/components/consultationPad/index.tsx
ROHIT-V-TW: Acknowledged. The cancelled flag prevents state updates on unmounted components. Adding AbortController with signal to the get<Encounter>() call is a valid improvement — will address as a follow-up since the current get() utility does not accept a signal parameter.

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
ROHIT-V-TW: The change from === 1 to >= 1 was done because FHIR search uses _lastUpdated sort, so [0] is the most recent. MULTIPLE_ENCOUNTERS_FOUND was never in the MatchReasonCode type union — it was only a comment. The console.warn surfaces the anomaly for debugging. This change predates this PR and is already on main.

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
ROHIT-V-TW: Acknowledged. The warn is a diagnostic aid, not a UI signal. As noted above, MULTIPLE_ENCOUNTERS_FOUND was never an actual enum value in MatchReasonCode — the old code returned matched: false with no specific reason code for this case. The current approach (pick first + warn) is a pragmatic improvement. If a dedicated error state is needed, it should be a separate ticket with full test coverage.

apps/clinical/src/components/consultationPad/index.tsx
ROHIT-V-TW: Fixed. Replaced with useSyncExternalStore that directly subscribes to the medication store and returns hasEditChanges() result reactively. No more .getState() call in render path.

packages/bahmni-widgets/src/medications/MedicationsTable.tsx
ROHIT-V-TW: Acknowledged. This is a valid architectural concern — the widget computes ownership internally. Extracting a shared hook or passing editable medication IDs from the app layer is the right long-term approach. Documented as a follow-up refactoring item to keep this PR focused on the edit feature.

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -126,6 +172,18 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
...
+  useEffect(() => {
+    const editMedications = encounterSessionStartContext.editMedications as
+      | MedicationRequest[]
+      | undefined;
+    if (editMedications?.length) {
+      getMedicationRequestStore(MEDICATIONS_INPUT_CONTROL_KEY)
+        .getState()
+        .setPendingFhirEdits(editMedications);
+    }
+  }, [encounterSessionStartContext.editMedications]);
```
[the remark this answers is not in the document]
ROHIT-V-TW: Fixed. The seeding useEffect now always calls setPendingFhirEdits, passing [] when editMedications is empty/undefined to prevent stale state reuse.

apps/clinical/src/components/forms/medicationRequest/components/SelectedMedicationRequestItem.tsx
```diff
@@ -127,6 +127,9 @@ const SelectedMedicationRequestItem: React.FC<SelectedMedicationRequestItemProps
     ]);
 
     useEffect(() => {
+      // Skip applying defaults for items loaded for edit — they already have
+      // their values populated from the FHIR resource.
+      if (entry.fhirResourceId) return;
       applyMountDefaults({
```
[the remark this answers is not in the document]
ROHIT-V-TW: Fixed. Added if (entry.fhirResourceId) return; guard to the STAT/frequency normalization useEffect as well, preventing it from nulling parsed FHIR frequency values on edit rows.

apps/clinical/src/components/forms/medicationRequest/MedicationRequestForm.tsx
```diff
@@ -91,13 +92,38 @@ const MedicationRequestForm: React.FC<{
...
+    ) {
+      editLoadedRef.current = true;
+      const storeKey = inputControlType as MedicationRequestStoreKey;
+      const store = getMedicationRequestStore(storeKey);
+      const entries = pendingFhirEdits.map((fhir) =>
+        parseFhirToMedicationInputEntry(fhir, medicationConfig),
+      );
+      store.getState().loadMedicationsForEdit(entries);
+    }
+  }, [pendingFhirEdits, medicationConfig, inputControlType]);
```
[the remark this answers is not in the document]
ROHIT-V-TW: Fixed. Removed editLoadedRef entirely. Since loadMedicationsForEdit() clears pendingFhirEdits to [], the pendingFhirEdits.length > 0 guard naturally prevents re-execution without blocking later sessions.

apps/clinical/src/components/forms/medicationRequest/store.ts
```diff
@@ -207,6 +215,10 @@ function applyDispenseUnitUpdate(
   return updated;
 }
 
+function withEditFlag(updated: MedicationInputEntry): MedicationInputEntry {
+  return updated.fhirResourceId ? { ...updated, isModified: true } : updated;
+}
```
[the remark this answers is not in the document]
ROHIT-V-TW: Acknowledged. The current withEditFlag marks any update call as modified. In practice, no-op updates (setting the same value) are rare since they require explicit user interaction with form controls. Will address as a follow-up optimization if it causes false-positive change detection in real usage.

apps/clinical/src/components/patientHeader/PatientHeader.tsx
```diff
@@ -49,6 +55,17 @@ const PatientHeader: React.FC<PatientHeaderProps> = ({
...
+  // Publish encounter session state so widgets can read it without DOM queries.
+  useEffect(() => {
+    setEncounterSessionState({
+      canEditOrCreate: editActiveEncounter,
+      activeEncounterUuid:
+        editActiveEncounter && activeEncounter?.id ? activeEncounter.id : null,
+      activePractitionerUuid: practitioner?.uuid ?? null,
+      isLoading,
+    });
+  }, [editActiveEncounter, activeEncounter, practitioner, isLoading]);
```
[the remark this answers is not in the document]
ROHIT-V-TW: Fixed. Added a cleanup useEffect that calls resetEncounterSession() on PatientHeader unmount to clear stale cross-bundle state.

packages/bahmni-services/src/encounterSessionService/encounterMatchDecisionMapper.ts
ravinderkabli: ok

apps/clinical/src/components/forms/medicationRequest/components/SelectedMedicationRequestItem.tsx
```diff
@@ -90,6 +90,9 @@ const SelectedMedicationRequestItem: React.FC<SelectedMedicationRequestItemProps
     }, [dosage, frequency, duration, durationUnit, id, updateDispenseQuantity]);
```
ravinderkabli: blocking: This effect runs on mount for every edit-mode entry loaded via `loadMedicationsForEdit`, calls `updateDispenseQuantity`, and since `updateDispenseQuantity` now wraps with `withEditFlag`, every entry immediately gets `isModified: true` before any user interaction.

This defeats the `hasEditChanges()` save-button guard: the Save button will always be enabled as soon as edit items load, even if nothing has changed.

The other two effects below this one (STAT normalization at line 95, mount defaults at line 135) both have `if (entry.fhirResourceId) return;` guards. This one is missing that guard.

Fix:
```typescript
useEffect(() => {
  if (entry.fhirResourceId) return; // skip auto-calc for FHIR-loaded edit entries
  const totalQuantity = calculateTotalQuantity(dosage, frequency, duration, durationUnit);
  updateDispenseQuantity(id, totalQuantity);
}, [dosage, frequency, duration, durationUnit, id, updateDispenseQuantity]);
```

apps/clinical/src/utils/fhir/medicationRequestParser.ts
```diff
@@ -0,0 +1,204 @@
+import {
+  MedicationRequest as FhirMedicationRequest,
+  Medication,
+} from 'fhir/r4';
+import { DURATION_UNIT_OPTIONS } from '../../components/forms/medicationRequest/constants';
```
ravinderkabli: suggestion: This import creates a layer inversion. `utils/fhir/` is a shared utility layer but it imports from `components/forms/medicationRequest/`. All other files in `utils/fhir/` (`medicationRequestResourceCreator.ts`, `medicationUtilities.ts`) only import from `models/` or top-level `constants/`, never from `components/`.

Consider moving `DURATION_UNIT_OPTIONS` to `apps/clinical/src/constants/medications.ts` (where `STAT_ORDER_VALIDITY_MS` already lives) and importing from there in both this file and the components constants file.

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -120,10 +124,45 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
...
+    });
+
+  const editEncounterUuid = encounterSessionStartContext.editEncounterUuid as
+    | string
+    | undefined;
+  const [editEncounter, setEditEncounter] = useState<Encounter | null>(null);
+  const [editEncounterLoading, setEditEncounterLoading] = useState(false);
+  useEffect(() => {
+    if (!editEncounterUuid) return;
+    let cancelled = false;
```
ravinderkabli: nit: The `cancelled` flag prevents state updates after unmount, but the in-flight HTTP request is not actually cancelled and continues running in the background. If the `get()` utility supports an `AbortController` signal, prefer that for true request cancellation.

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -120,10 +124,45 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
...
+    setEditEncounterLoading(true);
+    get<Encounter>(`/openmrs/ws/fhir2/R4/Encounter/${editEncounterUuid}`)
```
ravinderkabli: nit: This is the only place in `ConsultationPad` that calls `get<>()` directly with a raw FHIR URL instead of going through a named service function. Other FHIR calls use `getAllergies()`, `getMedicationsBundle()`, etc. from `@bahmni/services`.

Consider extracting `getEncounter(uuid)` to the encounter service for consistency and testability.

packages/bahmni-widgets/src/medications/components/actionHandlers.ts
```diff
@@ -5,10 +5,37 @@ export const handleAction = (
...
+  // All medications in a batch edit belong to the same encounter (pre-filtered).
+  const encounterRef = editMedications[0]?.encounter?.reference;
+  const editEncounterUuid = encounterRef?.split('/').pop() ?? undefined;
+
+  globalThis.dispatchEvent(
+    new CustomEvent('startConsultation', {
+      detail: {
+        encounterType,
+        editMedications,
+        editOnly: 'medication',
```
ravinderkabli: question: This string `'medication'` must match `MEDICATIONS_INPUT_CONTROL_KEY` defined in `apps/clinical/src/components/forms/medicationRequest/constants.ts`. Since `bahmni-widgets` cannot import from `apps/clinical`, the coupling is implicit. If the key is renamed, this breaks silently.

Would it make sense to move `MEDICATIONS_INPUT_CONTROL_KEY` to `@bahmni/services` so both sides can share the constant?

ravinderkabli (changes requested):
## Review for BAH-4654

Good progress — the props pattern fix (replacing `useEncounterSessionStore()` in `MedicationsTable` with props from `DashboardSection`) correctly addresses the prior review feedback.

apps/clinical/src/components/forms/medicationRequest/components/SelectedMedicationRequestItem.tsx
ROHIT-V-TW: Fixed. Added `if (entry.fhirResourceId) return;` guard to skip auto-calculation for FHIR-loaded edit entries. Additionally, replaced the `isModified` one-way flag with snapshot-based comparison in `hasEditChanges()` — the "Done" button now correctly re-disables when the user reverts values to their originals. `withEditFlag` has been removed entirely.

apps/clinical/src/utils/fhir/medicationRequestParser.ts
ROHIT-V-TW: Fixed. Moved `DURATION_UNIT_OPTIONS` to `apps/clinical/src/constants/medications.ts` (alongside `STAT_ORDER_VALIDITY_MS`). The parser now imports from the constants layer instead of crossing into `components/forms/`. The old `medicationRequest/constants.ts` re-exports from there for backward compatibility.

apps/clinical/src/components/consultationPad/index.tsx
ROHIT-V-TW: Fixed. Replaced the `cancelled` flag with `AbortController` — the in-flight HTTP request is now actually cancelled on unmount via `abortController.abort()`.

apps/clinical/src/components/consultationPad/index.tsx
ROHIT-V-TW: Fixed. Extracted `getEncounterByUuid(uuid, options?)` to `@bahmni/services/encounterService` and exported from the package. ConsultationPad now calls the named service function instead of raw `get<>()`.

packages/bahmni-widgets/src/medications/components/actionHandlers.ts
ROHIT-V-TW: Fixed. Moved `MEDICATIONS_INPUT_CONTROL_KEY` to `@bahmni/services` (in `medicationRequestService/constants.ts`). Both `actionHandlers.ts` in `@bahmni/widgets` and `apps/clinical` now import from the shared package — no more implicit string coupling.

ravinderkabli (approved):
Looks fine

sumaztwcode (approved):

PR #425, opened 2026-05-22
BAH-4711 | Add. Support To Establish Non Working Days

skjameela276:
**JIRA** → [BAH-4711](https://bahmni.atlassian.net/browse/BAH-4711)

Introduces the Appointment Unavailability management feature in the appointments app, allowing admins to define unavailability windows for providers and services with configurable date/time ranges and recurrence options.

  ---

  ### **Key Features**

  - New admin page for creating and managing appointment unavailability slots
  - Filterable multi-select dropdowns for providers and appointment services
  - Support for specifying date ranges, time ranges, and recurrence days
  - Privilege check to restrict access to the unavailability page
  - Full i18n support for all labels and messages

[BAH-4711]: https://bahmni.atlassian.net/browse/BAH-4711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/bahmni-services/src/appointmentService/models.ts
```diff
@@ -36,3 +36,30 @@ export interface AppointmentService {
...
+  voided: boolean;
+  dateCreated: string;
+  creatorName: string;
+}
+
+export interface CreateUnavailabilityRequest {
+  locationUuid: string;
+  appointmentServiceUuid?: string;
+  providerUuid?: string;
+  startDate: string;
```
rahu1ramesh, line 67: Won't it be better to have kept this as `startDateTime` and `endDateTime`?

packages/bahmni-services/src/appointmentService/models.ts
```diff
@@ -36,3 +36,30 @@ export interface AppointmentService {
   color: string;
   initialAppointmentStatus: string | null;
 }
+
+export interface AppointmentUnavailability {
+  uuid: string;
+  locationUuid: string;
```
rahu1ramesh: Instead can we send an OpenMRS Location, Appointment, Provider object ?

packages/bahmni-services/src/appointmentService/models.ts
```diff
@@ -36,3 +36,30 @@ export interface AppointmentService {
...
+  locationName: string;
+  appointmentServiceUuid: string;
+  appointmentServiceName: string;
+  providerUuid: string | null;
+  providerName: string | null;
+  startDate: string;
```
rahu1ramesh, line 54: Won't it be better to have kept this as startDateTime and endDateTime?

packages/bahmni-services/src/appointmentService/appointmentService.ts
```diff
@@ -144,3 +150,24 @@ export async function getPastAppointmentsPage(
...
+  AppointmentUnavailability[]
+> => {
+  return await get<AppointmentUnavailability[]>(APPOINTMENT_UNAVAILABILITY_URL);
+};
+
+/**
+ * Creates a new appointment unavailability
+ * @param data - The unavailability data to create
+ * @returns The created unavailability
+ */
```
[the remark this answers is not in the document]
rahu1ramesh: Please fix this.

packages/bahmni-services/src/appointmentService/__tests__/appointmentService.test.ts
```diff
@@ -327,4 +334,156 @@ describe('Appointment Service', () => {
       );
     });
   });
+
+  describe('getAppointmentUnavailabilities', () => {
+    const mockUnavailabilities: AppointmentUnavailability[] = [
```
rahu1ramesh: Move the mocks into the packages/bahmni-services/src/appointmentService/__mocks__/mocks.ts file.

packages/bahmni-services/src/appointmentService/__tests__/appointmentService.test.ts
```diff
@@ -327,4 +334,156 @@ describe('Appointment Service', () => {
...
+      mockedGet.mockRejectedValue(new Error('Unavailabilities API Error'));
+
+      await expect(getAppointmentUnavailabilities()).rejects.toThrow(
+        'Unavailabilities API Error',
+      );
+    });
+  });
+
+  describe('createAppointmentUnavailability', () => {
+    const mockCreateRequest: CreateUnavailabilityRequest[] = [
```
rahu1ramesh: Move mock to packages/bahmni-services/src/appointmentService/__mocks__/mocks.ts

packages/bahmni-services/src/providerService/constants.ts
```diff
@@ -1,4 +1,8 @@
 import { OPENMRS_REST_V1 } from '../constants/app';
 
 export const PROVIDER_RESOURCE_URL = (userUUID: string) =>
-  OPENMRS_REST_V1 + `/provider?user=${userUUID}&v=custom:(uuid,display,person)`;
+  OPENMRS_REST_V1 +
+  `/provider?user=${userUUID}&v=custom:(uuid,display,person,attributes)`;
+
+export const ALL_PROVIDERS_URL =
+  OPENMRS_REST_V1 + '/provider?v=custom:(uuid,display,person,attributes)';
```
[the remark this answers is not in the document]
rahu1ramesh: Good to confirm with @mohan-13 @angshu whether this endpoint is paginated, if so we would need to do more work here.

packages/bahmni-services/src/providerService/models.ts
```diff
@@ -37,13 +37,38 @@ export interface Person {
...
+  uuid: string;
+  display: string;
+}
+
+export interface LocationTag {
+  uuid?: string;
+  display: string;
+}
+
+export interface LocationAttributeValue {
```
rahu1ramesh: Is this not already in the LocationService?

packages/bahmni-services/src/providerService/providerService.ts
```diff
@@ -19,3 +20,50 @@ export async function getCurrentProvider(
...
   return providerResponse.results[0];
 }
+
+/**
+ * Fetches all providers from OpenMRS
+ * @returns Promise resolving to an array of Provider objects
+ */
+export async function getAllProviders(): Promise<Provider[]> {
+  const response = await get<ProviderResponse>(ALL_PROVIDERS_URL);
+  return response.results ?? [];
```
rahu1ramesh: Do we need this fallback?

packages/bahmni-services/src/providerService/providerService.ts
```diff
@@ -19,3 +20,50 @@ export async function getCurrentProvider(
...
+}
+
+/**
+ * Fetches login locations from user's provider attributes
+ * @param userUuid - The UUID of the user whose provider login locations to fetch
+ * @returns Promise resolving to an array of Location objects from provider's Login Locations attributes
+ */
+export async function getProviderLoginLocations(
```
rahu1ramesh, line 48: I would suggest we split this into 2 parts, use `getCurrentProvider` to get the current provider details and then on the appointments app, filter out the locations with tag "Appointment Location".

packages/bahmni-services/src/utils/utils.ts
```diff
@@ -401,3 +401,27 @@ export function convertToSentenceCase(str: string): string {
     .toLowerCase()
     .replace(/^./, (char) => char.toUpperCase());
 }
+
+/**
+ * Converts 12-hour time format (with AM/PM) to 24-hour format
+ * @param timeStr - Time string in 12-hour format (e.g., "02:30 PM", "11:45 AM")
+ * @returns Time string in 24-hour format (e.g., "14:30", "11:45"), or empty string if invalid
+ */
```
rahu1ramesh: Instead of the new fn can we use the date-fns function. date-fns doesn't have a direct "12h string → 24h string" converter, but date-fns/parse + date-fns/format covers it:

```
  import { parse, format } from 'date-fns';

  const parsed = parse('02:30 PM', 'hh:mm aa', new Date());
  format(parsed, 'HH:mm');
```

rahu1ramesh (changes requested):

packages/bahmni-services/src/providerService/providerService.ts
skjameela276: If the API has no providers, `response.results` itself will be `[]` but If `results` field is missing entirely from the response due to - Malformed API responses ,Network issues returning unexpected data then this fallback good to have it.

packages/bahmni-services/src/appointmentService/models.ts
skjameela276: Due to limitation of weekly availability model we are having date and time separately

packages/bahmni-services/src/appointmentService/models.ts
skjameela276: Due to limitation of weekly availability model we are having date and time separately

apps/appointments/src/pages/admin/appointmentUnavailability/constants.ts
```diff
@@ -0,0 +1,7 @@
+export const PROVIDER_ATTRIBUTE_AVAILABLE = 'Available for appointments';
+export const GET_APPOINTMENT_UNAVAILABILITY_PRIVILEGE =
```
rahu1ramesh, line 3: @skjameela276 Can we get these privilege constants from the config?

apps/appointments/src/pages/admin/appointmentUnavailability/constants.ts
```diff
@@ -0,0 +1,7 @@
...
+  'Get Appointment Unavailability';
+export const ADD_APPOINTMENT_UNAVAILABILITY_PRIVILEGE =
```
rahu1ramesh, line 5: @skjameela276 Can we get these privilege constants from the config?

apps/appointments/src/pages/admin/appointmentUnavailability/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+      headerWSideNav={<Header breadcrumbItems={breadcrumbs} />}
+      hasSideNav={false}
+      mainDisplay={
+        <div
+          id="appointment-unavailability-page"
+          data-testid="appointment-unavailability-page-test-id"
+          aria-label="appointment-unavailability-page-aria-label"
+          className={styles.page}
+        >
+          <div className={styles.header}>
```
rahu1ramesh: The ActionDataTable has capability to add the title and button, better to reuse that.

1. Remove the div.header block from JSX
2. Set title={t('ADMIN_UNAVAILABILITY_TITLE')} on ActionDataTable
3. Pass actionButton={!isFormVisible && canAddUnavailability ? { label: ..., onClick: ... } : undefined} — toolbar won't render when undefined, which matches the current hide-not-disable behavior
4. Remove Button import (no longer needed)
5. Delete .header, .title, .addBtn from the SCSS file

apps/appointments/src/pages/admin/appointmentUnavailability/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+      },
+    ];
+  }, [t]);
+
+  const rows = useMemo(() => {
+    if (!data) return [];
+
+    return data.map((item: AppointmentUnavailability) => ({
+      id: item.uuid,
+      startDateTime: `${formatDateTime(item.startDate).formattedResult}, ${formatTime(item.startTime)}`,
```
rahu1ramesh: Instead of the newly added formatTime, we can use formatDateTime itself here, which is available from bahmni/services, please do remove the newly added fn as well.

apps/appointments/src/pages/admin/appointmentUnavailability/utils.ts
```diff
@@ -0,0 +1,20 @@
+import { convertTo24HourFormat } from '@bahmni/services';
+
+export const formatTime = (time: string): string => {
```
rahu1ramesh: Instead of the newly added formatTime, we can use formatDateTime itself here, which is available from bahmni/services, please do remove the newly added fn as well.

apps/appointments/src/pages/admin/appointmentUnavailability/utils.ts
```diff
@@ -0,0 +1,20 @@
...
+  dateObj.setHours(parseInt(hours), parseInt(minutes));
+
+  return dateObj.toLocaleTimeString('en-US', {
+    hour: '2-digit',
+    minute: '2-digit',
+    hour12: true,
+  });
+};
+
+export const getTimeInMinutes = (timeStr: string): number | null => {
```
rahu1ramesh: Feels like a global concern, can we move this to bahmni/services

apps/appointments/src/pages/admin/appointmentUnavailability/constants.ts
```diff
@@ -0,0 +1,7 @@
...
+  'Add Appointment Unavailability';
+export const DATE_FORMAT = 'yyyy-MM-dd';
```
rahu1ramesh: @kousalyapt-tw do we need this date format hard coded, can we rely on the config

cc: @PratitiP @skjameela276

apps/appointments/src/pages/admin/appointmentUnavailability/styles/index.module.scss
```diff
@@ -0,0 +1,34 @@
...
+  z-index: 100; 
+}
+
+.title {
+  @include type-style("heading-compact-02");
+  color: $text-primary;
+  margin-left: 1rem;
+}
+
+.addBtn {
```
rahu1ramesh: Can be removed.

apps/appointments/src/pages/admin/appointmentUnavailability/styles/index.module.scss
```diff
@@ -0,0 +1,34 @@
...
+  justify-content: space-between;
+  align-items: center;
+  background-color: $layer-01;
+  height: 54px;
+   position: sticky;
+  top: 0; /* Sticks to the very top */
+  z-index: 100; 
+}
+
+.title {
```
rahu1ramesh: Can be removed.

apps/appointments/src/pages/admin/appointmentUnavailability/styles/index.module.scss
```diff
@@ -0,0 +1,34 @@
...
+
+.page {
+  overflow-y: auto;
+  margin: 0 auto;
+  margin-block: $spacing-10;
+  margin-inline: calc($spacing-12 + $spacing-05);
+  box-shadow: 0px 2px 6px 0px $shadow;
+}
+
+.header {
```
rahu1ramesh: Can be removed.

apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -0,0 +1,511 @@
...
+    endTime: '',
+    endTimePeriod: 'AM',
+  });
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const [showValidation, setShowValidation] = useState(false);
+  const [dateTimeError, setDateTimeError] = useState('');
+  const [servicesKey, setServicesKey] = useState(0);
+  const [providersKey, setProvidersKey] = useState(0);
+
+  const todayAtMidnight = useMemo(() => {
```
rahu1ramesh: Use getTodayDate from bahmni/services instead.

apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -0,0 +1,511 @@
...
+        message: t('ADMIN_UNAVAILABILITY_FORM_ERROR_MESSAGE'),
+        type: 'error',
+      });
+    } finally {
+      setIsSubmitting(false);
+    }
+  };
+
+  const formContent = (
+    <div className={styles.formContent}>
```
rahu1ramesh: Please add relevant id and data-testid to all major elements

apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -0,0 +1,511 @@
...
+  const [servicesKey, setServicesKey] = useState(0);
+  const [providersKey, setProvidersKey] = useState(0);
+
+  const todayAtMidnight = useMemo(() => {
+    const today = new Date();
+    today.setHours(0, 0, 0, 0);
+    return today;
+  }, []);
+
+  const findItemByUuid = <T extends { uuid: string }>(
```
rahu1ramesh: This is a oneliner that's used exactly once, do we need a seperate fn for this

apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -0,0 +1,511 @@
...
+  ): T | null => {
+    if (!uuid) return null;
+    return items.find((item) => item.uuid === uuid) ?? null;
+  };
+
+  const { data: loginLocations = [] } = useQuery({
+    queryKey: ['providerLoginLocations'],
+    queryFn: async () => {
+      const currentUser = await getCurrentUser();
+      if (!currentUser) {
```
rahu1ramesh: This check can be done outside of the query, making the query enabled only if currentUser and currentUser.uuid is truthy.

apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -0,0 +1,511 @@
...
+  const filteredServices = useMemo(
+    () => services.filter((s) => s.location?.uuid === formData.locationUuid),
+    [services, formData.locationUuid],
+  );
+
+  const availableProviders = useMemo(
+    () =>
+      providers.filter(
+        (p) =>
+          !p.person?.voided &&
```
rahu1ramesh: I believe we can limit the search to non voided providers, and that this can be a service defined in the bahmni/services for better reuse.

apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -0,0 +1,511 @@
...
+
+    if (
+      !formData.locationUuid ||
+      !formData.startDate ||
+      !formData.startTime ||
+      !formData.endDate ||
+      !formData.endTime ||
+      dateTimeError
+    ) {
+      addNotification({
```
rahu1ramesh: This is not the current design pattern for validation that we follow, if a field fails validation, we show an inline error, no need for a validation error message.

apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -0,0 +1,511 @@
...
+          dateTimeError ||
+          t('ADMIN_UNAVAILABILITY_FORM_VALIDATION_ERROR_MESSAGE'),
+        type: 'error',
+      });
+      return;
+    }
+
+    setIsSubmitting(true);
+    try {
+      const baseData = {
```
rahu1ramesh: I would better prefer if we move this whole block into utils under a function that have it here.

apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -0,0 +1,511 @@
...
+            <DatePickerInput
+              id="start-date"
+              labelText={t('ADMIN_UNAVAILABILITY_FORM_START_DATE_LABEL')}
+              invalid={showValidation && !formData.startDate}
+              invalidText={t('ADMIN_UNAVAILABILITY_FORM_REQUIRED')}
+            />
+          </DatePicker>
+        </div>
+        <div className={styles.timeField}>
+          <TimePicker
```
rahu1ramesh, line 246: With css is it possible to keep them closer together in the UI? Reduce this space

Image

apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -0,0 +1,511 @@
...
+            <DatePickerInput
+              id="end-date"
+              labelText={t('ADMIN_UNAVAILABILITY_FORM_END_DATE_LABEL')}
+              invalid={showValidation && !formData.endDate}
+              invalidText={t('ADMIN_UNAVAILABILITY_FORM_REQUIRED')}
+            />
+          </DatePicker>
+        </div>
+        <div className={styles.timeField}>
+          <TimePicker
```
rahu1ramesh, line 311: Same feedback as for the start time

packages/bahmni-services/src/date/date.ts
```diff
@@ -513,3 +514,17 @@ export function sortByDate(
     return ascending ? diff : -diff;
   });
 }
+
+/**
+ * Converts 12-hour time format (with AM/PM) to 24-hour format
+ * @param timeStr - Time string in 12-hour format (e.g., "02:30 PM", "11:45 AM")
+ * @returns Time string in 24-hour format (e.g., "14:30", "11:45"), or empty string if invalid
+ */
+export function convertTo24HourFormat(timeStr: string): string {
```
rahu1ramesh, line 523: Do add test for the fn

rahu1ramesh, line 523: This can be done with 

```
format(parse(`${formData.startTime} ${formData.startTimePeriod}`, 'hh:mm aa', new Date()), 'HH:mm')
```

good to really confirm if we need this

packages/bahmni-services/src/locationService/locationService.ts
```diff
@@ -11,3 +11,22 @@ export async function getLocationByTag(tag: string): Promise<Location[]> {
...
+ * @returns Promise resolving to an array of Location objects
+ */
+export async function getFHIRLocationsByTag(tag: string): Promise<Location[]> {
+  const response = await get<FHIRBundle>(FHIR_LOCATION_BY_TAG_URL(tag));
+
+  if (!response.entry || response.entry.length === 0) {
+    return [];
+  }
+
+  return response.entry.map((entry) => ({
```
rahu1ramesh: Why don't we use the FHIR bundle it self? If needed the consumer can do this mapping.

packages/bahmni-services/src/providerService/providerService.ts
```diff
@@ -19,3 +20,73 @@ export async function getCurrentProvider(
 
   return providerResponse.results[0];
 }
+
+/**
+ * Fetches providers from OpenMRS (single page)
+ * @param url - Optional URL to fetch (defaults to ALL_PROVIDERS_URL)
+ * @returns Promise resolving to an array of Provider objects
+ */
+export async function getPaginatedProviders(
```
rahu1ramesh: Why not call this getAllProviders instead?

packages/bahmni-services/src/providerService/providerService.ts
```diff
@@ -19,3 +20,73 @@ export async function getCurrentProvider(
...
+  const response = await get<ProviderResponse>(url);
+  return response.results ?? [];
+}
+
+/**
+ * Fetches all providers from OpenMRS with pagination support
+ * Iterates through all pages until no "next" link is found
+ * @returns Promise resolving to an array of all Provider objects
+ */
+export async function fetchAllProviders(): Promise<Provider[]> {
```
rahu1ramesh, line 28: Both getPaginatedProviders and fetchAllProviders from the name sounds very similar, are they both the same. 

I don't see any actual usage of getPaginatedProviders, is this fn not needed?

apps/appointments/src/pages/admin/appointmentUnavailability/index.tsx
rahu1ramesh, line 1: Good to add a privilege check for this page.

apps/appointments/src/pages/admin/appointmentUnavailability/__mocks__/unavailabilityMock.ts
rahu1ramesh, line 1: rename this as unavailabilityMock file instead

apps/appointments/src/pages/admin/appointmentUnavailability/index.tsx
```diff
@@ -0,0 +1,157 @@
...
+      endDateTime: `${formatDateTime(item.endDate).formattedResult}, ${formatTime(item.endTime)}`,
+      locationName: item.location.name,
+      appointmentServiceName:
+        item.service?.name ?? t('ADMIN_UNAVAILABILITY_ALL'),
+      providerName: item.provider?.name ?? t('ADMIN_UNAVAILABILITY_ALL'),
+    }));
+  }, [data, t]);
+
+  const handleFormSuccess = useCallback(() => {
```
rahu1ramesh: This fn should ideally do the create call not the component, better to refactor this to avoid confusion

apps/appointments/src/pages/admin/appointmentUnavailability/__tests__/UnavailabilityForm.test.tsx
rahu1ramesh, line 1: Consider it.each for the UnavailabilityForm validation tests. They all follow the same "submit with missing X → show error Y" structure and are prime candidates.

apps/appointments/src/pages/admin/appointmentUnavailability/__tests__/UnavailabilityForm.test.tsx
```diff
@@ -0,0 +1,704 @@
...
+  ...jest.requireActual('@bahmni/widgets'),
+  useNotification: jest.fn(() => ({ addNotification: mockAddNotification })),
+}));
+
+jest.mock('@bahmni/design-system', () => {
+  const actual = jest.requireActual('@bahmni/design-system');
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const React = require('react');
+
+  const MockDatePicker = ({
```
rahu1ramesh: Do we need this mock?

apps/appointments/src/pages/admin/appointmentUnavailability/__tests__/UnavailabilityForm.test.tsx
```diff
@@ -0,0 +1,704 @@
...
+
+jest.mock('@bahmni/widgets', () => ({
+  ...jest.requireActual('@bahmni/widgets'),
+  useNotification: jest.fn(() => ({ addNotification: mockAddNotification })),
+}));
+
+jest.mock('@bahmni/design-system', () => {
+  const actual = jest.requireActual('@bahmni/design-system');
+  // eslint-disable-next-line @typescript-eslint/no-require-imports
+  const React = require('react');
```
rahu1ramesh: Absolutley no

apps/appointments/src/pages/admin/appointmentUnavailability/__tests__/UnavailabilityForm.test.tsx
```diff
@@ -0,0 +1,704 @@
...
+  ...jest.requireActual('@tanstack/react-query'),
+  useQuery: jest.fn(),
+}));
+
+jest.mock('@bahmni/widgets', () => ({
+  ...jest.requireActual('@bahmni/widgets'),
+  useNotification: jest.fn(() => ({ addNotification: mockAddNotification })),
+}));
+
+jest.mock('@bahmni/design-system', () => {
```
rahu1ramesh: Do not mock the design system

apps/appointments/src/pages/admin/appointmentUnavailability/__tests__/index.test.tsx
rahu1ramesh, line 1: Do not mock the design system

rahu1ramesh (changes requested):

packages/bahmni-services/src/providerService/providerService.ts
skjameela276: If we want to use pagination with getPaginatedProviders we can use this, for fetching all the providers fetchAllProviders. It will fetch all the providers @rahu1ramesh

packages/bahmni-services/src/providerService/providerService.ts
skjameela276: this service has the pagination

apps/appointments/src/pages/admin/appointmentUnavailability/constants.ts
skjameela276: for the system-x config we have 	dd-MMM-yyyy coming from the global properties available in local storage. But  the unvaialibility we need ISO format instead of defining constant here we can move that to date service constant. confirmed with @PratitiP

packages/bahmni-services/src/date/date.ts
skjameela276: This method can be reuse if we keep it bahmni/services

apps/appointments/src/pages/admin/appointmentUnavailability/index.tsx
skjameela276: @mohan-13 please confirm is this privilege check needed for the entire page since we already have the privilege check to view and add the unavailabilities.

apps/appointments/src/pages/admin/appointmentUnavailability/constants.ts
skjameela276: why can't we keep it here ?

apps/appointments/src/pages/admin/appointmentUnavailability/constants.ts
skjameela276: why can't we keep it here ?

apps/appointments/src/pages/admin/appointmentUnavailability/constants.ts
mohan-13: @rahu1ramesh In this case, this is okay as this is an Admin functionality and anyways the API expects minimally atleast these privileges to be present for the logged in user. So this is okay. 

If this needs to be configurable we can bring that later.

apps/appointments/src/pages/admin/appointmentUnavailability/constants.ts
mohan-13: @rahu1ramesh In this case, this is okay as this is an Admin functionality and anyways the API expects minimally atleast these privileges to be present for the logged in user. So this is okay. 

If this needs to be configurable we can bring that later.

apps/appointments/src/pages/admin/appointmentUnavailability/index.tsx
mohan-13: This page only shows the dataTable for now and that API is already guarded without the Get Privilege no data can be retrieved / shown. So this is okay.

rahu1ramesh (approved):

PR #429, opened 2026-05-25
BAH-4721 | Fix. Remove Vaccination-Specific STAT And Frequency Hardcoding

rahu1ramesh:
JIRA → [BAH-4721](https://bahmni.atlassian.net/browse/BAH-4721)

 ## Description

Removes hardcoded vaccination-specific behaviour from the MedicationRequest form. Previously, isSTAT was forced to true for vaccinations regardless of configuration, and the frequency dropdown excluded "immediate" frequencies when STAT was active. These assumptions are now driven purely by attribute configuration (i.e. stat.default = true), making the form consistent across all input control types.

[BAH-4721]: https://bahmni.atlassian.net/browse/BAH-4721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Medication duration, duration unit, and start date now reset when toggling STAT/PRN; immediate frequencies are available in the frequency dropdown and can be applied as defaults.
  * New entries inherit STAT only from configured defaults (STAT no longer forced on creation).

* **Tests**
  * Updated STAT defaulting tests to reflect configurable defaults (STAT defaults to false unless configured).
  * Added/adjusted vaccination and default-frequency tests to verify immediate-frequency, duration, and start-date behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Bahmni/bahmni-apps-frontend/pull/429?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/apps/clinical/src/components/forms/medicationRequest/components/SelectedMedicationRequestItem.tsx
+++ b/apps/clinical/src/components/forms/medicationRequest/components/SelectedMedicationRequestItem.tsx
@@ -287,7 +287,7 @@ const SelectedMedicationRequestItem: React.FC<SelectedMedicationRequestItemProps
                 )}
                 aria-label="Frequency"
                 hideLabel
-                items={medicationConfig.frequencies ?? []}
+                items={medicationConfig.frequencies}
                 itemToString={(item) => (item ? item.name : '')}
                 selectedItem={frequency}
                 onChange={(e) => {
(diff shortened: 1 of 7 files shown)


apps/clinical/src/components/forms/medicationRequest/components/SelectedMedicationRequestItem.tsx
```diff
@@ -289,11 +287,7 @@ const SelectedMedicationRequestItem: React.FC<SelectedMedicationRequestItemProps
                 )}
                 aria-label="Frequency"
                 hideLabel
-                items={
```
mohan-13, line 292: We should not remove this check, Immediately is a default case in specific scenarios. If the user changes they can set to some other values

rahu1ramesh: If we don't have it here, default will not be set, better to keep it here for the intended behavior.

mohan-13 (approved):

PR #442, opened 2026-05-28
BAH-4754 Command Palette for Bahmni

lingeswaranTW:
## Summary by CodeRabbit

* **New Features**
  * Added Command Palette for quick app navigation and patient search accessible via keyboard shortcut (Cmd/Ctrl + K)
  * Enables patient filtering with custom search annotations and quick-access patient actions
  * Includes localization support for English and Spanish


packages/bahmni-widgets/src/commandPalette/styles/CommandPalette.module.scss
```diff
@@ -0,0 +1,423 @@
...
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  padding: 1px $spacing-02;
+  background: $layer-03;
+  border: 1px solid $border-subtle-02;
+  border-radius: 4px;
+  color: $text-helper;
+  @include type-style('code-01');
+  font-size: 0.6875rem;
```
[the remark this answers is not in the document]
lingeswaranTW: Thanks for the suggestion, but after checking the project configuration, this doesn't appear to be a real issue:
  
  1. declaration-empty-line-before — this rule is explicitly set to null (disabled) in the project's stylelint.config.js (line 32), so it cannot produce violations.
  2. Keyframe naming — stylelint is not installed as a dependency in this project, and packages/bahmni-widgets has no lint target configured. The stylelint.config.js file glob (src/**/*.{scss}) also doesn't resolve to this package's path.
  There is no CI step that would flag these names.

  The camelCase keyframe names (fadeIn, slideIn, ekgScroll) are valid CSS and won't cause any CI failures. No changes needed here.

apps/command-palette/vite.config.ts
```diff
@@ -0,0 +1,47 @@
...
+    },
+    lib: {
+      entry: 'src/index.ts',
+      name: '@bahmni/command-palette',
+      fileName: 'index',
+      formats: ['es' as const],
+    },
+    rollupOptions: {
+      external: [
+        'react',
```
ravinderkabli, line 39: blocking: `@bahmni/widgets` and `@bahmni/services` are not in the `external` list. In production, the distro webpack resolves `@bahmni/command-palette` from its pre-built `dist/index.js` (the alias to source only applies in dev mode). Since these packages aren't externalized, Vite bundles them into command-palette's dist — creating a separate copy of `CommandPaletteContext`.

This means `useCommandPalette()` imported from `@bahmni/widgets` by any widget would read from a different context object than the one where state actually lives, and throw at runtime.

Add both:
```ts
external: [
  'react', 'react-dom', 'react/jsx-runtime', 'react-router-dom',
  /^@carbon\/styles/,
  '@bahmni/widgets',
  '@bahmni/services',
]
```

apps/command-palette/src/utils/extensionUtils.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const basePathFromTemplate = (template: string): string =>
+  template.split('{{')[0].replace(/\/$/, '');
+
+export const pathTemplateToGetPath =
+  (template: string) =>
+  ({ patientUuid, patientIdentifier = '' }: PatientActionContext): string =>
+    formatUrl(template, { patientUuid, patientIdentifier });
+
+export const resolveLabel = (e: CommandPaletteExtension): string =>
```
ravinderkabli, line 13: blocking: `label ?? translationKey` gives `label` precedence, but `docs/command-palette-guide.md` (line 177) says "If both are present, `translationKey` takes precedence."

This matters for i18n: when `label` wins, the literal string (e.g., "Registration") is passed to `t()` — which works for English but breaks other locales since there's no matching locale key.

If `translationKey` should take precedence (recommended for i18n):
```ts
export const resolveLabel = (e: CommandPaletteExtension): string =>
  e.translationKey ?? e.label ?? '';
```

packages/bahmni-widgets/src/commandPalette/CommandPalette.tsx
```diff
@@ -0,0 +1,325 @@
...
+          setSearchTerm('');
+          return;
+        }
+      }
+      setSearchTerm(value);
+    },
+    [selectedAnnotation, searchAnnotations],
+  );
+
+  const handleInputKeyDown = useCallback(
```
ravinderkabli, line 89: suggestion: The ArrowLeft/ArrowRight handler queries `document.querySelector('[data-patient-uuid][data-selected="true"]')` to find the focused patient. This creates a fragile coupling — `data-selected="true"` is a `cmdk` internal attribute that could change between versions, and the patient UUID is already tracked in React state via `handleCmdValueChange`.

Consider using `selectedPatientUuid` from state directly instead of the DOM round-trip.

packages/bahmni-widgets/src/commandPalette/useDebounce.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from 'react';
+
+export function useDebounce<T>(value: T, delay: number = 500): T {
```
ravinderkabli, line 3: suggestion: An identical `useDebounce` hook already exists at `apps/clinical/src/hooks/useDebounce.ts`. Consider exporting this from `@bahmni/widgets` and having both consumers import from the shared location to avoid drift.

packages/bahmni-widgets/src/commandPalette/CommandPaletteProvider.tsx
```diff
@@ -0,0 +1,66 @@
...
+  searchAnnotations,
+}) => {
+  const [isOpen, setIsOpen] = useState(false);
+  const toggle = useCallback(() => setIsOpen((prev) => !prev), []);
+
+  useCommandPaletteKeyboard(isOpen, toggle, trigger);
+
+  return (
+    <CommandPaletteContext.Provider
+      value={{
```
ravinderkabli: suggestion: The `value` object is constructed inline, so React creates a new reference on every render — triggering context re-renders in all consumers even when no values changed.

Wrap in `useMemo`:
```tsx
const contextValue = useMemo(() => ({
  isOpen, toggle, setOpen: setIsOpen,
  navItems, patientActions, patientFieldsConfig, searchAnnotations,
}), [isOpen, toggle, navItems, patientActions, patientFieldsConfig, searchAnnotations]);
```

apps/command-palette/src/hooks/useCommandPaletteConfig.ts
```diff
@@ -0,0 +1,133 @@
...
+          : [];
+
+      if (appConfig) {
+        const cp = appConfig.commandPalette;
+        if (cp?.trigger) setTrigger(cp.trigger);
+        if (cp?.patientFields) setPatientFieldsConfig(cp.patientFields);
+        if (cp?.searchAnnotations) setSearchAnnotations(cp.searchAnnotations);
+      }
+
+      const currentPath = window.location.pathname;
```
ravinderkabli: suggestion: `window.location.pathname` is read once at mount (empty `[]` dep array). Since `CommandPaletteProvider` sits above `<Routes>` and mounts once, `appContext`-filtered items will be stale after SPA navigation.

Consider using `useLocation()` from react-router-dom and adding `location.pathname` to the dependency array, or re-evaluating the filter when the palette opens.

ravinderkabli (changes requested):
## Review for BAH-4754

Solid implementation — clean separation between config and UI layers, good test coverage, thorough documentation, and well-thought-out extensibility. Two issues need fixing before merge.

apps/command-palette/vite.config.ts
lingeswaranTW: Resolved

apps/command-palette/src/utils/extensionUtils.ts
lingeswaranTW: yeah,Resolved

packages/bahmni-widgets/src/commandPalette/CommandPalette.tsx
lingeswaranTW: Resolved

packages/bahmni-widgets/src/commandPalette/useDebounce.ts
lingeswaranTW: Resolved-exported from widgets in future we can use this in other places-

packages/bahmni-widgets/src/commandPalette/CommandPaletteProvider.tsx
lingeswaranTW: resolved

apps/command-palette/src/hooks/useCommandPaletteConfig.ts
lingeswaranTW: resolved

ravinderkabli (approved):
Looks fine

PR #445, opened 2026-06-01
BAH-4727|Fix. form save event script

Aishwarya-thoughtworks:
Problem

  Form save event scripts were failing with:
  Error in onFormSave event for form "History and Examination": n.getConceptName is not a function
  This prevented validation scripts from executing on forms like:
  - History and Examination (Chief Complaint validation)

  Root Cause -Why immutable.js v3→v4 upgrade broke form save events:
  Form save event scripts receive a FormContext object that holds form data as immutable ControlRecord objects. The script calls methods like:
  form.get('Chief Complaint Coded', n).getValue()
    → requires ControlRecord.getValue() method

  Immutable.js v3.8.1:
  ControlRecord.toJS() → partial conversion, nested children retained methods 

  Immutable.js v4.3.8:
  Record.prototype.toJS = function toJS() {
    return toJS(this);  // ← Deep recursive conversion of ALL nested structures
  }
  
  ControlRecord.toJS() → FULLY converted to plain JS objects, all methods stripped 
  
  What happened after the upgrade:
  Form data that was being converted with convertImmutableToPlainObject() (which calls .toJS()) lost all ControlRecord methods:
  Solution: Pass the raw immutable ControlRecord directly to executeOnFormSaveEvent instead of the converted plain object:

  // Before
  const formData = convertImmutableToPlainObject(containerState?.data);
  executeOnFormSaveEvent(..., formData);

  // After
  executeOnFormSaveEvent(..., containerState?.data);

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of observation form save and validation to reduce intermittent save errors.

* **Chores**
  * Updated form controls dependency to a newer patch/dev version for underlying component improvements.


ravinderkabli (approved):
Looks fine

PR #446, opened 2026-06-01
BAH-4760 | Fix. CSS Bleed Caused By Missing Hash Policy And Scoping

rahu1ramesh:
**JIRA** → [BAH-4760](https://bahmni.atlassian.net/browse/BAH-4760)

## **Description**

Fixes CSS bleed between micro-frontends caused by unscoped CSS module class names and Carbon Design System styles being bundled into each package. CSS module class names were colliding at runtime because Vite was not generating hashed, scoped names during library builds. Additionally, `@carbon/styles` was being bundled into each package, causing duplicate and conflicting global style injections.

## **Key Features**

  - CSS module class names are now deterministically scoped with a hash (`[name]_[local]__[hash:base64:5]`) across all apps and packages during library builds.
  - `@carbon/styles` is externalized from all packages and apps so it is loaded once by the host, not duplicated per micro-frontend.

[BAH-4760]: https://bahmni.atlassian.net/browse/BAH-4760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/clinical/vite.config.ts
+++ b/apps/clinical/vite.config.ts
@@ -49,6 +49,7 @@ export default defineConfig(() => ({
         'react-dom',
         'react/jsx-runtime',
         'react-router-dom',
+        '@tanstack/react-query',
         /^@carbon\/styles/
       ],
     },
--- a/apps/clinical/vite.config.ts
+++ b/apps/clinical/vite.config.ts
@@ -2,7 +2,7 @@
 import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
 import dts from 'vite-plugin-dts';
-import * as path from 'path';
+import * as path from 'node:path';
 
 export default defineConfig(() => ({
   root: __dirname,
@@ -15,6 +15,11 @@ export default defineConfig(() => ({
       tsconfigPath: path.join(__dirname, 'tsconfig.lib.json'),
     }),
   ],
+  css: {
+    modules: process.env['VITEST'] ? {} : {
+      generateScopedName: '[name]_[local]__[hash:base64:5]',
+    },
+  },
   // Uncomment this if you are using workers.
   // worker: {
   //  plugins: [ nxViteTsPaths() ],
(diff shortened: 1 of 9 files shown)


apps/clinical/vite.config.ts
```diff
@@ -39,8 +44,14 @@ export default defineConfig(() => ({
...
     rollupOptions: {
-      // External packages that should not be bundled into your library.
-      external: ['react', 'react-dom', 'react/jsx-runtime', 'react-router-dom'],
+      external: [
+        'react',
+        'react-dom',
+        'react/jsx-runtime',
+        'react-router-dom',
+        '@tanstack/react-query',
+        /^@carbon\/styles/
```
mohan-13, line 53: is the slash escaping needed here? i can see te @tanstanc/react-query added without escaping ? can we be consistent

rahu1ramesh: `@tanstanc/react-query` is imported as is, while '@carbon/styles' might be used as @carbon/styles/css/all.css. The regex helps match any such subpaths, recommended to keep it a regex and not a string.

mohan-13:
This changes seem to break the UI when the container is built.
image

mohan-13 (changes requested):

packages/bahmni-design-system/src/index.ts
```diff
@@ -1,5 +1,3 @@
-import '@carbon/styles/css/styles.css';
```
mohan-13, line 1: Why is this import removed ?

rahu1ramesh: Add it back in, was a shortcut to avoid carbon from taking precedence, but seems to be breaking product build. Will fix it on the consumer distro.

mohan-13:
> This changes seem to break the UI when the container is built. image

This is resolved after restoring the import.

mohan-13 (approved):

PR #448, opened 2026-06-02
BAH-4711 | Fix. Default Select Logged In Location In Unavailability Form

skjameela276:
**JIRA** → [BAH-4711](https://bahmni.atlassian.net/browse/BAH-4711)

## Summary
- Restrict default location selection in the Add Unavailability form to only the user's logged-in location when it exists in `loginLocations`; otherwise leave the field empty so the user must pick one explicitly.
- Add a separate `ADMIN_UNAVAILABILITY_FORM_LOCATION_PLACEHOLDER` translation key (English + Spanish) so the dropdown placeholder ("Select Location") is distinct from the required-field label ("Select Location*").
- Add tests covering the three reconciliation paths (cookie location is valid, cookie location not in `loginLocations`, empty `loginLocations`) and refresh the snapshot.

[BAH-4711]: https://bahmni.atlassian.net/browse/BAH-4711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Localization**
  * Removed the visible "*" from the location field label and moved the required indicator into the field title/tooltip in English and Spanish.

* **Bug Fixes**
  * Improved location handling so the form only retains/selects a location when it exists in the current available locations.

* **Tests**
  * Added tests covering location-selection behavior based on saved and available locations.


apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -181,7 +191,7 @@ const UnavailabilityForm: React.FC<UnavailabilityFormProps> = ({
         <Dropdown
           id="location-dropdown"
           data-testid="location-dropdown"
-          label={t('ADMIN_UNAVAILABILITY_FORM_LOCATION_LABEL')}
+          label={t('ADMIN_UNAVAILABILITY_FORM_LOCATION_PLACEHOLDER')}
```
rahu1ramesh: For the label, better to use the label key itself and update the below used "ADMIN_UNAVAILABILITY_FORM_LOCATION_LABEL" as title?

rahu1ramesh (changes requested):

apps/appointments/src/pages/admin/appointmentUnavailability/components/UnavailabilityForm.tsx
```diff
@@ -123,6 +123,16 @@ const UnavailabilityForm: React.FC<UnavailabilityFormProps> = ({
     [availableProviders],
   );
 
+  useEffect(() => {
+    if (loginLocations.length === 0) return;
+    const loggedinUuid = getInitialLocationUuid();
+    const isValid = loginLocations.some((l) => l.uuid === loggedinUuid);
```
rahu1ramesh: isValid is a slightly ambiguous constant name, would be great if you can add a better name.

rahu1ramesh (approved):

PR #459, opened 2026-06-07
BAH-4741 | Fix. data range filter and minor styles fixes for select filter

skjameela276:
## Summary by CodeRabbit

* **Style**
  * Improved visual styling and layout for table filters and refined empty-state padding and alignment.
* **Bug Fixes**
  * Date-range filtering now treats the end date as inclusive so full end-day timestamps are included.
* **Behavior**
  * Select filter shows item titles more consistently and date filter layout is clearer.
* **Tests**
  * Added a test verifying inclusive same-day date-range behavior.


packages/bahmni-design-system/src/organisms/dataTable/utils.ts
```diff
@@ -25,6 +25,15 @@ const toTimestamp = (value: unknown): number | null => {
   return null;
 };
 
+// Helper function to get end of day timestamp in UTC
+const getEndOfDay = (timestamp: number): number => {
+  const date = new Date(timestamp);
+  const year = date.getUTCFullYear();
```
mohan-13: Are we assuming the timestamp will always be in UTC ?

mohan-13 (approved):

PR #460, opened 2026-06-08
BAH-4756.fix(frontend): allow SPA routing for top-level app paths tha…

arshiyaTW2021:
**JIRA** → [BAH-4756](https://bahmni.atlassian.net/browse/BAH-4756?atlOrigin=eyJpIjoiNWM5MjdmYmUzYWIxNDA5Njk3Mzc4YTlmYThmZTlmYzMiLCJwIjoiaiJ9)

### **Description**

Remove the `RewriteCond %{REQUEST_FILENAME} !-d` condition from the SPA rewrite configuration in httpd_directory.conf

### **Problem**

The Bahmni Apps Frontend build output contains directories such as:

home/
registration/
clinical/
appointments/

When navigating to /bahmni-new/home, Apache resolves home as a physical directory. Because the rewrite rule only applies when the request is neither a file nor a directory, the SPA fallback is skipped.

As a result, Apache attempts to serve the directory directly and returns:

403 Forbidden
Cannot serve directory .../home/: No matching DirectoryIndex found
This prevents React Router from handling the route.

### **Root Cause**

The following condition blocks SPA fallback for requests that match existing directories:

`RewriteCond %{REQUEST_FILENAME} !-d`

Since **home/** exists in the build output, requests to **/bahmni-new/home** are treated as directory requests rather than application routes.

### **Change**

Removed:
RewriteCond %{REQUEST_FILENAME} !-d

### **Result**

Removing RewriteCond %{REQUEST_FILENAME} !-d allows requests such as /bahmni-new/home to resolve to the SPA entry page instead of being treated as directory requests and returning 403.

[BAH-4756]: https://bahmni.atlassian.net/browse/BAH-4756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SPA routing so static file requests are correctly served while client-side routes fallback to index.html, reducing 404s and improving navigation stability.


ravinderkabli (approved):
Approved. Arshiya had validated approach with Mohan

PR #464, opened 2026-06-08
BAH-4725 | Frontend support for Patient.photo via FHIR R4

Varun-beta:
## Summary
- Use `photo[0].url` from the FHIR Patient response to fetch patient photos instead of hardcoded URLs
- Photo is only fetched when the backend confirms it exists (no `photo` field = no photo on disk)
- Remove legacy `getPatientPhotoDataUrl` and `PATIENT_IMAGE_URL` — replaced by `fetchPatientPhotoFromUrl`
- DOB estimated checkbox now only checks when typing in the **years** field, not months/days

### Photo — `@bahmni/services`
- **`patientService.ts`** — Added `fetchPatientPhotoFromUrl(url)` that takes the URL directly from FHIR response. Removed `getPatientPhotoDataUrl`. Added `photoUrl` to `formatPatientData` output.
- **`models.ts`** — Added `photoUrl` to `FormattedPatientData` interface
- **`constants.ts`** — Removed `PATIENT_IMAGE_URL` (no longer needed)
- **`index.ts`** — Updated exports

### Photo — `@bahmni/widgets`
- **`PatientDetails.tsx`** — Uses `patient.photoUrl` from `FormattedPatientData` instead of constructing URL from UUID. Guarded by `Get Patient Photo` privilege via `useHasPrivilege`.
- **`PatientDetails.integration.test.tsx`** — Updated to mock `fetchPatientPhotoFromUrl` and `useHasPrivilege`

### Photo — `@bahmni/registration-app`
- **`usePatientPhoto.ts`** — Accepts `photoUrl` (from FHIR response) instead of `patientUuid`. Only fetches when URL exists and user has privilege.
- **`PatientRegister.tsx`** — Extracts `photoUrl` from `patientDetails?.photo?.[0]?.url`
- **`fhirPatientMapper.ts`** — Maps `photo` field in FHIR patient payload

### DOB Estimated fix
- **`dateAgeUtils.ts`** — `setDobEstimated(field === 'ageYears')` — estimated checkbox only checked when typing years, not months/days

## Flow
1. `GET /Patient/{uuid}` → backend includes `photo[0].url` only if photo exists on disk
2. Frontend extracts `photoUrl` from response → calls `fetchPatientPhotoFromUrl(url)` → displays image
3. No photo on disk → no `photo` field → no API call → placeholder shown

## Depends on
- Backend PR: https://github.com/Bahmni/bahmni-module-fhir2-addl-extension/pull/81

## Test plan
- [x] `@bahmni/services` — 117 tests pass
- [x] `@bahmni/widgets` — 25 tests pass (unit + integration)
- [x] `@bahmni/registration-app` — usePatientPhoto (3), fhirPatientMapper (18), Profile (44), PatientRegister (51), PatientPhotoUpload (35) — all pass

## Release Notes

* **New Features**
  * Added photo upload capability to the patient registration form
  * Implemented privilege-based access control for patient photos—users must have the "Get Patient Photo" privilege to view patient images

* **Improvements**
  * Updated photo retrieval mechanism to use direct URLs for more efficient data fetching
  * Enhanced patient data model to include photo URLs from patient records

### DIFF
--- a/apps/registration/src/components/forms/profile/__tests__/Profile.test.tsx
+++ b/apps/registration/src/components/forms/profile/__tests__/Profile.test.tsx
@@ -422,29 +422,37 @@ describe('Profile', () => {
       expect(checkbox.checked).toBe(true);
     });
 
-    it('should set dobEstimated only when years field is changed, not months or days', () => {
-      const { unmount } = render(
-        <Profile ref={ref} initialDobEstimated={false} />,
-      );
+    it('should have dobEstimated true when initialized with age years', async () => {
+      await act(async () => {
+        render(
+          <Profile
+            ref={ref}
+            initialData={createBasicInfoData({ ageYears: '30' })}
+            initialDobEstimated
+          />,
+        );
+      });
 
-      // Verify initial state
+      const data = ref.current?.getData();
+      expect(data?.dobEstimated).toBe(true);
+    });
+
+    it('should not set dobEstimated when only months or days are changed', () => {
+      render(<Profile ref={ref} initialDobEstimated={false} />);
+
       expect(ref.current?.getData()?.dobEstimated).toBe(false);
 
-      // Type in months — should NOT set estimated
       const monthsInput = screen.getByLabelText(
         /Months\(Age\)/,
       ) as HTMLInputElement;
       fireEvent.change(monthsInput, { target: { value: '6' } });
       expect(ref.current?.getData()?.dobEstimated).toBe(false);
 
-      // Type in days — should NOT set estimated
       const daysInput = screen.getByLabelText(
         /Days\(Age\)/,
       ) as HTMLInputElement;
       fireEvent.change(daysInput, { target: { value: '15' } });
       expect(ref.current?.getData()?.dobEstimated).toBe(false);
-
-      unmount();
     });
   });
 
--- a/apps/registration/src/components/forms/profile/__tests__/Profile.test.tsx
+++ b/apps/registration/src/components/forms/profile/__tests__/Profile.test.tsx
@@ -423,18 +423,24 @@ describe('Profile', () => {
     });
 
     it('should set dobEstimated only when years field is changed, not months or days', () => {
-      const { unmount } = render(<Profile ref={ref} initialDobEstimated={false} />);
+      const { unmount } = render(
+        <Profile ref={ref} initialDobEstimated={false} />,
+      );
 
       // Verify initial state
       expect(ref.current?.getData()?.dobEstimated).toBe(false);
 
       // Type in months — should NOT set estimated
-      const monthsInput = screen.getByLabelText(/Months\(Age\)/) as HTMLInputElement;
+      const monthsInput = screen.getByLabelText(
+        /Months\(Age\)/,
+      ) as HTMLInputElement;
       fireEvent.change(monthsInput, { target: { value: '6' } });
       expect(ref.current?.getData()?.dobEstimated).toBe(false);
 
       // Type in days — should NOT set estimated
-      const daysInput = screen.getByLabelText(/Days\(Age\)/) as HTMLInputElement;
+      const daysInput = screen.getByLabelText(
+        /Days\(Age\)/,
+      ) as HTMLInputElement;
       fireEvent.change(daysInput, { target: { value: '15' } });
       expect(ref.current?.getData()?.dobEstimated).toBe(false);
 
--- a/apps/registration/src/components/forms/profile/__tests__/Profile.test.tsx
+++ b/apps/registration/src/components/forms/profile/__tests__/Profile.test.tsx
@@ -421,6 +421,25 @@ describe('Profile', () => {
 
       expect(checkbox.checked).toBe(true);
     });
+
+    it('should set dobEstimated only when years field is changed, not months or days', () => {
+      const { unmount } = render(<Profile ref={ref} initialDobEstimated={false} />);
+
+      // Verify initial state
+      expect(ref.current?.getData()?.dobEstimated).toBe(false);
+
+      // Type in months — should NOT set estimated
+      const monthsInput = screen.getByLabelText(/Months\(Age\)/) as HTMLInputElement;
+      fireEvent.change(monthsInput, { target: { value: '6' } });
+      expect(ref.current?.getData()?.dobEstimated).toBe(false);
+
+      // Type in days — should NOT set estimated
+      const daysInput = screen.getByLabelText(/Days\(Age\)/) as HTMLInputElement;
+      fireEvent.change(daysInput, { target: { value: '15' } });
+      expect(ref.current?.getData()?.dobEstimated).toBe(false);
+
+      unmount();
+    });
   });
 
   describe('Patient ID Format Selection', () => {
--- a/apps/registration/src/components/forms/profile/dateAgeUtils.ts
+++ b/apps/registration/src/components/forms/profile/dateAgeUtils.ts
@@ -182,7 +182,7 @@ export const createDateAgeHandlers = <
         if (age.years > 0 || age.months > 0 || age.days > 0) {
           const birthISO = AgeUtils.calculateBirthDate(age);
           updated.dateOfBirth = birthISO;
-          setDobEstimated(true);
+          setDobEstimated(field === 'ageYears');
 
           const maxAgeDate = new Date();
           maxAgeDate.setFullYear(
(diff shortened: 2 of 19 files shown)


apps/registration/src/components/forms/profile/dateAgeUtils.ts
```diff
@@ -182,7 +182,7 @@ export const createDateAgeHandlers = <
         if (age.years > 0 || age.months > 0 || age.days > 0) {
           const birthISO = AgeUtils.calculateBirthDate(age);
           updated.dateOfBirth = birthISO;
-          setDobEstimated(true);
+          setDobEstimated(field === 'ageYears');
```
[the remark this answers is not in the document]
Varun-beta: Added test coverage — verifies that typing in months or days fields does NOT set dobEstimated, only the years field does. The behavior is intentional: only a rough year-based age is an estimate, an exact month/day entry is precise.

apps/registration/src/utils/fhirPatientMapper.ts
```diff
@@ -213,6 +214,9 @@ export function buildFhirPatient(input: MapperInput): Patient {
     }),
     ...(extensions.length > 0 && { extension: extensions }),
     ...(fhirAddresses.length > 0 && { address: fhirAddresses }),
+    photo: profile.image
+      ? [{ contentType: 'image/jpeg', data: profile.image }]
+      : [],
```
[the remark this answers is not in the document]
Varun-beta: This is intentional. The frontend always explicitly sends `photo: []` in every payload — verified from actual create/update payloads. When the user has no photo or deletes it, `profile.image` is undefined and `photo: []` signals the backend to delete the existing photo. The backend handles this correctly — it checks if a photo exists on disk before deleting.

packages/bahmni-widgets/src/patientDetails/__tests__/PatientDetails.test.tsx
```diff
@@ -107,6 +115,21 @@ describe('PatientDetails Component', () => {
...
       );
     });
+
+    it('does not fetch photo when user lacks privilege', () => {
+      mockUseHasPrivilege.mockReturnValue(false);
+      mockPatientQuery({
+        data: createMockPatient(),
+        isLoading: false,
+        error: null,
+      });
```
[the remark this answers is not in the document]
Varun-beta: Good catch. Fixed — the test now passes `photoUrl` in the mock patient data so it actually validates the privilege gate rather than passing due to the missing photo URL.

apps/registration/src/utils/fhirPatientMapper.ts
ravinderkabli, line 219: blocking: `photo: []` is always sent when `profile.image` is falsy (user didn't touch the photo widget). On a PUT, this will silently delete the patient's existing photo on the server.

Other optional arrays in this mapper use conditional spreading to avoid this:
```ts
...(fhirAddresses.length > 0 && { address: fhirAddresses }),
```

Suggested fix — omit `photo` entirely when no image was provided:
```ts
...(profile.image && {
  photo: [{ contentType: 'image/jpeg', data: profile.image }],
}),
```

For the explicit delete case (user clicks the cross button), you may need a sentinel to distinguish "delete" (`photo: []`) from "not touched" (omit field).

packages/bahmni-services/src/patientService/patientService.ts
```diff
@@ -189,16 +189,10 @@ export const getFormattedPatientById = async (
...
- * @param patientUUID - The UUID of the patient whose photo to fetch
- * @returns A base64 data URL string for the patient photo
- */
-export const getPatientPhotoDataUrl = async (
-  patientUUID: string,
+export const fetchPatientPhotoFromUrl = async (
+  photoUrl: string,
 ): Promise<string> => {
-  const blob = await get<Blob>(PATIENT_IMAGE_URL(patientUUID), {
+  const blob = await get<Blob>(photoUrl, {
```
ravinderkabli, line 198: suggestion: `photoUrl` comes directly from the FHIR response and is passed to `get()` without any origin validation. The old `getPatientPhotoDataUrl` always used a known hardcoded path. If a misconfigured server returns an arbitrary URL, the browser will make a credentialed request to that origin.

Consider adding a guard:
```ts
if (!photoUrl.startsWith('/')) {
  throw new Error('Photo URL must be a relative path');
}
```

apps/registration/src/hooks/usePatientPhoto.ts
```diff
@@ -1,15 +1,19 @@
-import { getPatientPhotoDataUrl } from '@bahmni/services';
+import { fetchPatientPhotoFromUrl } from '@bahmni/services';
+import { useHasPrivilege } from '@bahmni/widgets';
 import { useQuery } from '@tanstack/react-query';
 
+const GET_PATIENT_PHOTO_PRIVILEGE = 'Get Patient Photo';
```
ravinderkabli: suggestion: This string is duplicated here and in `PatientDetails.tsx`. The codebase already has a pattern for shared privilege constants (`CONSULTATION_PAD_PRIVILEGES` in `consultationPadPrivileges.ts`). Consider adding this to a shared constants file exported from `@bahmni/widgets` so both consumers import from one source.

apps/registration/src/components/forms/profile/__tests__/Profile.test.tsx
```diff
@@ -421,6 +421,31 @@ describe('Profile', () => {
 
       expect(checkbox.checked).toBe(true);
     });
+
+    it('should set dobEstimated only when years field is changed, not months or days', () => {
```
ravinderkabli: suggestion: This test covers the negative cases (months/days do NOT set `dobEstimated`) but not the positive case. Adding a `fireEvent.change` on the years input with an assertion that `dobEstimated === true` would catch a regression where `setDobEstimated(field === 'ageYears')` always returns `false`.

packages/bahmni-services/src/patientService/models.ts
```diff
@@ -12,6 +12,7 @@ export interface FormattedPatientData {
   formattedAddress: string | null;
   formattedContact: string | null;
   identifiers: Map<string, string>;
+  photoUrl: string | undefined;
```
ravinderkabli: nit: `photoUrl: string | undefined` makes the key required in object literals — every caller must explicitly write `photoUrl: undefined`. This causes TS errors in test mocks (e.g., `createMockPatient` omits it, hidden by babel-jest). Since a patient may have no photo URL, `photoUrl?: string` (optional property) is more appropriate.

apps/registration/src/utils/fhirPatientMapper.ts
Varun-beta: Verified — HAPI FHIR cannot distinguish between 'photo key absent' and 'photo: []' in the JSON payload. Both result in `getPhoto().isEmpty() = true` after deserialization. Tested with `FhirContext.newJsonParser()` — `hasPhoto()`, `getPhoto().size()`, even re-serialization give identical results for both cases.

So the frontend must always send the `photo` key explicitly: `photo: [{data}]` to save, `photo: []` to delete. Omitting the key would be indistinguishable from delete intent on the backend. The current implementation is the correct approach.

packages/bahmni-services/src/patientService/patientService.ts
Varun-beta: Fixed — added `if (!photoUrl.startsWith('/')) throw new Error('Photo URL must be a relative path')` guard in `fetchPatientPhotoFromUrl`.

apps/registration/src/hooks/usePatientPhoto.ts
Varun-beta: Fixed — extracted to `patientPhotoPrivileges.ts` in `@bahmni/widgets/src/userPrivileges/`, exported from `@bahmni/widgets`. Both `usePatientPhoto.ts` and `PatientDetails.tsx` now import from the shared constant.

apps/registration/src/components/forms/profile/__tests__/Profile.test.tsx
Varun-beta: Fixed — added positive test case that verifies `dobEstimated` is `true` when initialized with age years. Also kept the negative test verifying months/days don't set estimated.

packages/bahmni-services/src/patientService/models.ts
Varun-beta: Fixed — changed `photoUrl: string | undefined` to `photoUrl?: string` in `FormattedPatientData`.

ravinderkabli (approved):
Looks fine

PR #479, opened 2026-06-09
Fix. Use DELETE+POST for same-session allergy reaction rem…

sahupadma:
Fix: Reaction removal not working for same-session allergy edits                                                     
                                                                                                                                                                               
  Problem                                                                                                                                                                      
                                                                                                                                                                               
  When editing an allergy that was recorded in the current consultation (same encounter), removing a reaction had no effect — the removed reaction persisted after save.       
                                                                                                                                                                               
  The same-session edit path used a PUT request. OpenMRS FHIR2 does not replace reactions on PUT — it appends to existing reactions. This means sending fewer reactions in the 
  PUT body does not remove the ones left off. Confirmed via direct Bruno test against the FHIR endpoint.                                                                     
                                                                                                                                                                               
  The cross-session path (allergy from a previous visit) already worked correctly because it uses DELETE+POST, which wipes the old resource and creates a fresh one with only  
  the selected reactions.
                                                                                                                                                                               
  Fix                                                                                                                                                                        

  In createAllergiesBundleEntries, before issuing a same-session PUT, compare the original reaction codes from rawFhirResource against the user's current selection:           
   
  const manifestationUUIDSet = new Set(manifestationUUIDs);                                                                                                                    
  const reactionsRemoved = [...existingManifestationByCode.keys()].some(                                                                                                     
    (code) => !manifestationUUIDSet.has(code),                                                                                                                                 
  );
                                                                                                                                                                               
  - No reactions removed → PUT (safe, OpenMRS append behaviour doesn't matter when nothing is being removed)                                                                   
  - Reactions removed → DELETE+POST, identical to the cross-session path — DELETE removes the existing resource, POST creates a clean one with only the selected reactions
                                                                                                                                                                               
  Decision: why not always DELETE+POST?                                                                                                                                        
                                                                                                                                                                               
  PUT is preferred when no reactions are removed because it preserves the existing resource UUID, criticality, recordedDate, recorder attribution, and FHIR version history.   
  DELETE+POST creates a new resource with a new server-assigned UUID, losing those fields. Falling through to DELETE+POST only when necessary keeps the impact minimal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed allergy reaction management during clinical consultations. The system now correctly processes the removal of allergy reactions within the same encounter, ensuring patient allergy records accurately reflect current selections and completely remove unwanted reactions. This prevents data inconsistencies and improves clinical safety.

### DIFF
--- a/apps/clinical/src/services/consultationBundleService.ts
+++ b/apps/clinical/src/services/consultationBundleService.ts
@@ -124,6 +124,35 @@ export function createDiagnosisBundleEntries({
   return diagnosisEntries;
 }
 
+function createDeleteAndPostAllergyEntries(
+  allergy: AllergyInputEntry,
+  manifestationUUIDs: string[],
+  severity: 'mild' | 'moderate' | 'severe',
+  encounterSubject: Reference,
+  encounterReference: string,
+  practitionerUUID: string,
+): BundleEntry[] {
+  const deleteURL = `${ALLERGY_INTOLERANCE_RESOURCE_TYPE}/${allergy.resourceId}`;
+  const newResource = createEncounterAllergyResource(
+    allergy.id,
+    [allergy.type] as Array<'food' | 'medication' | 'environment' | 'biologic'>,
+    [{ manifestationUUIDs, severity }],
+    encounterSubject,
+    createEncounterReferenceFromString(encounterReference),
+    createPractitionerReference(practitionerUUID),
+    allergy.note,
+  );
+  return [
+    createBundleEntry(
+      deleteURL,
+      createDeleteAllergyResource(allergy.resourceId!),
+      'DELETE',
+      deleteURL,
+    ),
+    createBundleEntry(`urn:uuid:${crypto.randomUUID()}`, newResource, 'POST'),
+  ];
+}
+
 /**
  * Creates bundle entries for allergies as part of consultation bundle
  * @param params - Parameters required for creating allergy bundle entries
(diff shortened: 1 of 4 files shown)


apps/clinical/src/services/consultationBundleService.ts
```diff
@@ -202,17 +202,55 @@ export function createAllergiesBundleEntries({
...
+
+        if (!reactionsRemoved) {
+          const putResource = updateEncounterAllergyResource(
+            allergy.rawFhirResource,
+            manifestations,
+            severity,
+            createEncounterReferenceFromString(encounterReference),
+            allergy.note,
+          );
+          const putURL = `AllergyIntolerance/${allergy.resourceId}`;
```
sumaztwcode: Do not hardcode URL, use available constants, same below as well

apps/clinical/src/services/consultationBundleService.ts
```diff
@@ -202,17 +202,55 @@ export function createAllergiesBundleEntries({
...
+          allergyEntries.push(
+            createBundleEntry(putURL, putResource, 'PUT', putURL),
+          );
+        } else {
```
sumaztwcode, line 255: suggestion: This DELETE+POST block is structurally identical to the cross-session `else` branch below (~10 lines down). Both build a DELETE entry with `createDeleteAllergyResource` and a POST entry with `createEncounterAllergyResource` using the same arguments.

Consider extracting a helper (e.g. `createAllergyDeleteAndPostEntries(...)`) to avoid maintaining the same logic in two places.

apps/clinical/src/services/consultationBundleService.ts
```diff
@@ -202,17 +202,55 @@ export function createAllergiesBundleEntries({
...
+          const deleteURL = `AllergyIntolerance/${allergy.resourceId}`;
+          allergyEntries.push(
+            createBundleEntry(
+              deleteURL,
```
sumaztwcode: nit: `allergy.resourceId!` — the non-null assertion is safe here (guarded by `isExisting`), but TypeScript can't narrow through an intermediate boolean. Pre-existing in the cross-session path too, so not introduced by this PR.

sumaztwcode (commented):
## Review for BAH-4772

Well-scoped fix for OpenMRS FHIR2's append-only PUT behavior on allergy reactions. The detection logic is sound and the approach is conservative (preferring PUT when possible). A couple of things to discuss before merge.

Found 0 blocking issues, 2 suggestions, and 1 nit.

### Issues not in diff

| File | Type | Issue |
|------|------|-------|
| `consultationBundleService.test.ts` | suggestion | No tests cover the new same-session DELETE+POST path. At minimum: (1) reaction removed → expect DELETE+POST, (2) reaction only added → expect PUT, (3) all reactions replaced → expect DELETE+POST. The cross-session path has its own test block — the same-session reaction-removal path deserves equivalent coverage. |

sumaztwcode (approved):
Capture the trade-off for now, in the card

apps/clinical/src/services/consultationBundleService.ts
sahupadma: added

apps/clinical/src/services/consultationBundleService.ts
sahupadma: noted will take up in next card

apps/clinical/src/services/consultationBundleService.ts
sahupadma: noted will take up in next card

PR #485, opened 2026-06-11
BAH-4776 | Kanchi | Fix. Encounter date while adding or editing the consulatation on consultation pad.

Kanchi-Oswal-twk:
**Fix. Encounter Date not populated on consultation pad**

What Changed

- encounterDetailsStore.ts: Added isConsultationDateReady flag (defaults false). setConsultationDate now also sets this flag to true internally; reset clears it back to false.
- consultationPad/index.tsx: Added a useEffect to explicitly set consultationDate:
  - from activeEncounter.period.start for edit/session encounters — validated, so it falls back to the current date if the value is not a valid date;
  - new Date() for new consultations;
  - new Date() when an edit-encounter fetch resolves without a usable encounter (not found / error), so the field never stays stuck on the loading skeleton.

The edit-encounter fetch is now done with TanStack useQuery (the established pattern in apps/clinical) instead of a manual useEffect + useState + AbortController. isLoading is true from the first render (no manual initialiser needed), request cancellation uses the query's signal, and the error notification is driven by a small useEffect keyed on the query error.
- EncounterDetails.tsx: Changed DatePicker from defaultValue to value={consultationDate} for reactive rendering; wrapped in a skeleton that shows until isConsultationDateReady is true.

Why

The encounter date field was blank and showed a flash because of multiple bugs:
1. defaultValue only sets the input on first mount and never updates — changed to value for reactive rendering.
2. consultationDate in the store always started as new Date() (today), then got overwritten asynchronously when the encounter loaded — causing a visible flash from today's date to the encounter's actual date.
3. For edit consultations, setConsultationDate was never called with the encounter's period.start — added a useEffect to wire this up from activeEncounter.

The isConsultationDateReady flag ensures the DatePicker shows a skeleton until the correct date is confirmed, preventing any flash.

Review feedback addressed

- Edit-encounter fetch failure (Ravinder): when editEncounterUuid is set but getEncounterByUuid fails, activeEncounter is null and neither original branch fired, leaving the field stuck on the loading skeleton. Added a fallback branch (editEncounterUuid && !editEncounterLoading) that seeds the current date once the fetch settles.
- Use useQuery instead of a manual fetch (Ravinder): replaced the hand-rolled useEffect + useState + AbortController with useQuery, aligning with the existing pattern in apps/clinical. isLoading starts true on first render automatically, so the fallback branch does not fire prematurely — no manual loading initialiser required.
- Invalid period.start guard (CodeRabbit): a malformed period.start previously produced an Invalid Date (the dd/mm/yyyy placeholder symptom). It is now validated with isNaN(date.getTime()) and falls back to the current date.
- Test coverage (Ravinder): added isConsultationDateReady lifecycle coverage to the store tests, and added EncounterDetails tests for the date-picker skeleton toggling on isConsultationDateReady.

Tests

- setConsultationDate is called with period.start when available.
- Falls back to the current date when activeEncounter has no period.start.
- Falls back to the current date when period.start is an invalid date.
- Falls back to the current date when an edit-encounter fetch fails.
- Does not seed a date while the edit-encounter fetch is pending.
- Seeds the fetched encounter's period.start in edit mode.
- isConsultationDateReady: false initially, true after setConsultationDate, false after reset.
- EncounterDetails shows the date-picker skeleton when isConsultationDateReady is false and the DatePicker when true.
   
image

image

image

     

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved synchronization of the consultation date with the selected encounter, including correct fallbacks for missing/invalid dates and consistent behavior during edit-mode loading and errors.
  * The encounter date picker now waits for “consultation date readiness,” showing placeholders while loading and using a stable, controlled value for display.

* **Tests**
  * Expanded unit/integration coverage for consultation date seeding timing and for the new “consultation date readiness” store behavior.


apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -199,6 +199,16 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
 
   const activeEncounter = editEncounterUuid ? editEncounter : sessionEncounter;
 
+  useEffect(() => {
```
ravinderkabli, line 197: When editEncounterUuid is set but the getEncounterByUuid fetch fails, editEncounter remains null, so activeEncounter is null. In the useEffect logic neither branch fires.

Can you add a third condition to handle the error/fallback case:

editEncounterLoading already exists in the component (transitions to false in the finally block at line ~196) and can detect the "fetch done but no encounter" state.

Suggested Fix: Add a third condition to handle the error/fallback case:

useEffect(() => {
  if (activeEncounter?.period?.start) {
    useEncounterDetailsStore
      .getState()
      .setConsultationDate(new Date(activeEncounter.period.start));
  } else if (!editEncounterUuid) {
    useEncounterDetailsStore.getState().setConsultationDate(new Date());
  } else if (editEncounterUuid && !editEncounterLoading) {
    // Fetch resolved but encounter not found/errored — fall back to current date
    useEncounterDetailsStore.getState().setConsultationDate(new Date());
  }
}, [activeEncounter, editEncounterUuid, editEncounterLoading]);

Kanchi-Oswal-twk: @ravinderkabli this is resolved now

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -172,7 +172,9 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
     });
 
   const [editEncounter, setEditEncounter] = useState<Encounter | null>(null);
-  const [editEncounterLoading, setEditEncounterLoading] = useState(false);
+  const [editEncounterLoading, setEditEncounterLoading] = useState(
+    Boolean(editEncounterUuid),
```
ravinderkabli: `useQuery` is an established pattern used everywhere else in `apps/clinical`, so instead of `Boolean(editEncounterUuid)` which uses a manual `useEffect + useState`, I would recommend using `useUqery`

With TanStack Query, `isLoading` starts as `true` automatically from the first render when `enabled: !!editEncounterUuid` — no manual initialiser needed:

```ts
const { data: editEncounter, isLoading: editEncounterLoading } = useQuery({
  queryKey: ['encounter', editEncounterUuid],
  queryFn: ({ signal }) => getEncounterByUuid(editEncounterUuid!, { signal }),
  enabled: !!editEncounterUuid,
});
```

ravinderkabli (commented):
## Review for BAH-4776

Good progress — the previous MAJOR (stuck skeleton on edit-encounter fetch failure) is now fixed with the third branch, `editEncounterLoading` initialisation, and the `isNaN` guard. All acceptance criteria are met. A few minor test-coverage gaps to tidy up when convenient.

Found 0 blocking issues, 4 nits, and 1 question.

### Issues not in diff

| File:Line | Type | Issue |
|-----------|------|---------|
| apps/clinical/src/stores/__tests__/encounterDetailsStore.test.ts | nit | `isConsultationDateReady` lifecycle has zero coverage in the store test file (initial false, true after set, false after reset) |
| apps/clinical/src/components/forms/encounterDetails/__tests__/EncounterDetails.test.tsx | nit | No test exercises `isConsultationDateReady: false` → skeleton shown; default mock hardcodes `true` |

apps/clinical/src/components/consultationPad/index.tsx
Kanchi-Oswal-twk: @ravinderkabli this is resolved

ravinderkabli (dismissed):
Looks fine

ravinderkabli (approved):
Approved

PR #504, opened 2026-06-17
Adds useFormPatientContext to set patient data

sahupadma:
https://bahmni.atlassian.net/browse/BAH-4702                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                  
  What Changed                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                  
  Forms rendered via CarbonContainer previously only received { uuid } as the patient context. form.getPatient() inside form event scripts now returns full patient demographics so scripts can make clinical decisions based on patient data.    
                                                                                                                                                                                                                                                  
  - New useFormPatientContext hook fetches patient profile and composes it with active visit/encounter into an enriched PatientData object                                                                                                        
  - ObservationFormsContainer calls the hook and passes the result to both CarbonContainer (form renderer) and executeOnFormSaveEvent (form save scripts)                                                                                       
  - executeOnFormSaveEvent signature widened from patientUuid: string to patient: PatientData                                                                                                                                                     
                                                                                                                                                                                                                                                
  Fields exposed via form.getPatient()                                                                                                                                                                                                            
                                                                                                                                                                                                                                                
  uuid, identifier, display, givenName (firstName+middleName), familyName (lastName), age, ageInDays, birthdate, birthtime, gender, activeVisitUuid, currentEncounterUuid, getAgeDetails()   

## Summary by CodeRabbit

* **Refactor**
  * Enhanced observation form flow to use enriched patient context (now including encounter support), with improved loading/skeleton and failure handling, and “save/continue” disabled when patient context can’t be resolved.
  * Updated form save events to pass the full patient context instead of only a patient identifier.

* **New Features**
  * Added patient age computation helper (`computeAgeDetails`) and exposed it via the date package exports.

* **Tests**
  * Added/updated unit tests for patient context enrichment, correct propagation into save events, and the new `useFormPatientContext` hook behavior.


apps/clinical/src/hooks/useFormPatientContext.ts
```diff
@@ -0,0 +1,145 @@
...
+
+        const patientData: PatientData = {
+          uuid: profile.uuid,
+          identifier: preferredIdentifier?.identifier ?? undefined,
+          display: display,
+          givenName: givenName,
+          familyName: familyName,
+          age: person.age ?? undefined,
+          ageInDays: ageDetails?.ageInDays ?? undefined,
+          birthdate: person.birthdate ?? undefined,
```
Aishwarya-thoughtworks: **Issue**:
The hook passes `person.birthdate` and `person.birthtime` straight through. OpenMRS REST returns these as date-only / time-only strings (e.g. `birthdate: "1996-01-01"`, `birthtime: "08:00:00"`). The JIRA AC sample explicitly shows full ISO datetimes:

"birthdate" : "2026-04-12T18:30:00.000Z",
"birthtime" : "2026-04-13T00:30:00.000Z"

Form scripts that do `new Date(patient.birthtime)` or `patient.birthdate.getFullYear()` will misbehave on the string `"08:00:00"`. This is the primary user-facing contract of the PR.

**Suggested Fix**:
- For `birthdate`: convert to ISO datetime, e.g. `person.birthdate ? new Date(person.birthdate).toISOString() : undefined`.
- For `birthtime`: combine date + time into a single Date / ISO string, e.g. `new Date(\`${person.birthdate}T${person.birthtime}\`).toISOString()`.
- Update tests to use real ISO outputs.

apps/clinical/src/hooks/useFormPatientContext.ts
```diff
@@ -0,0 +1,145 @@
...
+  activeEncounterUuid: string | null | undefined;
+}
+
+interface UseFormPatientContextResult {
+  patient: PatientData | null;
+  isLoading: boolean;
+  error: Error | null;
+}
+
+export const useFormPatientContext = ({
```
Aishwarya-thoughtworks, line 30: apps/registration/src/hooks/usePatientDetails.ts` already calls `getPatientProfile` via `useQuery`. The project uses react-query for fetch-style hooks (see also `useObservationFormData`). This PR reinvents caching, dedupe, error normalization, cancellation, and loading state by hand. Side effects:
- No cross-view caching (every form open refetches).
- No request dedupe.
- No devtools visibility.
- Larger surface area to maintain.

**Suggested Fix**:
```ts
const { data: profile, isLoading, error } = useQuery({
  queryKey: ['patientProfile', patientUUID],
  queryFn: () => getPatientProfile(patientUUID!),
  enabled: !!patientUUID,
});
const patient = useMemo(
  () => profile ? buildPatientData(profile, activeVisitUuid, activeEncounterUuid) : null,
  [profile, activeVisitUuid, activeEncounterUuid],
);
```

apps/clinical/src/hooks/useFormPatientContext.ts
```diff
@@ -0,0 +1,145 @@
...
+        const givenName =
+          [firstName, middleName].filter(Boolean).join(' ') || undefined;
+        const familyName = lastName || undefined;
+        const display =
+          preferredName?.display ??
+          ([firstName, middleName, lastName].filter(Boolean).join(' ') ||
+            undefined);
+
+        // Compute age details from birthdate
+        let ageDetails: AgeDetails | null = null;
```
Aishwarya-thoughtworks: `getAgeDetails()` can return `null`, contradicting AC contract

**Issue**:
When `person.birthdate` is missing or `computeAgeDetails` throws, `ageDetails` stays `null`, and `getAgeDetails()` returns `null`. JIRA AC defines `getAgeDetails()` to return `{ year, month, day, ageInDays, ageText }` — scripts doing `patient.getAgeDetails().year` will crash on null.

**Suggested Fix**:
Return a safe default `{ year: 0, month: 0, day: 0, ageInDays: 0, ageText: '' }` when birthdate is missing, OR explicitly document and type the nullable contract so consumers know to guard.

apps/clinical/src/hooks/useFormPatientContext.ts
```diff
@@ -0,0 +1,145 @@
...
+
+interface AgeDetails {
+  year: number;
+  month: number;
+  day: number;
+  ageInDays: number;
+  ageText: string;
+}
+
+const computeAgeDetails = (birthdate: string): AgeDetails => {
```
Aishwarya-thoughtworks: computeAgeDetails` duplicates the existing `calculateAge` utility
**Issue**:
`packages/bahmni-services/src/date/date.ts` already exports `calculateAge` that returns `{ years, months, days }` using the same date-fns primitives (`parseISO`/`addYears`/`addMonths`/`differenceIn*`). The new code renames the keys (`years→year`, etc.), adds `ageInDays`, and concatenates `ageText`. The hook also imports `date-fns` directly instead of going through `@bahmni/services` date utilities. Matches the team pattern flagged in past reviews ("we are already leveraging date-fns" → reuse helpers).

**Suggested Fix**:
Extend `calculateAge` (or add `computeAgeDetails` next to it in `packages/bahmni-services/src/date/date.ts`) to also produce `ageInDays` and `ageText`, then export from `@bahmni/services`. Hook imports the shared util.

apps/clinical/src/hooks/__tests__/useFormPatientContext.test.ts
```diff
@@ -0,0 +1,509 @@
+import { getPatientProfile } from '@bahmni/services';
+import { renderHook, waitFor } from '@testing-library/react';
+import { useFormPatientContext } from '../useFormPatientContext';
+
+jest.mock('@bahmni/services', () => ({
+  getPatientProfile: jest.fn(),
+}));
+
+// Mock date-fns so age calculations are deterministic
+jest.mock('date-fns', () => ({
```
Aishwarya-thoughtworks: `date-fns` fully mocked → age computation logic is untested

**Issue**:
Tests mock every date-fns function with constants (`differenceInDays: 10950`, `differenceInYears: 30`, etc.). The assertion at line 151 (`{year: 30, month: 0, day: 10950, ageText: '30y 0m 10950d'}`) only proves the hook plumbs whatever date-fns returns — it does not prove `computeAgeDetails` produces correct values. Real-world bugs (timezone-driven off-by-one on `parseISO("1996-01-01")` + local-midnight `new Date()`, leap year, DST) are invisible.

**Suggested Fix**:
Either (a) extract `computeAgeDetails` into a pure util and unit-test it with real date-fns + `jest.useFakeTimers().setSystemTime(...)`, or (b) remove the date-fns mock and use fake timers in this test.

Aishwarya-thoughtworks (changes requested):

apps/clinical/src/hooks/useFormPatientContext.ts
sahupadma: done

apps/clinical/src/hooks/useFormPatientContext.ts
sahupadma: done

apps/clinical/src/hooks/useFormPatientContext.ts
sahupadma: done

apps/clinical/src/hooks/useFormPatientContext.ts
sahupadma: done

apps/clinical/src/hooks/__tests__/useFormPatientContext.test.ts
sahupadma: done

Aishwarya-thoughtworks (approved):

apps/clinical/src/hooks/useFormPatientContext.ts
ravinderkabli, line 1: This hook fetches /openmrs/ws/rest/v1/patientprofile/{uuid}?v=full on every form open. However, PatientDetails widget — which is always mounted on the same page — already calls:

useQuery({ queryKey: ['patient', patientUUID], queryFn: () => getFormattedPatientById(patientUUID!) })
That data is in the React Query cache before the user ever opens a form. useFormPatientContext duplicates that fetch against a different endpoint and throws away the caching benefit entirely.

Before this PR, ObservationFormsContainer passed patient={{ uuid: patientUUID }} to CarbonContainer and called executeOnFormSaveEvent(metadata, obs, patientUUID!, formData) — no loading state, no error state, no extra network call. 

That worked because PatientData.uuid is the only required field.

use the existing cache entry instead of a new hook

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
```diff
@@ -60,11 +62,21 @@ const ObservationFormsContainer: React.FC<ObservationFormsContainerProps> = ({
...
   const { t } = useTranslation();
   const patientUUID = usePatientUUID();
   const { user } = useActivePractitioner();
-  const { episodeOfCare } = useClinicalAppData();
+  const { episodeOfCare, activeVisitId } = useClinicalAppData();
+  const {
+    patient: patientContext,
+    isLoading: isPatientLoading,
+    error: patientError,
+  } = useFormPatientContext({
```
ravinderkabli, line 75: The patient data needed here is already in the React Query cache from PatientDetails — calling useQuery(['patient', patientUUID]) would return synchronously from cache with no UI impact.

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
```diff
@@ -168,6 +180,11 @@ const ObservationFormsContainer: React.FC<ObservationFormsContainerProps> = ({
   };
 
   const validateAndSave = () => {
+    if (!patientContext) {
```
ravinderkabli, line 183: This silently prevents saves while the patient profile is loading. With the cache-based approach this guard is unnecessary — patientContext would be derived from already-cached data and available immediately on render.

ravinderkabli:
One potential concern with switching to the FHIR cache is birthtime. However, OpenMRS FHIR2 exposes birthtime via the standard FHIR primitive extension pattern on _birthDate. Let's check if we can utilise that

apps/clinical/src/hooks/useFormPatientContext.ts
```diff
@@ -0,0 +1,125 @@
...
+  activeVisitUuid,
+  activeEncounterUuid,
+}: UseFormPatientContextOptions): UseFormPatientContextResult => {
+  const {
+    data: profileResponse,
+    isLoading,
+    error: queryError,
+  } = useQuery({
+    queryKey: ['patientProfile', patientUUID],
+    queryFn: () => getPatientProfile(patientUUID!),
```
ravinderkabli, line 41: ConsultationPage.tsx already calls useQuery({ queryKey: ['patient', patientUUID], ... }) at line 130, before ConsultationPad (and therefore    
  this form) renders. PatientDetails widget relies on that cache hit — it's an established pattern in this codebase. This hook bypasses that cache 
  entirely by calling a different endpoint (/patientprofile), making it the only component in the tree that ignores the pre-fetched data.

PR #511, opened 2026-06-19
BAH-4848 | Fix. Multiple Dividers in between the forms

ROHIT-V-TW:
## Summary by CodeRabbit

* **New Features**
  * Added an optional “action-triggered” flag for input controls.
  * Action-triggered controls are now supported so they can be conditionally available during editing.

* **Bug Fixes**
  * Improved consultation form control filtering to exclude action-triggered entries during standard consultation loading, while still showing the correct entry when in edit-only mode.

* **Tests**
  * Added coverage to verify action-triggered control inclusion/exclusion behavior (including the edit-only scenario).


apps/clinical/src/components/forms/models.ts
```diff
@@ -19,12 +20,8 @@ export interface InputControl {
   createBundleEntries?: (ctx: EncounterContext) => BundleEntry[];
   updateItemCDSCards?: (itemId: string, cards: CDSCard[]) => void;
   hasCriticalCDSCards?: () => boolean;
-  /**
-   * Optional direct submit handler. When present, the ConsultationPad calls this
-   * instead of including the form's data in the consultation bundle.
-   * Used for forms that call dedicated FHIR operations (e.g., $stop).
-   */
   onDirectSubmit?: () => Promise<void>;
+  isActive?: (ctx: EncounterSessionStartContext) => boolean;
```
ravinderkabli: suggestion: `isActive` appears to be dead/speculative code — no registered control sets it (grep across all `registerInputControl` calls). Only the new unit tests reference it; the actual divider fix is driven entirely by `onActionTriggered`. The new required `ctx` param threaded through `getActiveEntries` + the `useMemo` exists solely to feed this unused predicate.

Either remove `isActive` (and `ctx`) until a real consumer exists, or express the stopMedications gating via `isActive: ctx => !!ctx.stopMedication` and drop the separate `onActionTriggered` flag — pick one mechanism.

apps/clinical/src/components/consultationPad/utils.ts
```diff
@@ -40,14 +41,18 @@ export function loadEncounterInputControls(
...
+  ctx: EncounterSessionStartContext,
   editOnlyKey?: string,
 ): InputControl[] {
   return registry.filter((entry) => {
     const matchesEncounterType =
       !entry.encounterTypes || entry.encounterTypes.includes(encounterType);
     if (!matchesEncounterType) return false;
 
-    // When editOnly is set, show only the target form + encounterDetails.
+    if (entry.onActionTriggered && entry.key !== editOnlyKey) return false;
```
ravinderkabli, line 50: suggestion: This ships two overlapping conditional-visibility mechanisms — `onActionTriggered` (boolean, used) and `isActive` (predicate, unused) — on top of existing `editOnly`, `encounterTypes`, and `privilege` gating. The prior commit (`2ff47657`) used a single `shouldRender(ctx)` predicate; `5d013d4e` split it into two. `onActionTriggered` is fully expressible as `isActive`/`shouldRender`, so shipping both invites per-flag proliferation (each future action-only control adds another boolean).

Consolidate on one predicate and document it as the convention.

apps/clinical/src/components/consultationPad/utils.ts
```diff
@@ -40,14 +41,18 @@ export function loadEncounterInputControls(
...
+
+    if (entry.isActive && !entry.isActive(ctx)) return false;
```
ravinderkabli: suggestion: `isActive` does not respect `editOnlyKey`, unlike `onActionTriggered` one line above (which guards with `&& entry.key !== editOnlyKey`). This check runs unconditionally *before* the `editOnly` branch, so if a control ever defines `isActive` and becomes an edit target while `isActive(ctx)` returns false for that context, the control the user explicitly opened for editing is silently dropped — the edit panel would render only `encounterDetails`. It's masked today only because nothing uses `isActive`, but it's a correctness trap that contradicts the `editOnly` invariant (target form + encounterDetails always shown).

Fix: `if (entry.isActive && entry.key !== editOnlyKey && !entry.isActive(ctx)) return false;` — or move both gates inside the `editOnly` branch so edit mode always wins for the target.

apps/clinical/src/components/forms/models.ts
```diff
@@ -19,12 +20,8 @@ export interface InputControl {
   createBundleEntries?: (ctx: EncounterContext) => BundleEntry[];
   updateItemCDSCards?: (itemId: string, cards: CDSCard[]) => void;
   hasCriticalCDSCards?: () => boolean;
-  /**
-   * Optional direct submit handler. When present, the ConsultationPad calls this
-   * instead of including the form's data in the consultation bundle.
-   * Used for forms that call dedicated FHIR operations (e.g., $stop).
-   */
   onDirectSubmit?: () => Promise<void>;
```
ravinderkabli, line 28: nit: The JSDoc that documented `onDirectSubmit` (it bypasses the consultation bundle and calls a dedicated FHIR operation such as `$stop`) was removed in this PR. That behavior is non-obvious from the signature and is still live (`stopMedications/index.ts` implements it). Looks like an accidental deletion — please restore the comment.

apps/clinical/src/components/forms/observations/index.ts
```diff
@@ -1,10 +1,12 @@
+import { CONSULTATION_PAD_PRIVILEGES } from '@bahmni/widgets';
 import { createObservationBundleEntries } from '../../../services/encounterBundleService';
 import { useObservationFormsStore } from '../../../stores/observationFormsStore';
 import { registerInputControl } from '../registry';
 import ObservationFormsPanel from './ObservationFormsPanel';
 
 registerInputControl({
   key: 'observationForms',
+  privilege: CONSULTATION_PAD_PRIVILEGES.OBSERVATIONS,
```
ravinderkabli: nit: `observationForms` becomes the only control declaring `privilege` in its `registerInputControl(...)` call — every other control gets privilege from JSON config (`inputControlConfig.privileges`) and/or checks it inside its component. Combined with the `loadEncounterInputControls` fallback change, this is the only control whose registry privilege is actually enforced, and it's now gated in two places (renderer + inner `ObservationForms`). Acceptable as the fix, but please note the divergence in the PR description, or follow up by migrating all controls to registry-level privilege for uniformity.

apps/clinical/src/components/consultationPad/utils.ts
```diff
@@ -31,7 +32,7 @@ export function loadEncounterInputControls(
               : inputControlConfig.encounterTypes,
           privilege: inputControlConfig.privileges?.length
             ? inputControlConfig.privileges
-            : undefined,
+            : entry.privilege,
```
ravinderkabli: question: This fallback change (`undefined` → `entry.privilege`) is what actually activates the observations privilege gate — a cross-cutting, security-relevant behavior change (config privileges take precedence; registry privilege becomes the default) bundled into a divider-visibility bugfix. Correct and safer, but worth calling out explicitly in the PR description since it affects all controls, not just the divider.

ravinderkabli (commented):
## Review for BAH-4848

ravinderkabli (approved):
Looks fine

PR #518, opened 2026-06-24
 BAH-4702 | Refactor. Use FHIR cache for form patient context instead…

sahupadma:
JIRA: https://bahmni.atlassian.net/browse/BAH-4702                                                                                                                                                         
                                                                                                                                                                               
  Summary                                                                                                                                                                      
                                                                                                                                                                               
  Refactors the form patient context (introduced in #504) to source data from the FHIR cache that ConsultationPage already populates, instead of making a separate             
  getPatientProfile call on every form open.
                                                                                                                                                                               
  - Deletes useFormPatientContext hook and its 595-line test file                                                                                                              
  - Adds givenName, familyName, birthtime, and identifier to the shared FormattedPatientData model in @bahmni/services
  - Replaces the hook call in ObservationFormsContainer with useQuery(['patient', patientUUID]) — hitting the existing cache, not firing a new request                         
  - Adds a useMemo to map FormattedPatientData → form patient context inline                                                                                                   
                                                                                                                                                                               
  Why                                                                                                                                                                          
                                                                                                                                                                               
  useFormPatientContext called GET /openmrs/ws/rest/v1/patient/{uuid}?v=full on every form open. ConsultationPage already fetches the FHIR patient into cache before any form  
  renders — the same cache hit PatientDetails widget uses. This PR eliminates the duplicate request by reusing that cache.
                                                                                                                                                                               
  Fixes included                                                                                                                                                               
   
  - identifier now selects by use='official' before falling back to index 0                                                                                                    
  - birthdate passed as a date-only string — avoids UTC-midnight timezone shift from new Date(dateOnly).toISOString()                                                        
  - gender converted from FHIR ('male') to OpenMRS codes ('M') for form script compatibility                                                                                   
  - AGE_DETAILS_DEFAULT moved to module scope (was recreated on every render)                                                                                                  
  - !patientContext guard removed from validateAndSave — context is derived from cached data synchronously                                                                     
                                                                                                                                                                               

## Summary by CodeRabbit

* **New Features**
  * Patient details now display enriched formatted fields, including preferred identifier selection, mapped gender, birth time extraction, and clearer given/family names.
  * Observation forms now load patient context directly and consistently while preparing the same fields for save flows.

* **Bug Fixes**
  * Improved age/birth-time fallback behavior when patient birth data is incomplete.
  * Observation form error messaging is now shown more reliably with localized “Error fetching patient data”.

* **Tests**
  * Updated and expanded automated tests to reflect the new patient data flow and richer formatted output.


apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
```diff
@@ -68,15 +86,36 @@ const ObservationFormsContainer: React.FC<ObservationFormsContainerProps> = ({
...
-  const {
-    patient: patientContext,
-    isLoading: isPatientLoading,
-    error: patientError,
-  } = useFormPatientContext({
-    patientUUID,
-    activeVisitUuid: activeVisitId,
-    activeEncounterUuid,
+
+  const { data: fhirPatient } = useQuery({
```
ravinderkabli: blocking: Loading and error state dropped from `useQuery` — `isLoading` and `error` are not destructured, `isPrimaryButtonDisabled` is hardcoded to `false` (line 470), and the loading skeleton no longer accounts for patient data loading.

If the FHIR cache is cold (direct navigation, first load), `fhirPatient` is `undefined` → `patientContext` is `null` → the form area shows `OBSERVATION_FORM_LOADING_METADATA_ERROR` (misleading — metadata loaded fine), and the Save button is enabled but silently does nothing when clicked.

Suggested fix — destructure `isLoading` and `error`:
```ts
const { data: fhirPatient, isLoading: isPatientLoading, error: patientError } = useQuery({...});
```
Then restore `isLoadingMetadata || isPatientLoading` in the skeleton condition, `error || patientError` in the error branch, and `isPrimaryButtonDisabled={isPatientLoading || !patientContext}`.

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
```diff
@@ -180,11 +219,6 @@ const ObservationFormsContainer: React.FC<ObservationFormsContainerProps> = ({
   };
 
   const validateAndSave = () => {
-    if (!patientContext) {
-      setValidationErrorType(VALIDATION_STATE_SCRIPT_ERROR);
-      setValidationErrorMessage(t('OBSERVATION_FORM_LOADING_METADATA_ERROR'));
-      return;
-    }
     if (formContainerRef.current) {
```
ravinderkabli, line 226: blocking: The `!patientContext` null guard was removed, but `patientContext` is typed as `{...} | null` (from `useMemo`). It's passed directly to `executeOnFormSaveEvent` which expects non-nullable `PatientData` — this is a TS2345 error under `strict: true`.

While the render guard at line 415 prevents the form from mounting when `patientContext` is null (so `formContainerRef.current` won't be set), TypeScript can't infer this across the callback boundary.

Suggested fix — restore the guard:
```ts
if (!patientContext) {
  setValidationErrorType(VALIDATION_STATE_SCRIPT_ERROR);
  setValidationErrorMessage(t('OBSERVATION_FORM_LOADING_METADATA_ERROR'));
  return;
}
```

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
```diff
@@ -31,13 +35,27 @@ import {
...
 
+const AGE_DETAILS_DEFAULT: AgeDetails = {
+  year: 0,
+  month: 0,
+  day: 0,
+  ageInDays: 0,
+  ageText: '',
+};
+
+const mapGenderFromFhir = (fhirGender: string): string => {
```
ravinderkabli: suggestion: `mapGenderFromFhir` is already exported from `apps/registration/src/utils/fhirUtils.ts` with identical logic. Two copies of the same mapping will drift over time.

Consider moving it to `@bahmni/services` (e.g., alongside `formatPatientData`) so both apps import from one place.

packages/bahmni-services/src/patientService/__tests__/patientService.test.ts
```diff
@@ -832,6 +847,63 @@ describe('Patient Service', () => {
...
+    it('should return null age fields when birthDate is absent', () => {
+      const patient: Patient = {
+        resourceType: 'Patient',
+        id: 'test-uuid',
+      };
+
+      const result = formatPatientData(patient);
+
+      expect(result.birthDate).toBeNull();
+      expect(result.birthtime).toBeNull();
```
ravinderkabli, line 905: suggestion: All `birthtime` assertions in the new tests only cover the `null` case. The epoch-date-replacement regex and the FHIR `_birthDate` extension traversal are never exercised.

Add a test with `_birthDate: { extension: [{ url: BIRTH_TIME_EXT_URL, valueDateTime: '1970-01-01T08:30:00+05:30' }] }` and `birthDate: '1990-01-01'`, asserting `birthtime === '1990-01-01T08:30:00+05:30'`.

packages/bahmni-services/src/patientService/patientService.ts
```diff
@@ -153,26 +154,47 @@ export const formatPatientData = (patient: Patient): FormattedPatientData => {
...
+  const birthDateExt = (patient as unknown as Record<string, unknown>)
+    ._birthDate as
+    | { extension?: { url: string; valueDateTime?: string }[] }
+    | undefined;
+  const birthtimeRaw =
+    birthDateExt?.extension?.find((e) => e.url === BIRTH_TIME_EXT_URL)
+      ?.valueDateTime ?? null;
+  // OpenMRS FHIR2 sets the date portion of valueDateTime to epoch (1970-01-01)
+  // as a placeholder. Replace it with the actual birthDate.
+  const birthtime =
```
ravinderkabli, line 188: suggestion: The old hook returned `birthtime` as a UTC ISO string (via `new Date(...).toISOString()`). The new code passes the FHIR `valueDateTime` directly, which includes a timezone offset (e.g., `'1990-01-01T08:30:00+05:30'`). Form scripts parsing `birthtime` may silently receive a different format.

Either normalize to UTC ISO (`new Date(birthtime).toISOString()`) for backward compatibility, or document this as a deliberate format change.

packages/bahmni-services/src/constants/fhir.ts
```diff
@@ -1,3 +1,6 @@
+export const BIRTH_TIME_EXT_URL =
```
ravinderkabli, line 1: nit: This constant is also defined in `apps/registration/src/utils/fhirUtils.ts`. `@bahmni/services` is the right canonical home — consider updating the registration app to import from here (could be a follow-up PR).

ravinderkabli (commented):
## Review for BAH-4702

Good approach — reusing the FHIR cache eliminates a redundant API call and the model enrichment is well thought out. A few things need addressing before merge.

packages/bahmni-services/src/constants/fhir.ts
sahupadma: done

packages/bahmni-services/src/patientService/patientService.ts
sahupadma: Kept the timezone-offset format as a deliberate format change. the new test documents this behavior explicitly

packages/bahmni-services/src/patientService/__tests__/patientService.test.ts
sahupadma: added

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
sahupadma: done

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
sahupadma: restored

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
sahupadma: done

ravinderkabli (approved):
Looks fine

PR #520, opened 2026-06-24
BAH-4773 | Fix patient relationship section bugs

himanshiaaradhana:
## Summary

Fixes two bugs in the patient relationship section of the registration form.

## Changes

- **Fix "Valid Till" datepicker** — re-selecting today's date after switching to other dates was silently rejected. `minDate={new Date()}` carries the current time, so today at midnight was treated as earlier than `minDate` by flatpickr. Fixed by normalising `minDate` to the start of today (`setHours(0,0,0,0)`).

- **Prevent duplicate/self patient in relationship dropdown** — the patient-search ComboBox had no guard against selecting a patient already chosen in another row, or selecting the patient being edited as their own relative. Fixed by filtering each row's suggestions to exclude already-selected patient UUIDs and (in edit mode) the current patient's UUID.

## Files Changed

- `apps/registration/src/components/forms/patientRelationships/RelationshipRow.tsx`
- `apps/registration/src/components/forms/patientRelationships/PatientRelationships.tsx`

## Fixes

BAH-4773

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved patient relationship entry so date selection is stored and restored consistently.
  * Fixed “till date” picking to use a standard date format and prevent off-by-one-day issues.
  * Updated relationship suggestions to better avoid duplicate patient selections across rows, including excluding the current patient where needed.

* **Tests**
  * Added regression coverage for date handling and patient suggestion filtering.


apps/registration/src/components/forms/patientRelationships/PatientRelationships.tsx
```diff
@@ -88,10 +93,28 @@ export const PatientRelationships = ({
...
-      const suggestions = getPatientSuggestions(rel.id);
+      const excludedUuids = new Set(selectedPatientUuids);
+      excludedUuids.delete(rel.patientUuid);
+      if (currentPatientUuid) {
+        excludedUuids.add(currentPatientUuid);
+      }
+
+      const suggestions = getPatientSuggestions(rel.id).filter(
+        (suggestion) => !excludedUuids.has(suggestion.id),
+      );
```
ravinderkabli: blocking: Cross-row exclusion logic is in the component render body rather than the state-owning `usePatientRelationship` hook.

`usePatientRelationship` already centralizes relationship state and exposes `getPatientSuggestions`. With this change, `getPatientSuggestions(rel.id)` returns *unfiltered* results and the caller must remember to re-filter them — a leaky abstraction, and it fragments relationship domain logic across two layers (the new test has to mock both `RelationshipRow` and the whole hook to reach it).

Consider moving the exclusion into the hook — e.g. `getPatientSuggestions(rel.id)` internally excludes other non-deleted rows' `patientUuid`s and the current patient, with `currentPatientUuid` passed into the hook once. Keeps the component presentational and consistent with how all other relationship state is sourced.

apps/registration/src/components/forms/patientRelationships/RelationshipRow.tsx
```diff
@@ -57,6 +57,9 @@ export const RelationshipRow = ({
 }: RelationshipRowProps) => {
   const isExisting = relationship.isExisting === true;
 
+  const minSelectableDate = new Date();
+  minSelectableDate.setHours(0, 0, 0, 0);
```
ravinderkabli: suggestion: Use `startOfDay` from date-fns instead of mutating `new Date()` + `setHours`.

`date-fns` (^4.1.0) is a root dependency and the established convention across the registration app (`fhirPatientToFormData.ts`, `ageUtils.ts`, `patientDataConverter.ts`).

```ts
import { startOfDay } from 'date-fns';
const minSelectableDate = startOfDay(new Date());
```

ravinderkabli (commented):
## Review for BAH-4773

himanshiaaradhana:
All review comments have been addressed in the latest commits:

- **[Blocking] Cross-row exclusion logic in the component render body** (`PatientRelationships.tsx`) — Moved the exclusion logic into the state-owning `usePatientRelationship` hook. `getPatientSuggestions(rowId)` now returns already-filtered results (excluding other non-deleted rows' patients + the current patient), so the component stays presentational and relationship domain logic is no longer split across layers.
- **Use counts instead of a plain `Set` for per-row exclusion** — The hook now tracks UUID counts in a `Map` and only re-allows the current row's UUID when its count is exactly 1, so a patient duplicated across rows stays correctly excluded everywhere.
- **Use `startOfDay` from date-fns instead of mutating `new Date()` + `setHours`** (`RelationshipRow.tsx`) — Done, using the existing `date-fns` dependency per the app convention.
- **Midnight race in the regression test** (`RelationshipRow.test.tsx`) — The reference `today` is now captured before deriving `minDate`.

Filtering test cases were migrated into `usePatientRelationship.test.tsx` (including a new duplicate-UUID case), and the component-level `PatientRelationships.suggestionFiltering.test.tsx` was removed since it mocked both the hook and `RelationshipRow`. Lint is clean and all 721 registration tests pass.

@ravinderkabli can you please review this?

ravinderkabli (approved):
Looks fine

PR #524, opened 2026-06-26
BAH-4826 Appointment Search: Introduce appointmentservice type rule and render Action buttons only when all the enabled rules are satisifed

kousalyapt-tw:
Action buttons can now be conditionally rendered based on a combination of privilege, appointment status, date, and appointment service rules — all declared in the registration config JSON.

## Summary by CodeRabbit

* **New Features**
  * Added a “check in and start visit” appointment action, including a configurable submit endpoint.
  * Expanded action enablement rules with a new appointment-service rule type (supports allow/deny lists) and app date checks.
  * Appointment search results now include the appointment service UUID.
* **Bug Fixes**
  * Action button rendering now evaluates against the current appointment row to prevent incorrect enabled/disabled states.
  * Actions that don’t meet enablement rules are hidden rather than displayed disabled.
* **Tests**
  * Expanded test coverage for the new action flow and enhanced rule evaluation logic.


apps/registration/src/pages/patientSearchPage/appointmentSearchResultActionHandler.ts
```diff
@@ -89,42 +90,55 @@ export const handleActionButtonClick = async (
...
     privilegeCheck: privilegeValidator(userPrivileges),
     statusCheck: statusValidator,
     appDateCheck: appDateValidator,
   };
 
-  return enabledRules.every((rule) =>
-    ruleValidatorMap[rule.type](rule.values, row),
-  );
+  return action.enabledRule.every((rule) => {
+    if (rule.type === 'appointmentService') {
```
mohan-13: Please do not add conditional check like this, instead enhance the existing validator to handle cases commonly. You can have that as an optional parameter

packages/bahmni-services/src/appointmentService/appointmentService.ts
```diff
@@ -82,6 +83,19 @@ export const updateAppointmentStatus = async (
   });
 };
 
+/**
```
mohan-13: We should avoid adding JSDoc, the variables and functions are self explanatory

mohan-13 (dismissed):

mohan-13 (approved):

PR #527, opened 2026-06-26
BAH-4821 | Add. Config-Driven Clinical List With Extension And Search Modules

rahu1ramesh:
JIRA → [BAH-4821](https://bahmni.atlassian.net/browse/BAH-4821)

## Description

Introduces a config-driven clinical queue page powered by a new Extensions system in bahmni-widgets. Extensions are read from a remote JSON config, validated via AJV schemas, privilege-filtered, and rendered as tabs. Each tab fetches and validates its own per-widget config. A Search module allows search widgets to register themselves and receive their config automatically via a HOC. The clinical app's list route is wired to consume this system.

## Key Features

- Config-driven tab rendering on the clinical queue page — tabs, icons, and privilege gating all driven by extension.json
- Two-level AJV schema validation: parent extension config + per-tab widget config
- Search widget auto-wrapping: registerSearchWidget automatically applies SearchWidgetConfigProvider via the withSearchConfig HOC
- AdvancedSearchWidget registered as the first search widget (placeholder, tab config delivered via context)

[BAH-4821]: https://bahmni.atlassian.net/browse/BAH-4821?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/clinical/src/pages/list/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const CLINICAL_SEARCH_CONFIG_URL =
```
mohan-13: This is not only for search extensions. The extension points can be for different features.

packages/bahmni-widgets/src/extensions/Extensions.tsx
```diff
@@ -0,0 +1,14 @@
+import TabView from './components/tabView';
+import ExtensionConfigProvider from './provider';
+
+interface ExtensionsProps {
+  configUrl: string;
+}
+
+const Extensions = ({ configUrl }: ExtensionsProps) => (
+  <ExtensionConfigProvider configUrl={configUrl}>
+    <TabView />
```
mohan-13: @rahu1ramesh @SrijanRamSN We cant assume that all extensions will be only for search. Here is where we need to use extensionPointId as well to filter out

apps/clinical/src/pages/list/constants.ts
mohan-13: Also, with the new app, rather than a seperate file, we have had the extensions within the same app.json config file. Please look into card details

apps/clinical/src/pages/list/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <BaseLayout
+        header={<Header breadcrumbItems={breadcrumbItems} />}
+        main={
+          <div
+            id="clinical-list-tabs"
+            data-testid="clinical-list-tabs-test-id"
+            aria-label="Clinical List Tabs"
+            className={styles.tabs}
+          >
+            <Extensions configUrl={CLINICAL_SEARCH_CONFIG_URL} />
```
mohan-13: I would recommend not assuming the entire extensions to be for search

mohan-13 (changes requested):

packages/bahmni-widgets/src/extensions/components/tabView/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+          return (
+            <TabPanel
+              key={ext.id}
+              id={`${ext.id}-panel`}
+              data-testid={`${ext.id}-panel-test-id`}
+              aria-label={`Panel ${ext.id}`}
+            >
+              {Widget ? (
+                <Suspense fallback={null}>
+                  <Widget configUrl={ext.extensionParams.configUrl} />
```
mohan-13: Dont scope it to only configURL, it is upto the widget to decide what all params will be needed.

apps/clinical/public/locales/locale_en.json
```diff
@@ -437,6 +442,7 @@
   "PARTICIPANT": "Participant(s)",
   "PATIENT_HEADER_LABEL": "Patient Header",
   "PATIENT_HEADER_SECTION": "Patient Information Section",
+  "PENDING_ORDERS_SEARCH_LABEL": "Pending Orders",
```
mohan-13: Why is this key needed here ? Seems an implementation specific key

apps/clinical/public/locales/locale_es.json
```diff
@@ -437,6 +442,7 @@
   "PARTICIPANT": "Partícipe(es)",
   "PATIENT_HEADER_LABEL": "Cabecera del Paciente",
   "PATIENT_HEADER_SECTION": "Sección de Información del Paciente",
+  "PENDING_ORDERS_SEARCH_LABEL": "Órdenes Pendientes",
```
mohan-13: Same here Seems an implementation specific key

apps/clinical/src/pages/list/extensions/searchHandler/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+                : undefined
+            }
+          >
+            {t(ext.translationKey)}
+          </Tab>
+        ))}
+      </TabList>
+      <TabPanels>
+        {extensions.map((ext) => {
+          const widget = getSearchWidget(ext.type);
```
mohan-13: rather than by ext.type shouldn't this be by the extensionParam.searchHandler

apps/clinical/src/pages/list/utils.ts
```diff
@@ -0,0 +1,13 @@
+import { type Extension } from '../../extensions';
+
+export const groupExtensionsByPoint = (
```
mohan-13: shouldnt this move into the root extensions utils you alrady have ? what is specific to list here ?

apps/clinical/src/pages/list/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+      header={<Header breadcrumbItems={breadcrumbItems} />}
+      main={
+        <div
+          id="clinical-list-page"
+          data-testid="clinical-list-page-test-id"
+          aria-label="Clinical List Page"
+          className={styles.page}
+        >
+          {visibleHandlers.length > 0 ? (
+            visibleHandlers.map(({ pointId, Handler, filtered }) => (
```
mohan-13, line 54: IMHO, this still feels like an overkill here, why would multiple extension points come at the same layout here ?

apps/clinical/src/providers/clinicalConfig/schema.json
```diff
@@ -211,6 +211,34 @@
...
+          "icon": { "type": "string" },
+          "requiredPrivileges": {
+            "type": "array",
+            "items": { "type": "string" }
+          },
+          "extensionParams": {
+            "type": "object",
+            "additionalProperties": true,
+            "properties": {
+              "searchHandler": { "type": "string" },
```
mohan-13, line 246: We should mkae this enum and keep adding supported handlers as we evolve. This can make the schema validator too broad just accepting string

apps/clinical/src/routes/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+export const routes: Routes = [
+  {
+    path: ':patientUuid',
+    component: ConsultationPage,
+    name: 'Consultation',
+  },
+  {
+    path: '/',
+    component: ClinicalList,
+    name: 'ClinicalList',
```
mohan-13: What is this name field used for ?

packages/bahmni-widgets/src/search/provider.tsx
```diff
@@ -0,0 +1,60 @@
...
+  SearchWidgetProps,
+} from './models';
+import searchWidgetConfigSchema from './schema.json';
+
+interface SearchWidgetConfigProviderProps {
+  configUrl: string;
+  children: ReactNode;
+}
+
+const SearchWidgetConfigProvider = ({
```
mohan-13: Is this context and provider really needed ? reading from configUrl will be a specific case for commonSearch, other search handlers will be simple and extensionParams will just suffice them

packages/bahmni-widgets/src/search/provider.tsx
```diff
@@ -0,0 +1,60 @@
...
+  configUrl,
+  children,
+}: SearchWidgetConfigProviderProps) => {
+  const Provider = useMemo(
+    () =>
+      createConfigProvider<SearchWidgetConfig, SearchWidgetConfigContextType>({
+        context: SearchWidgetConfigContext,
+        queryKey: ['searchWidgetConfig', configUrl],
+        queryFn: () =>
+          getConfig<SearchWidgetConfig>(configUrl, searchWidgetConfigSchema),
```
mohan-13: also remeber, why would all search widgets even if they define their own config url have the same schema

packages/bahmni-widgets/src/search/provider.tsx
mohan-13: This is an overkill for a generalised config provider for all components, seems to be specific for commonSearch widget only.

apps/clinical/src/providers/clinicalConfig/schema.json
rahu1ramesh: Later on if an implementation wants to add their own search handler, keeping it an enum here would mean they will need to first edit this. Unless we give them an option to update the schema, a more generic type is better.

apps/clinical/src/pages/list/index.tsx
rahu1ramesh: Not necessarily the same layout, since now we have only on defined, this can be improved later as we add more handlers.

apps/clinical/src/routes/index.tsx
rahu1ramesh: removed

packages/bahmni-widgets/src/search/provider.tsx
rahu1ramesh: Let's discuss this, for a search widget i feel the contract should be the same, the config coming in from the configURL, irrespective of how simple / short it is.

packages/bahmni-widgets/src/search/provider.tsx
rahu1ramesh: I feel this should be the same schema across, but we can discuss once on this.

packages/bahmni-widgets/src/search/provider.tsx
mohan-13: Requiring a config file for all extensions will be an overkill, for example an sqlSearchHandler can be simply configured with just two or three params, why demand a config file for such case

apps/clinical/src/providers/clinicalConfig/schema.json
```diff
@@ -223,6 +223,33 @@
         }
       }
     },
+    "extensions": {
+      "type": "array",
+      "description": "List of app-specific extension configurations",
+      "items": {
+        "type": "object",
+        "required": ["id", "extensionPointId", "translationKey"],
+        "additionalProperties": true,
```
mohan-13, line 232: additionalProperties does not need to be true

mohan-13 (approved):

apps/clinical/src/providers/clinicalConfig/schema.json
rahu1ramesh: This is until we finalize the flow, will set it to false once a few extensions are built

PR #529, opened 2026-06-29
BAH-4704 | Fix. Observation Forms to check fields with conditional mandatory fields onHidden

ROHIT-V-TW:
## JIRA
[BAH-4704](https://bahmni.atlassian.net/browse/BAH-4704)

## Description

Observation forms were silently saving when a mandatory field was hidden via `isHidden` scripting and then became visible again — `form2-controls` does not propagate mandatory errors to `getValue().errors` for fields that were never interacted with after becoming visible. The same gap affected always-visible mandatory fields that the user never touched.

## Key Changes

- **`hasMissingMandatoryVisibleField`** — new recursive helper in `bahmni-services` that walks the Container's Immutable state data (via `toJS()`) to detect visible, non-voided mandatory leaf fields with no value. Handles text/number/date (`value.value`), coded/select (`value.concept`), and multi-select (array concept).
- **`ObservationFormsContainer.validateAndSave`** — integrates the helper alongside `getValue().errors`. The `isEmpty` guard now runs _after_ `hasMissingMandatory` is computed so a form with only untouched mandatory fields shows "mandatory field missing" rather than "empty form".
- **Hidden subtree fix** — the recursive traversal now skips the entire subtree when a group node is `hidden` or `voided`, preventing false positives on sections that form2-controls has hidden.

## Acceptance Criteria

| # | Criteria | Status |
|---|---|---|
| AC1 | Mandatory field that was hidden and revealed via `isHidden` scripting blocks save | ✅ |
| AC2 | Always-visible mandatory field never touched blocks save with correct error message | ✅ |
| AC3 | Filled mandatory fields (text, coded, multi-select) do not false-positive | ✅ |
| AC4 | Hidden/voided fields and hidden group subtrees are not flagged | ✅ |

## Testing

- Observation form with a mandatory field hidden by default via `isHidden` — reveal it without filling, click Save → blocked with mandatory error
- Observation form with an always-visible mandatory field never touched → blocked with mandatory error  
- Fill all mandatory fields → saves successfully
- Coded (select) mandatory field — answered → saves, unanswered → blocked
- Group/section with `isHidden` hiding it → mandatory child inside is not flagged

[BAH-4704]: https://bahmni.atlassian.net/browse/BAH-4704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
+++ b/apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
@@ -272,15 +272,6 @@ const ObservationFormsContainer: React.FC<ObservationFormsContainerProps> = ({
       const isEmpty = !hasAnyValue; // Empty if no values (including empty strings), even if there are notes
       const hasErrors = errors && errors.length > 0;
 
-      if (isEmpty) {
-        setValidationErrorType(VALIDATION_STATE_EMPTY);
-        return;
-      }
-
-      // form2-controls does not propagate mandatory errors to getValue().errors for
-      // fields that were never interacted with — including always-visible mandatory
-      // fields and fields revealed via isHidden scripting. Check the Container data
-      // directly to catch these violations.
       const containerStateData = (
         formContainerRef.current as {
           state?: { data?: Record<string, unknown> | { toJS?: () => unknown } };
@@ -291,6 +282,11 @@ const ObservationFormsContainer: React.FC<ObservationFormsContainerProps> = ({
           | Record<string, unknown>
           | undefined,
       );
+
+      if (isEmpty && !hasMissingMandatory) {
+        setValidationErrorType(VALIDATION_STATE_EMPTY);
+        return;
+      }
 
       if (hasErrors || hasMissingMandatory) {
         const hasMandatoryError =
--- a/apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
+++ b/apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
@@ -23,6 +23,7 @@ import {
   extractNotesFromFormData,
   type AgeDetails,
   computeAgeDetails,
+  hasMissingMandatoryVisibleField,
 } from '@bahmni/services';
 import { useActivePractitioner, usePatientUUID } from '@bahmni/widgets';
 import { useQuery } from '@tanstack/react-query';
@@ -276,14 +277,31 @@ const ObservationFormsContainer: React.FC<ObservationFormsContainerProps> = ({
         return;
       }
 
-      if (hasErrors) {
-        const hasMandatoryError = errors
-          .flat()
-          .some(
-            (err: { get?: (key: string) => string; message?: string }) =>
-              (err.get?.('message') ?? err.message) ===
-              VALIDATION_STATE_MANDATORY,
-          );
+      // form2-controls does not propagate mandatory errors to getValue().errors for
+      // fields that were never interacted with — including always-visible mandatory
+      // fields and fields revealed via isHidden scripting. Check the Container data
+      // directly to catch these violations.
+      const containerStateData = (
+        formContainerRef.current as {
+          state?: { data?: Record<string, unknown> | { toJS?: () => unknown } };
+        } | null
+      )?.state?.data;
+      const hasMissingMandatory = hasMissingMandatoryVisibleField(
+        convertImmutableToPlainObject(containerStateData) as
+          | Record<string, unknown>
+          | undefined,
+      );
+
+      if (hasErrors || hasMissingMandatory) {
+        const hasMandatoryError =
+          hasMissingMandatory ||
+          errors
+            .flat()
+            .some(
+              (err: { get?: (key: string) => string; message?: string }) =>
+                (err.get?.('message') ?? err.message) ===
+                VALIDATION_STATE_MANDATORY,
+            );
         const errorType = hasMandatoryError
           ? VALIDATION_STATE_MANDATORY
           : VALIDATION_STATE_INVALID;
--- a/packages/bahmni-services/src/observationFormsService/observationFormsTransformer.ts
+++ b/packages/bahmni-services/src/observationFormsService/observationFormsTransformer.ts
@@ -524,7 +524,45 @@ function processControlForNotesExtraction(
 }
 
 /**
- * Checks if any visible, non-voided mandatory fields in the Container data have no values.
+ * Returns true when a leaf control's value is considered empty.
+ *
+ * form2-controls serializes values differently per control type:
+ * - Free-text / numeric / date: { value: "abc" | 123 | null }
+ * - Coded / select:             { value: null, concept: { uuid, name } }
+ * - Multi-select:               { value: null, concept: [{ uuid, name }] }
+ *
+ * A field is empty only when both `value.value` AND `value.concept` are absent.
+ */
+function isLeafValueEmpty(data: Record<string, unknown>): boolean {
+  const valueObj =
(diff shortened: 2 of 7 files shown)


packages/bahmni-services/src/observationFormsService/observationFormsTransformer.ts
```diff
@@ -522,3 +522,51 @@ function processControlForNotesExtraction(
...
+
+  const control = data.control as Record<string, unknown> | undefined;
+  const properties = control?.properties as Record<string, unknown> | undefined;
+
+  if (properties?.mandatory === true && !data.hidden && !data.voided) {
+    const valueObj =
+      data.value && typeof data.value === 'object'
+        ? (data.value as Record<string, unknown>)
+        : null;
+    const actualValue = valueObj ? valueObj.value : data.value;
```
ravinderkabli: suggestion: This only inspects `value.value`, but for coded/select and multiselect controls the selected answer lives under `value.concept` (a `{uuid, name}` object) while `value.value` can be null/undefined. A mandatory coded field the user *did* answer could therefore be flagged as missing and block a valid save.

The sibling `processControlForNotesExtraction` already extracts `valueObj.concept.uuid` (around line 463) — mirror that here (and consider a shared `isControlValueEmpty(control)` helper to keep the two in sync). Worth verifying against a real Container `state.data` sample for a coded mandatory field, and adding a test for it.

packages/bahmni-services/src/observationFormsService/observationFormsTransformer.ts
```diff
@@ -522,3 +522,51 @@ function processControlForNotesExtraction(
...
+ */
+export function hasMissingMandatoryVisibleField(
+  data: Record<string, unknown> | undefined,
+): boolean {
+  if (!data || typeof data !== 'object') return false;
+
+  const control = data.control as Record<string, unknown> | undefined;
+  const properties = control?.properties as Record<string, unknown> | undefined;
+
+  if (properties?.mandatory === true && !data.hidden && !data.voided) {
```
ravinderkabli: suggestion: For a mandatory node this returns `true` on its own empty value before ever checking `children`. Group/section controls typically have their own `value: null` while the real data lives in the children — so a mandatory group whose children are all filled would be falsely flagged as missing.

Consider only short-circuiting for leaf nodes (no `children`); a container node should be considered satisfied when its descendants satisfy the requirement. A test for a mandatory group with filled children would lock this in.

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
```diff
@@ -276,14 +277,31 @@ const ObservationFormsContainer: React.FC<ObservationFormsContainerProps> = ({
...
+          state?: { data?: Record<string, unknown> | { toJS?: () => unknown } };
+        } | null
+      )?.state?.data;
+      const hasMissingMandatory = hasMissingMandatoryVisibleField(
+        convertImmutableToPlainObject(containerStateData) as
+          | Record<string, unknown>
+          | undefined,
+      );
+
+      if (hasErrors || hasMissingMandatory) {
```
ravinderkabli, line 308: nit: The `isEmpty` guard above (line ~275) returns `VALIDATION_STATE_EMPTY` before this block is reached. So a form whose only requirement is a single empty mandatory field (nothing else filled) reports "empty form" instead of "mandatory field missing". The save is still blocked (core bug doesn't recur) — it's only the message that's off.

Consider computing `hasMissingMandatory` before the `isEmpty` guard and preferring it, e.g. `if (isEmpty && !hasMissingMandatory) { ...EMPTY; return; }`.

ravinderkabli (commented):
## Review for BAH-4704

Solid fix for the silent-save bug — the new `hasMissingMandatoryVisibleField` predicate is well-placed in the service layer, follows the existing Container-traversal pattern, and the core scenario is well tested. All 4 acceptance criteria are functionally met.

My main concern is a couple of **false-positive** edge cases in the emptiness check that could block a *valid* save. Both depend on how form2-controls serializes coded/group nodes, so please verify against a real `state.data` sample.

Posting the 3 highest-value findings inline below (2 suggestions + 1 nit). A few additional minor/informational items (shared-helper extraction, duplicated `.state.data` cast, unrelated `yarn.lock` cmdk churn, test-coverage gaps) are in the local review notes.

**Acceptance Criteria**: 4/4 met (AC2 has a message-fidelity caveat — see the nit).

yarn.lock
```diff
@@ -6960,7 +6960,7 @@ clone@^1.0.2:
 
 cmdk@^1.1.1:
   version "1.1.1"
-  resolved "https://registry.yarnpkg.com/cmdk/-/cmdk-1.1.1.tgz#b8524272699ccaa37aaf07f36850b376bf3d58e5"
+  resolved "https://registry.npmjs.org/cmdk/-/cmdk-1.1.1.tgz#b8524272699ccaa37aaf07f36850b376bf3d58e5"
```
ravinderkabli: nit: This changes `cmdk`'s resolved registry from `registry.yarnpkg.com` to `registry.npmjs.org` — same version (1.1.1) and same integrity hash, so it's harmless. But it's unrelated to the validation fix and is likely incidental churn from a local npm/yarn registry config.

Consider reverting this hunk to keep the PR diff scoped to the fix.

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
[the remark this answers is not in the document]
ROHIT-V-TW: Addressed in commit 94c65b1 — `hasMissingMandatory` is now computed before the `isEmpty` guard, and the guard was updated to `if (isEmpty && !hasMissingMandatory)`.

packages/bahmni-services/src/observationFormsService/observationFormsTransformer.ts
ROHIT-V-TW: Addressed in commit 94c65b1 — the `isLeafValueEmpty` helper checks both `value.value` (text/number/date) and `value.concept` (coded/select). For multi-select it handles the array case (`Array.isArray(concept) ? concept.length === 0`). A mandatory coded field with a selected answer will have a non-null `value.concept` and will not be flagged as missing.

packages/bahmni-services/src/observationFormsService/observationFormsTransformer.ts
ROHIT-V-TW: Addressed in commit 94c65b1 — the function now only checks emptiness on leaf nodes (those without `children`). A mandatory group/section node is skipped for its own value and the recursion descends into its children instead. A mandatory group with all filled children will not be falsely flagged.

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
ROHIT-V-TW: Addressed in commit 94c65b1 — `hasMissingMandatory` is now computed before the `isEmpty` guard, and the guard was changed to `if (isEmpty && !hasMissingMandatory)`. A form with only untouched mandatory fields (nothing else filled) now correctly reports the mandatory-field-missing state.

yarn.lock
ROHIT-V-TW: The `cmdk` entry in yarn.lock is already on `registry.yarnpkg.com` in the current branch — the incidental churn was reverted in commit 94c65b1.

ravinderkabli (dismissed):
Done

ravinderkabli (approved):
Looks fine

PR #532, opened 2026-07-01
BAH- 4801. Create Registration Encounter When a Visit Starts

arshiyaTW2021:
**JIRA** → [BAH-4801](https://bahmni.atlassian.net/jira/software/c/projects/BAH/boards/131?selectedIssue=BAH-4801)
     
  **Summary**

  Migrates the registration encounter flow so the encounter is created once, when the patient's first visit is successfully 
  saved as a single FHIR R4 POST already linked to the visit via partOf instead of being created eagerly at patient
  registration and then linked with a follow-up PUT. This replaces the older BAH-4655 create-then-update approach and removes the now-dead code.

  **Background / previous behavior**

  - A registration encounter was created immediately after patient registration (useCreatePatient).
  - On visit start, the code searched for that encounter (GET), created one if missing, then updated it (PUT) to attach partOf
  + visit period => a create-then-update flow with extra round-trips.

  **New behavior (desired end state)**

  - No registration encounter is created during patient registration.
  - The encounter is created only after POST /visit succeeds, via a single POST /fhir2/R4/Encounter that already contains
  partOf: { reference: "Encounter/<visitUuid>" }.
  - No GET to search for an existing encounter; no PUT to link it.
  - Created only on the first visit for the patient - the existing hasActiveVisit guard prevents a duplicate encounter on
  re-entry while a visit is active.
  - Audit: logs EDIT_ENCOUNTER with encounter UUID, encounter type, patient UUID, and module: registration (module passed explicitly, since the shared EDIT_ENCOUNTER event has no default module and Clinical relies on that).

  **Changes**

  @bahmni/services
  - visitService: corrected createVisitForPatient / createVisit return type from Promise<string> to Promise<CreatedVisit>
  (annotation fix - runtime already returned the visit object). Added minimal CreatedVisit (Pick<ActiveVisitResult, 'uuid' | 
  'startDatetime'>), scoped to the visit service.
  - auditLogService: removed the now-unused CREATE_ENCOUNTER constant; exported MODULE_LABELS (additive, non-breaking) so the registration app can attribute the module explicitly. The shared EDIT_ENCOUNTER constant is unchanged.

  **Registration app**
  - fhirEncounterMapper: buildRegistrationEncounterPayload accepts optional visitUuid (emits partOf) and periodStart,
  normalizing the OpenMRS datetime (colon-less offset) to a valid FHIR instant.
  - registrationEncounterService: createRegistrationEncounterForPatient accepts { visitUuid, periodStart } and dispatches
  EDIT_ENCOUNTER (module registration).
  - useVisit: captures the created visit and creates the encounter with partOf immediately after POST /visit succeeds; removed  the find/create-if-missing/link sequence.
  - useCreatePatient: removed eager registration-encounter creation.
  - Cleanup (Phase 3): deleted the dead findValidRegistrationEncounterInSession, linkRegistrationEncounterToVisit, and their
  private helpers (isEncounterInSession, the registration-local sortByMostRecent), plus their tests and now-unused imports.
  (The identically-named sortByMostRecent in encounterSessionService is a different, live function and was left untouched.)

[BAH-4801]: https://bahmni.atlassian.net/browse/BAH-4801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Visit creation now returns richer details (UUID and start time), enabling downstream registration encounter creation with correct FHIR linkage.
  * Registration encounters can include visit linkage and use the visit start time for generated FHIR `period.start`.
* **Bug Fixes**
  * Patient creation no longer triggers an extra registration encounter.
  * `periodStart` is normalized/validated, with a safe fallback when missing or invalid.
  * Visit/encounter creation flow now more reliably handles configured encounter types, failures, and existing active visits.
* **Audit/Telemetry**
  * Registration encounter audit events were updated for more accurate edit logging.
* **Tests**
  * Updated/expanded visit-patient-encounter flow coverage and FHIR payload mapping assertions.


packages/bahmni-services/src/visitService/visitService.ts
```diff
@@ -50,10 +51,10 @@ const getVisitLocationUUID = async (
 /**
  * Create a new visit for a patient
  * @param visitData - The visit data including patient UUID, visit type, and location
- * @returns Promise<string> - The created visit UUID
+ * @returns Promise<CreatedVisit> - The created visit object
  */
-const createVisit = async (visitData: VisitData): Promise<string> => {
-  return post<string>(CREATE_VISIT_URL, visitData);
+const createVisit = async (visitData: VisitData): Promise<CreatedVisit> => {
+  return post<CreatedVisit>(CREATE_VISIT_URL, visitData);
```
ravinderkabli, line 57: suggestion: The return type now asserts `CreatedVisit` (required `uuid` + `startDatetime`), but `CREATE_VISIT_URL` (`/visit`) sends no `v=custom` representation param — unlike `GET_ACTIVE_VISIT_URL`, which explicitly requests `startDatetime` because default reps omit fields. So this relies on the POST default representation including `startDatetime`, unverified at runtime and untested (the existing `visitService` tests mock `post` → `{}`).

If `startDatetime` comes back missing/malformed, it flows into the mapper as `periodStart` and `new Date(periodStart).toISOString()` throws `RangeError: Invalid time value` — caught by the try/catch in `useVisit`, but the result is a visit with no registration encounter (silent AC1 failure).

Consider adding `?v=custom:(uuid,startDatetime)` to `CREATE_VISIT_URL` and/or guarding `periodStart` in the mapper, plus a `visitService` test mocking a realistic `{uuid, startDatetime}` response.

apps/registration/src/utils/fhirEncounterMapper.ts
```diff
@@ -56,5 +60,15 @@ export const buildRegistrationEncounterPayload = ({
...
-  period: { start: new Date().toISOString() },
+  ...(visitUuid && {
+    partOf: { reference: `Encounter/${visitUuid}` },
+  }),
+  // Normalize to a FHIR-valid instant. OpenMRS returns datetimes with a
+  // colon-less timezone offset (e.g. 2026-07-01T10:00:00.000+0000), which FHIR
+  // rejects; new Date(...).toISOString() yields a valid `...Z` instant.
+  period: {
+    start: periodStart
+      ? new Date(periodStart).toISOString()
```
ravinderkabli: suggestion: This normalizes the OpenMRS colon-less offset via native `new Date(periodStart).toISOString()`, but colon-less timezone offsets are not in the ISO 8601 subset ECMAScript mandates `Date.parse` to support — parsing is engine-dependent. The registration module already standardizes on date-fns for date ops (`fhirPatientToFormData.ts`, `ageUtils.ts`, `patientDataConverter.ts`, `RelationshipRow.tsx`).

Consider using date-fns (`parseISO`/`format`, or `parse` with an explicit `yyyy-MM-dd'T'HH:mm:ss.SSSxx` pattern) for a deterministic, convention-aligned conversion, and guard against an invalid/missing `periodStart`. The current test only covers `+0000`; a `+0530` case would be worth adding.

ravinderkabli (commented):
## Review for BAH-4801

packages/bahmni-services/src/visitService/visitService.ts
ROHIT-V-TW: Fixed — added `?v=custom:(uuid,startDatetime)` to `CREATE_VISIT_URL` so the POST response explicitly includes `startDatetime`. Updated the `visitService` tests to mock a realistic `{uuid, startDatetime}` response and added assertions verifying both fields are returned and the URL contains the custom representation param.

apps/registration/src/utils/fhirEncounterMapper.ts
ROHIT-V-TW: Fixed — replaced `new Date(periodStart).toISOString()` with `parseISO` from date-fns (already used across the registration app) for deterministic parsing of colon-less offsets across engines. Added an `isValid` guard that falls back to `new Date()` on invalid/missing input. Added test cases for a `+0530` offset, missing `periodStart`, and an invalid `periodStart` string.

ravinderkabli (approved):
looks fine

PR #537, opened 2026-07-05
BAH-4828 | Kanchi | Add. Disable already-submitted observation forms …

Kanchi-Oswal-twk:
**BAH-4828 | Disable Already-Submitted Observation Forms in ConsultationPad Form Selector**

**What & Why**

In the ConsultationPad form selector, an observation form already submitted in the current encounter was still selectable — a clinician could pick it again and create a duplicate submission.

This PR disables such forms in the search dropdown: they appear greyed out with an "(Already added)" label and can't be re-selected. A brand-new encounter (or an expired/mismatched session) shows all forms as selectable — no carry-over.

**How It Works**

1. New hook useSubmittedEncounterForms(allForms) reads the active encounter from useEncounterSessionStore(), gated on matchReasons.includes('MATCHED') — the same gate ConsultationPad uses to decide encounterForSubmission.
2. It fetches that encounter's observations (getObservationsBundleByEncounterUuid), reads each observation's form-namespace-path extension, and resolves the form name via the shared extractFormName util → maps to the loaded forms' UUIDs → returns a memoized Set<string>.
3. It refetches on consultationSaved (for the current patient) so newly-submitted forms disable correctly in the "Continue Consultation" multi-bundle flow.
4. ObservationFormsPanel passes submittedFormUuids into ObservationForms, which disables + relabels matching ComboBox items.

▎ Mirrors the existing Investigations/Orders selector pattern (InvestigationsForm.tsx) — a proven approach, not a new one.

**Changes**

- apps/clinical/src/hooks/useSubmittedEncounterForms.ts — New. Derives the set of submitted form UUIDs for the active encounter.
- apps/clinical/src/hooks/__tests__/useSubmittedEncounterForms.test.tsx — New. Unit tests (fetch-gating, mapping, dedup, fail-open, refetch).
- apps/clinical/src/components/forms/observations/ObservationFormsPanel.tsx — Calls the hook, passes submittedFormUuids down.
- apps/clinical/src/components/forms/observations/ObservationForms.tsx — New optional submittedFormUuids prop; a form is disabled when already selected or already submitted.
- apps/clinical/src/components/forms/observations/__tests__/* — Updated selector + panel tests.
- packages/bahmni-widgets/src/forms/utils.ts — New extractFormName(observation) util (wraps extractFormFieldPath, parses the form name).
- packages/bahmni-widgets/src/forms/__tests__/utils.test.ts — Tests for extractFormName.
- packages/bahmni-widgets/src/index.ts — Export extractFormName.

**Out of Scope**

- Edit flow for submitted forms (separate story per the ticket).
- Pinned/default form cards — only the search dropdown is affected.
- No changes to the store, ConsultationPad, or design system; reuses the existing OBSERVATION_FORMS_FORM_ALREADY_ADDED i18n key.

**Acceptance Criteria**

- [x] Submitted form appears disabled/greyed and cannot be selected
- [x] Only submitted forms disabled; unsubmitted remain selectable
- [x] Multiple submitted forms all disabled
- [x] Feedback: (Already added) label, matching Orders & Procedures
- [x] New encounter / expired session → all forms selectable (no carry-over)

Review Follow-ups Addressed

- Shared parsing util: co-located form-name parsing into @bahmni/widgets as extractFormName (removes duplication with extractFormFieldPath).
- Version parsing: single-segment strip (/\.\d+$/) so dotted form names aren't over-stripped.
- Error handling: useQuery error is surfaced via console.error in a useEffect.
- Minor: stable EMPTY_SET default, patient-scoped query key, trimmed useSubscribeConsultationSaved deps.

Reviewer Notes

- Fail-open by design: if the observations fetch fails, no forms are disabled (degrades to prior behaviour) rather than blocking form entry — disabling is a convenience guard, not a hard gate.
- MATCHED-only gate is intentional: it mirrors ConsultationPad's encounterForSubmission. Under LOCATION_MISMATCH/PROVIDER_MISMATCH the pad creates a new encounter, so those forms should stay selectable.

**Screenshots:** 

image

image

image

## Summary

* **New Features**
  * Observation form picker now detects forms already submitted in the current active matched encounter and marks them unavailable.
  * Added support for extracting form names from observation namespace information to power this behavior.
* **Bug Fixes**
  * Disabled submitted forms can’t be selected/added, and the UI updates when saved-consultation events occur.
* **Tests**
  * Expanded unit/integration coverage for the picker/panel behavior, the submitted-forms hook, and new form-name extraction logic.

### DIFF
--- a/apps/clinical/src/components/forms/observations/ObservationForms.tsx
+++ b/apps/clinical/src/components/forms/observations/ObservationForms.tsx
@@ -31,6 +31,8 @@ interface ObservationFormsProps {
   allForms: ObservationForm[];
   isAllFormsLoading: boolean;
   observationFormsError: Error | null;
+  // Set of form UUIDs already submitted in the active encounter (BAH-4828)
+  submittedFormUuids?: Set<string>;
 }
 
 /**
@@ -58,6 +60,7 @@ const ObservationForms: React.FC<ObservationFormsProps> = React.memo(
     allForms,
     isAllFormsLoading,
     observationFormsError,
+    submittedFormUuids = new Set(),
   }) => {
     const { t } = useTranslation();
     const canAddObservations = useHasPrivilege(
@@ -197,18 +200,20 @@ const ObservationForms: React.FC<ObservationFormsProps> = React.memo(
         ];
       }
 
-      // Map forms to ComboBox items with proper labeling for already selected forms
+      // Map forms to ComboBox items with proper labeling for already-selected or already-submitted forms
       const results = availableForms.map((form: ObservationForm) => {
         const isAlreadySelected = selectedForms.some(
           (selected: ObservationForm) => selected.uuid === form.uuid,
         );
+        const isSubmitted = submittedFormUuids.has(form.uuid);
+        const disabled = isAlreadySelected || isSubmitted;
 
         return {
           id: form.uuid,
-          label: isAlreadySelected
+          label: disabled
             ? `${form.name} (${t('OBSERVATION_FORMS_FORM_ALREADY_ADDED')})`
             : form.name,
-          disabled: isAlreadySelected,
+          disabled,
         };
       });
 
@@ -219,6 +224,7 @@ const ObservationForms: React.FC<ObservationFormsProps> = React.memo(
       searchTerm,
       availableForms,
       selectedForms,
+      submittedFormUuids,
       t,
     ]);
 
--- a/apps/clinical/src/components/forms/observations/ObservationForms.tsx
+++ b/apps/clinical/src/components/forms/observations/ObservationForms.tsx
@@ -19,6 +19,8 @@ import {
 import { useObservationFormsStore } from '../../../stores/observationFormsStore';
 import styles from './styles/ObservationForms.module.scss';
 
+const EMPTY_SET = new Set<string>();
+
 interface ObservationFormsProps {
   onFormSelect?: (form: ObservationForm) => void;
   selectedForms?: ObservationForm[];
@@ -31,7 +33,6 @@ interface ObservationFormsProps {
   allForms: ObservationForm[];
   isAllFormsLoading: boolean;
   observationFormsError: Error | null;
-  // Set of form UUIDs already submitted in the active encounter (BAH-4828)
   submittedFormUuids?: Set<string>;
 }
 
@@ -60,7 +61,7 @@ const ObservationForms: React.FC<ObservationFormsProps> = React.memo(
     allForms,
     isAllFormsLoading,
     observationFormsError,
-    submittedFormUuids = new Set(),
+    submittedFormUuids = EMPTY_SET,
   }) => {
     const { t } = useTranslation();
     const canAddObservations = useHasPrivilege(
--- a/apps/clinical/src/hooks/useSubmittedEncounterForms.ts
+++ b/apps/clinical/src/hooks/useSubmittedEncounterForms.ts
@@ -4,20 +4,10 @@ import {
   useEncounterSessionStore,
   useSubscribeConsultationSaved,
 } from '@bahmni/services';
-import { usePatientUUID, extractFormFieldPath } from '@bahmni/widgets';
+import { usePatientUUID, extractFormName } from '@bahmni/widgets';
 import { useQuery } from '@tanstack/react-query';
 import type { Observation } from 'fhir/r4';
-import { useMemo } from 'react';
-
-function parseFormName(valueString: string | undefined): string | undefined {
-  if (!valueString) return undefined;
-  const beforeSlash = valueString.split('/')[0];
-  const name = beforeSlash
-    .split('^')
-    .pop()
-    ?.replace(/(\.\d+)+$/, '');
-  return name ?? undefined;
-}
+import { useEffect, useMemo } from 'react';
 
 /**
  * Returns the set of form UUIDs that have already been submitted in the active encounter.
@@ -47,10 +37,12 @@ export function useSubmittedEncounterForms(
     queryFn: () => getObservationsBundleByEncounterUuid(activeEncounterUuid!),
   });
 
-  if (error) {
-    // eslint-disable-next-line no-console
-    console.error('Failed to fetch submitted encounter forms', error);
-  }
+  useEffect(() => {
+    if (error) {
+      // eslint-disable-next-line no-console
+      console.error('Failed to fetch submitted encounter forms', error);
+    }
+  }, [error]);
 
   useSubscribeConsultationSaved(
     (payload) => {
@@ -71,8 +63,7 @@ export function useSubmittedEncounterForms(
     const submittedUuids = new Set<string>();
 
     for (const obs of observations) {
-      const valueString = extractFormFieldPath(obs);
-      const formName = parseFormName(valueString);
+      const formName = extractFormName(obs);
       if (!formName) continue;
 
       const matched = allForms.find((f) => f.name === formName);
--- a/apps/clinical/src/hooks/useSubmittedEncounterForms.ts
+++ b/apps/clinical/src/hooks/useSubmittedEncounterForms.ts
@@ -1,0 +1,103 @@
+import {
+  FHIR_OBSERVATION_FORM_NAMESPACE_PATH_URL,
+  ObservationForm,
+  getObservationsBundleByEncounterUuid,
+  useEncounterSessionStore,
+  useSubscribeConsultationSaved,
+} from '@bahmni/services';
+import { usePatientUUID } from '@bahmni/widgets';
+import { useQuery } from '@tanstack/react-query';
+import type { Observation } from 'fhir/r4';
+import { useMemo } from 'react';
+
+/**
+ * Parse the form name from an observation's form-namespace-path valueString.
+ *
+ * The valueString format is either:
+ *   - `{namespace}^{formName}.{version}/{fieldPath}`  e.g. `Bahmni^Vitals.1/10-0`
+ *   - `{formName}.{version}/{fieldPath}`              e.g. `Vitals.1/1-0`
+ *
+ * Returns the bare form name (e.g. `Vitals`), or undefined if the value is absent/malformed.
+ */
+function parseFormName(valueString: string | undefined): string | undefined {
+  if (!valueString) return undefined;
+  const beforeSlash = valueString.split('/')[0];
+  // Strip optional `{namespace}^` prefix, then strip trailing `.{version}`
+  const name = beforeSlash
+    .split('^')
+    .pop()
+    ?.replace(/\.\d+$/, '');
+  return name ?? undefined;
+}
+
+/**
+ * Extract the form-namespace-path valueString from a FHIR Observation extension.
+ * Uses the canonical URL constant from @bahmni/services to avoid hardcoding.
+ * This inlines the same logic as `extractFormFieldPath` from @bahmni/widgets/src/forms/utils
+ * to avoid a cross-package deep import.
+ */
+function getFormNamespacePath(obs: Observation): string | undefined {
+  return obs.extension?.find(
+    (ext) => ext.url === FHIR_OBSERVATION_FORM_NAMESPACE_PATH_URL,
+  )?.valueString;
+}
+
+/**
+ * Returns the set of form UUIDs that have already been submitted in the active encounter.
+ *
+ * - Returns an empty set when there is no MATCHED encounter session (new encounter → all forms selectable).
+ * - Automatically refetches after any consultation save for the current patient (handles the
+ *   "Continue Consultation" multi-bundle flow).
+ *
+ * Mirrors the pattern in InvestigationsForm.tsx lines 43-92.
+ */
+export function useSubmittedEncounterForms(
+  allForms: ObservationForm[],
+): Set<string> {
+  const patientUUID = usePatientUUID();
+  const { activeEncounter, matchReasons } = useEncounterSessionStore();
+
+  // Only consider an encounter if the session is actively MATCHED
+  const activeEncounterUuid = matchReasons.includes('MATCHED')
+    ? activeEncounter?.id
+    : undefined;
+
+  const { data: bundle, refetch } = useQuery({
+    queryKey: ['submittedEncounterForms', activeEncounterUuid],
+    enabled: !!activeEncounterUuid && !!patientUUID,
+    staleTime: 30_000,
+    queryFn: () => getObservationsBundleByEncounterUuid(activeEncounterUuid!),
+  });
+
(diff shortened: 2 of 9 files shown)


apps/clinical/src/hooks/useSubmittedEncounterForms.ts
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Parse the form name from an observation's form-namespace-path valueString.
+ *
+ * The valueString format is either:
+ *   - `{namespace}^{formName}.{version}/{fieldPath}`  e.g. `Bahmni^Vitals.1/10-0`
+ *   - `{formName}.{version}/{fieldPath}`              e.g. `Vitals.1/1-0`
+ *
+ * Returns the bare form name (e.g. `Vitals`), or undefined if the value is absent/malformed.
+ */
+function parseFormName(valueString: string | undefined): string | undefined {
```
sumaztwcode: Please check if there is already any utility file to parse this.

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
```diff
@@ -0,0 +1,103 @@
...
+  const beforeSlash = valueString.split('/')[0];
+  // Strip optional `{namespace}^` prefix, then strip trailing `.{version}`
+  const name = beforeSlash
+    .split('^')
+    .pop()
+    ?.replace(/\.\d+$/, '');
+  return name ?? undefined;
+}
+
+/**
```
sumaztwcode: Remove comments - where code is self explanatory

apps/clinical/src/components/forms/observations/ObservationForms.tsx
```diff
@@ -58,6 +60,7 @@ const ObservationForms: React.FC<ObservationFormsProps> = React.memo(
     allForms,
     isAllFormsLoading,
     observationFormsError,
+    submittedFormUuids = new Set(),
```
sumaztwcode: nit: `submittedFormUuids = new Set()` creates a new `Set` on every render when the prop isn't passed, which breaks the `useMemo` dependency below (since `new Set() !== new Set()` by reference). Currently `ObservationFormsPanel` always passes this prop so runtime impact is minimal, but it's a footgun for future callers.

```ts
const EMPTY_SET = new Set<string>();
// then in destructuring:
submittedFormUuids = EMPTY_SET,
```

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Extract the form-namespace-path valueString from a FHIR Observation extension.
+ * Uses the canonical URL constant from @bahmni/services to avoid hardcoding.
+ * This inlines the same logic as `extractFormFieldPath` from @bahmni/widgets/src/forms/utils
+ * to avoid a cross-package deep import.
+ */
+function getFormNamespacePath(obs: Observation): string | undefined {
+  return obs.extension?.find(
+    (ext) => ext.url === FHIR_OBSERVATION_FORM_NAMESPACE_PATH_URL,
+  )?.valueString;
```
sumaztwcode: suggestion: This `getFormNamespacePath` function does the same thing as `extractFormFieldPath` in `packages/bahmni-widgets/src/forms/utils.ts`. The widget utility isn't exported from the public API, which is why it was re-implemented here. Consider exporting it from `@bahmni/widgets` so both packages share the same utility, or add a TODO comment noting the duplication.

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
```diff
@@ -0,0 +1,103 @@
...
+    (payload) => {
+      if (payload.patientUUID === patientUUID) {
+        refetch();
+      }
+    },
+    [patientUUID, refetch],
+  );
+
+  return useMemo(() => {
+    const observations: Observation[] =
```
sumaztwcode, line 57: nit: `useSubscribeConsultationSaved` uses a `callbackRef` pattern internally — it always calls the latest callback regardless of deps. The deps array only controls when the event listener is torn down and re-added. Including `refetch` in `[patientUUID, refetch]` causes unnecessary listener churn without correctness benefit.

```ts
useSubscribeConsultationSaved(
  (payload) => {
    if (payload.patientUUID === patientUUID) {
      refetch();
    }
  },
  [patientUUID], // refetch omitted — callbackRef handles staleness
);
```

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
```diff
@@ -0,0 +1,103 @@
...
+  const activeEncounterUuid = matchReasons.includes('MATCHED')
+    ? activeEncounter?.id
+    : undefined;
+
+  const { data: bundle, refetch } = useQuery({
+    queryKey: ['submittedEncounterForms', activeEncounterUuid],
+    enabled: !!activeEncounterUuid && !!patientUUID,
+    staleTime: 30_000,
+    queryFn: () => getObservationsBundleByEncounterUuid(activeEncounterUuid!),
+  });
```
sumaztwcode, line 38: nit: Consider including `patientUUID` in the query key to scope the cache to a patient-encounter pair, matching the `enabled` guard:

```ts
queryKey: ['submittedEncounterForms', patientUUID, activeEncounterUuid],
```

Encounter UUIDs are globally unique so this is an edge-case concern, but it's a defensive best practice.

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
```diff
@@ -0,0 +1,103 @@
...
+ * Returns the bare form name (e.g. `Vitals`), or undefined if the value is absent/malformed.
+ */
+function parseFormName(valueString: string | undefined): string | undefined {
+  if (!valueString) return undefined;
+  const beforeSlash = valueString.split('/')[0];
+  // Strip optional `{namespace}^` prefix, then strip trailing `.{version}`
+  const name = beforeSlash
+    .split('^')
+    .pop()
+    ?.replace(/\.\d+$/, '');
```
sumaztwcode: nit: The regex `/\.\d+$/` strips a single trailing `.{digits}` version segment. This works for the documented format (`Vitals.1/10-0`), but would fail for multi-part versions like `Vitals.1.2/1-0` (would leave `Vitals.1`). If multi-part versions are possible in production, consider:

```ts
.replace(/(\\.\d+)+$/, '')
```

sumaztwcode (commented):
## Review for BAH-4828

Clean implementation that follows existing patterns well. All 5 acceptance criteria are met. A few minor suggestions below.

Found 0 blocking issues, 5 suggestions, and 2 informational notes.

apps/clinical/src/components/forms/observations/ObservationForms.tsx
Kanchi-Oswal-twk: @sumaztwcode this is fixed. Extracted a module-level const EMPTY_SET = new Set<string>() and used it as the default, so the reference is stable across renders and the useMemo dep behaves correctly.

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
Kanchi-Oswal-twk: Done — removed the redundant JSDoc/inline comments on the helpers. Kept only the JSDoc on the exported useSubmittedEncounterForms hook since it documents non-obvious public behaviour (empty set when not MATCHED, refetch-on-save).

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
Kanchi-Oswal-twk: Done — query key is now ['submittedEncounterForms', patientUUID, activeEncounterUuid], matching the enabled guard and consistent with InvestigationsForm.

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
```diff
@@ -0,0 +1,86 @@
...
+
+function parseFormName(valueString: string | undefined): string | undefined {
+  if (!valueString) return undefined;
+  const beforeSlash = valueString.split('/')[0];
+  const name = beforeSlash
+    .split('^')
+    .pop()
+    ?.replace(/(\.\d+)+$/, '');
+  return name ?? undefined;
+}
```
ravinderkabli: suggestion: `parseFormName`'s version-strip regex `(\.\d+)+$` over-strips — it greedily removes *all* trailing dotted-number groups, not just the single version segment.

- `"COVID.19.2/1-0"` → `"COVID"` (a form named "COVID.19" loses `.19`)
- `"Form 2.0.1/1-0"` → `"Form 2"`

When the parsed name no longer matches an `allForms` entry, that form isn't added to the submitted set and silently stays selectable — the exact duplicate submission this feature is meant to prevent (fail-open). OpenMRS form versions are a single integer, so strip just one segment: `.replace(/\.\d+$/, '')`, and add `it.each` cases for form names ending in `.<digits>`.

Minor follow-on: since `extractFormFieldPath` was just promoted to the public `@bahmni/widgets` surface, consider co-locating `parseFormName` there too and unit-testing it in `forms/__tests__/utils.test.ts` — they're two halves of the same "read a form name from an observation" operation.

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
```diff
@@ -0,0 +1,86 @@
...
+ */
+export function useSubmittedEncounterForms(
+  allForms: ObservationForm[],
+): Set<string> {
+  const patientUUID = usePatientUUID();
+  const { activeEncounter, matchReasons } = useEncounterSessionStore();
+
+  const activeEncounterUuid = matchReasons.includes('MATCHED')
+    ? activeEncounter?.id
+    : undefined;
```
ravinderkabli, line 27: suggestion: The gate is `matchReasons.includes('MATCHED')` only, but `canResumeOwnInSessionEncounter` also allows editing an own in-session encounter under `['LOCATION_MISMATCH']` (same practitioner, different location). In that resume case the user can add observations, yet this hook returns an empty set — already-submitted forms won't be greyed out. It's not one of the written ACs, but it's a gap in the "already added" guarantee. Consider gating on the same predicate that decides own-encounter editability rather than `MATCHED` alone.

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
```diff
@@ -0,0 +1,86 @@
...
+    queryKey: ['submittedEncounterForms', patientUUID, activeEncounterUuid],
+    enabled: !!activeEncounterUuid && !!patientUUID,
+    staleTime: 30_000,
+    queryFn: () => getObservationsBundleByEncounterUuid(activeEncounterUuid!),
+  });
+
+  if (error) {
+    // eslint-disable-next-line no-console
+    console.error('Failed to fetch submitted encounter forms', error);
+  }
```
ravinderkabli: nit: `console.error` runs in the render body, so it re-logs on every re-render while the error state persists. Move it into an effect:

```ts
useEffect(() => {
  if (error) console.error('Failed to fetch submitted encounter forms', error);
}, [error]);
```

packages/bahmni-widgets/src/index.ts
```diff
@@ -25,6 +25,7 @@ export { TaskList } from './tasks';
 export { PatientProgramsTable } from './patientPrograms';
 export { ImmunizationHistory } from './immunizationHistory';
 export { ProgramDetails } from './programDetails';
+export { extractFormFieldPath } from './forms/utils';
```
ravinderkabli, line 35: nit: This exports from the concrete module path `./forms/utils` rather than the `forms` barrel. It resolves fine and creates no collision — purely for consistency you could surface it via `forms/index.ts`. Non-blocking.

ravinderkabli (commented):
Review for BAH-4828

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
Kanchi-Oswal-twk: Done — changed the deps to [patientUUID]. Agreed, useSubscribeConsultationSaved uses a callbackRef internally, so refetch in the deps only caused unnecessary listener churn.

packages/bahmni-widgets/src/index.ts
Kanchi-Oswal-twk: @ravinderkabli  The barrel now exports only extractFormName (the function the outside code actually uses), and it's exported the same way as everything else in this file — straight from its module path, like ./hooks/usePatientUUID. extractFormFieldPath stays private to the package since nothing outside needs it.

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
Kanchi-Oswal-twk: Fixed — moved it into useEffect(() => { if (error) console.error(...) }, [error]) so it only logs when the error changes, not on every re-render.

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
Kanchi-Oswal-twk: @ravinderkabli  (\.\d+)+$ was a regression (I'd changed it from \.\d+$ in an earlier round). Reverted to /\.\d+$/ since OpenMRS form versions are single integers, so a name like COVID.19 is preserved. Added it.each cases for names ending in .<digits> (COVID.19.1/1-0, Bahmni^COVID.19.2/1-0).

Also did the follow-on: moved the name parsing into @bahmni/widgets/forms/utils.ts as extractFormName(observation) (it wraps extractFormFieldPath), exported it, and unit-tested it in forms/__tests__/utils.test.ts. The hook now just calls extractFormName(obs).

apps/clinical/src/hooks/useSubmittedEncounterForms.ts
Kanchi-Oswal-twk: @ravinderkabli , I looked into it and I think keeping the MATCHED-only check is the safer option here, but keen to hear your thoughts.
 From what I can see, the ConsultationPad only adds observations to the existing encounter when the session is MATCHED (index.tsx:208-210). For LOCATION_MISMATCH/PROVIDER_MISMATCH it actually starts a new encounter. So if we greyed out forms in those cases, we'd be basing it on an encounter the pad isn't saving to — and the user might see "already added" on a form they genuinely need in the new encounter.
 Matching the same MATCHED check the pad uses keeps our "already added" list in sync with where the data actually gets saved. Totally open to revisiting if the resume logic changes or if I've missed a case — let me know!

sumaztwcode (approved):

ravinderkabli (approved):
Looks fine

PR #541, opened 2026-07-08
BAH-4815 | Handle logout failures with proper backend error message display

bhagyakarada:
## JIRA
[BAH-4815](https://bahmni.atlassian.net/browse/BAH-4815) — Handle Logout Failures with Backend Error Message Display

## What Changed
- `userService.logout()`: re-throws the original error instead of replacing it with a hardcoded generic key, so real backend/network messages propagate to callers.
- `UserProfileMenu.handleLogout()`: displays the actual formatted error message from `getFormattedError(error)`, falling back to the i18n key only when the message is the generic default (`'An unexpected error occurred'`).
- Loading state cleared in `finally` so the logout button stays clickable after a failure.
- 4 test updates: updated 1 existing test, added 2 new tests covering network error display and generic fallback.

## Why
When logout failed, users saw a hardcoded generic message regardless of the actual cause. This change surfaces the real error (network failure, 400 bad request, etc.) so users understand what went wrong and can decide to retry.

## How Tested
- [x] Unit tests added/updated
- [x] All BAH-4815 affected tests passing (1177 service tests, 44 home tests)
- [x] Lint clean (both packages)

## Acceptance Criteria
- [x] AC1: Network error captured and shown in notification
- [x] AC3: Generic fallback for 500+ without body message
- [x] AC5: Timeout message propagates
- [x] AC6: 400 validation message extracted and shown
- [x] AC7: Logout button stays clickable after failure (retry supported)
- [x] AC8: Loading indicator cleared on error
- [x] AC2: 500+ body message extraction (deferred — requires shared util change)
- [x] AC4: Cookie clearing on 401 (deferred — requires interceptor change)

## Follow-up Items
- **AC2**: `getFormattedError` returns generic server error for 500/502/503/504 without reading response body — extracting body message requires modifying shared utility (dedicated follow-up story).
- **AC4**: Axios interceptor already silently redirects on 401; cookie-clearing before redirect is a shared-infrastructure change (deferred).

[BAH-4815]: https://bahmni.atlassian.net/browse/BAH-4815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logout error handling with clearer messages for network, timeout, server, and session-expiry failures.
  * Expired sessions now redirect directly to the login screen without showing a misleading notification.
  * Backend error messages are handled more safely and truncated when necessary.
  * Logout controls remain disabled while a logout request is in progress and support retry after failure.

* **Localization**
  * Added English and Spanish translations for specific logout failure scenarios.


packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -106,23 +123,26 @@ export const getFormattedError = (
         case 503:
         case 504:
           title = 'Server Error';
-          message = 'The server encountered an error. Please try again later.';
+          message =
```
ravinderkabli: blocking: This changes a SHARED utility. `getFormattedError` has ~24 consumers across apps/home, registration, clinical, patient-documents, widgets, AND the `api/client.ts` interceptor. Base behavior for 5xx was a fixed generic string; now every 500/502/503/504 surfaces raw `data.error`/`data.message` to end users app-wide. For OpenMRS/Tomcat, 5xx bodies often carry stack-trace / SQL / internal exception text — an information-disclosure + UX change well beyond logout.

The PR description itself marks this shared-util change (AC2) as "deferred", yet it ships here undocumented.

Suggestion: reserve `extractBackendMessage` for the 400 (validation) case; keep the generic message for 5xx and log the backend detail to console/telemetry. Or split this into its own PR with product/security sign-off and document the ~24-consumer blast radius.

packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -41,6 +41,25 @@ const parseFHIRError = (outcome: FHIROperationOutcome): string | null => {
   return null;
 };
 
+const extractBackendMessage = (data: unknown): string | undefined => {
```
ravinderkabli, line 45: question: `extractBackendMessage` returns the raw backend string with no length cap/trim. A long stack-trace string would be passed whole to the Carbon toast subtitle. XSS is not a risk (Carbon renders text, no dangerouslySetInnerHTML), but consider truncating (~200 chars) if this is user-facing. Ties into the 5xx info-disclosure concern above.

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
```diff
@@ -31,10 +32,14 @@ export const UserProfileMenu: React.FC = () => {
       await logout();
       window.location.href = LOGIN_PATH;
     } catch (error) {
-      const { title } = getFormattedError(error);
+      if (axios.isAxiosError(error) && error.response?.status === 401) {
```
ravinderkabli: blocking: This 401 handling is redundant with the pre-existing global interceptor in `api/client.ts`, which already does `if (axios.isAxiosError(error) && error.response?.status === 401) { globalThis.location.href = LOGIN_PATH; ... }`. Since `logout()`→`del()` uses that client, on a 401 the interceptor redirects first, then userService clears cookies + re-throws, then this component redirects again — a triple handling of one event. Also, this is the only component under apps/*/src that imports `axios` directly, leaking transport concerns into the UI, and the same predicate now lives in 3 places.

Suggestion: rely on the interceptor for the 401 redirect; keep only the genuinely-new cookie-clear in the service. Extract an `isUnauthorizedError(error)` helper in @bahmni/services so no UI component imports axios.

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
```diff
@@ -31,10 +32,14 @@ export const UserProfileMenu: React.FC = () => {
...
+        window.location.href = LOGIN_PATH;
+        return;
+      }
+      const { title, message } = getFormattedError(error);
```
ravinderkabli: blocking: For non-401 failures, the `api/client.ts` interceptor has ALREADY run `getFormattedError` on the real AxiosError and rejected with the string `"Server Error: <message>"`. So `error` here is a string, and this second `getFormattedError(error)` hits the `typeof error === 'string'` branch → returns `{ title: 'Error', message: 'Server Error: <message>' }`. Result: the notification always shows a generic "Error" title with the real title embedded in the message, and your new status-specific branches (400/5xx/timeout) never run via this component. The tests don't catch it because they mock both `logout` and `getFormattedError`.

Suggestion: pick one formatting layer — either bypass the interceptor for logout so this sees the real AxiosError, or display the interceptor's message directly and drop this second format call.

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
```diff
@@ -31,10 +32,14 @@ export const UserProfileMenu: React.FC = () => {
...
       addNotification({
         title,
-        message: t('HOME_ERROR_LOGOUT_FAILED'),
+        message,
```
ravinderkabli: blocking: Localization regression. This removes the localized `HOME_ERROR_LOGOUT_FAILED` (which had en + es) and shows raw English from `getFormattedError` (e.g. 'Server Error', 'Request timed out. Please try again.') with no `t()`. Spanish users now see untranslated English. `getFormattedError` also returns an inconsistent mix of English text and bare i18n-keys (e.g. 'ERROR_DUPLICATE_ACTIVE_MEDICATION'), so the call site can't reliably localize.

Suggestion: have `getFormattedError` return i18n keys (or a typed discriminator) localized at the call site, or keep a localized fallback for the generic case and add the new strings to the locale files.

ravinderkabli (commented):
## Review for BAH-4815

packages/bahmni-services/src/errorHandling/utils.ts
bhagyakarada: I updated the implementation to avoid changing the shared behavior of getFormattedError() for server errors.

The utility now continues to extract backend messages only for the appropriate client-side error cases (e.g. 400), while retaining the existing generic fallback for 5xx responses.

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
bhagyakarada: Addressed. The logout notification now uses the existing localized i18n strings (HOME_ERROR_LOGOUT_FAILED_TITLE / HOME_ERROR_LOGOUT_FAILED) instead of displaying the raw strings returned by getFormattedError(). getFormattedError() is now only used for logging, so this avoids the localization regression while preserving the error details for debugging

packages/bahmni-services/src/errorHandling/utils.ts
bhagyakarada: We no longer surface backend messages for 5xx responses through the shared getFormattedError() utility and instead retain the existing generic fallback message. As a result, raw stack traces or other long backend error messages are no longer displayed in the Carbon toast for server errors.

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
bhagyakarada: I removed the second getFormattedError() call from the component, so we're no longer formatting an already formatted error. The logout flow now relies on a single formatting layer instead of re-processing the interceptor output, which avoids the generic "Error" title with the formatted message embedded in it. The notification shown by the component is now driven by the localized logout fallback, while 401s continue to follow the existing interceptor flow.

ravinderkabli (commented):
#Review

packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -41,6 +41,51 @@ const parseFHIRError = (outcome: FHIROperationOutcome): string | null => {
...
+  | 'timeout' // request aborted because it timed out
+  | 'server' // HTTP 5xx
+  | 'unknown'; // anything else (non-Axios error, other 4xx, etc.)
+
+export const getErrorKind = (error: unknown): ErrorKind => {
+  if (axios.isAxiosError(error)) {
+    if (error.response) {
+      if (error.response.status === 401) return 'unauthorized';
+      if (error.response.status >= 500) return 'server';
+      return 'unknown';
```
ravinderkabli, line 81: blocking: AC6 ("400 validation message extracted and shown") is checked off in the PR description, but this line collapses 400 (and 403/404/409/422) into `unknown`, which `UserProfileMenu` maps to the generic `HOME_ERROR_LOGOUT_FAILED`. The backend validation message is never surfaced — the new `extractBackendMessage()` helper is only wired into `getFormattedError`, not this path.

Either mark AC6 as deferred (like AC2), or add a `badRequest` kind here that surfaces `extractBackendMessage(error.response.data)`.

packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -41,6 +41,51 @@ const parseFHIRError = (outcome: FHIROperationOutcome): string | null => {
...
+};
+
+export type ErrorKind =
+  | 'unauthorized' // HTTP 401 — session already ended
+  | 'network' // no response received (offline / DNS / CORS)
+  | 'timeout' // request aborted because it timed out
+  | 'server' // HTTP 5xx
+  | 'unknown'; // anything else (non-Axios error, other 4xx, etc.)
+
+export const getErrorKind = (error: unknown): ErrorKind => {
```
ravinderkabli, line 76: suggestion: `getFormattedError` still exists and still classifies errors — so this adds a second, exported classifier with a *different* taxonomy. `getFormattedError` distinguishes 403/404/400; `getErrorKind` flattens all of them to `unknown`. The name reads as a general-purpose classifier but the behaviour is logout-shaped.

Prefer making the taxonomy complete (`forbidden`/`notFound`/`badRequest`/…) and refactoring `getFormattedError` to build on top of `getErrorKind`, so classification lives in one place. If it's really logout-only, scope the name (e.g. `getLogoutErrorKind`) rather than exporting it as a shared primitive.

packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -108,18 +151,18 @@ export const getFormattedError = (
...
           message =
-            responseData?.message ??
+            extractBackendMessage(axiosError.response.data) ??
             axiosError.message ??
             'An unknown error occurred';
-        }
       }
-    } else if (error instanceof Error) {
-      message = error.message;
+    } else if (axiosError.code === 'ECONNABORTED') {
```
ravinderkabli, line 161: suggestion: This replaces the previous `else if (error instanceof Error) { message = error.message }` branch. Since every AxiosError is `instanceof Error`, all no-response Axios errors across the (~17) `getFormattedError` consumers now get generic strings instead of the raw `error.message`. Likely the intended fix, but it's a shared-util behaviour change with a wide blast radius — worth calling out in the description and confirming no consumer relied on the raw message.

packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -41,6 +41,51 @@ const parseFHIRError = (outcome: FHIROperationOutcome): string | null => {
...
+    typeof backendError === 'object' &&
+    backendError !== null &&
+    typeof (backendError as Record<string, unknown>).message === 'string'
+  ) {
+    message = (backendError as Record<string, unknown>).message as string;
+  } else if (typeof responseData.message === 'string') {
+    message = responseData.message;
+  }
+  if (message && message.length > MAX_MESSAGE_LENGTH) {
+    return message.substring(0, MAX_MESSAGE_LENGTH) + '...';
```
ravinderkabli, line 64: question: Truncating to 200 chars limits payload size but doesn't sanitize. It's safe today because Carbon renders notification text escaped by React — worth a one-line comment noting this string must never reach `dangerouslySetInnerHTML`.

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
```diff
@@ -31,10 +31,24 @@ export const UserProfileMenu: React.FC = () => {
       await logout();
       window.location.href = LOGIN_PATH;
     } catch (error) {
-      const { title } = getFormattedError(error);
+      const kind = getErrorKind(error);
+      if (kind === 'unauthorized') {
+        window.location.href = LOGIN_PATH;
```
ravinderkabli, line 36: suggestion: AC4 asks for local session/auth data to be *cleared* before this silent redirect. The redirect itself is correct (no notification), but nothing clears cookies/session state here — stale local state could linger. At minimum clear local auth keys before redirecting; track the interceptor-level cookie clearing as the referenced follow-up.

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
```diff
@@ -31,10 +31,24 @@ export const UserProfileMenu: React.FC = () => {
...
+        return;
```
ravinderkabli, line 37: nit: `return` here doesn't skip `finally`, so `setIsLoggingOut(false)` still runs after navigation has started — a redundant state update on a navigating component (harmless, but the spinner should really stay until unload). Consider `if (kind !== 'unauthorized') setIsLoggingOut(false)` instead of the `finally`.

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
```diff
@@ -31,10 +31,24 @@ export const UserProfileMenu: React.FC = () => {
...
+      }
+
+      const messageKey =
```
ravinderkabli, line 40: nit: A 4-level nested ternary — a `Record<ErrorKind, string>` lookup is flatter and auto-extends when new `ErrorKind` values are added:
`const kindToKey: Record<ErrorKind, string> = { network: 'HOME_ERROR_LOGOUT_NETWORK', timeout: 'HOME_ERROR_LOGOUT_TIMEOUT', server: 'HOME_ERROR_LOGOUT_SERVER', unauthorized: 'HOME_ERROR_LOGOUT_FAILED', unknown: 'HOME_ERROR_LOGOUT_FAILED' };`

packages/bahmni-services/src/errorHandling/__tests__/utils.test.ts
```diff
@@ -295,3 +308,51 @@ describe('getFormattedError', () => {
     });
   });
 });
+
+describe('getErrorKind', () => {
```
ravinderkabli, line 312: nit: Nice coverage of `getErrorKind`. Two gaps worth locking down: a `getFormattedError` test for a 4xx with `{ error: { message } }` (the new `extractBackendMessage` path) incl. the 200-char truncation boundary, and a component-level test asserting 400/403 → generic logout message.

packages/bahmni-services/src/cdssService/__tests__/cdssService.test.ts
```diff
@@ -1,4 +1,4 @@
-import { get, post } from '../../api';
+import { post } from '../../api';
```
ravinderkabli, line 1: question: This unused-import cleanup is unrelated to BAH-4815 — fine to leave, but ideally a separate cleanup commit (or a line in the description) so the change surface stays focused.

packages/bahmni-services/src/errorHandling/utils.ts
ravinderkabli, line 81: 🔴 **CRITICAL — AC6 claimed done but not implemented**

The PR body checks `- [x] AC6: 400 validation message extracted and shown`. In reality, this `return 'unknown'` line is reached for **every** non-401, non-5xx response — including 400. The PR's own test confirms it: _"classifies a non-5xx/non-401 response (e.g. 400) as unknown"_.

`'unknown'` maps to `HOME_ERROR_LOGOUT_FAILED` ("Logout failed. Please try again.") in `UserProfileMenu`. A `400` with `{"error":"Field X is required"}` shows the generic fallback — the backend validation message is never read or shown.

**Root cause**: `extractBackendMessage()` is wired only into `getFormattedError()`, not into the `handleLogout` catch path which uses `getErrorKind()` exclusively. AC6's requirement ("error message from response is extracted and displayed") has no code path in the logout flow.

**Fix**: Either (a) correct the PR description to mark AC6 as deferred — matching AC2's honest treatment — or (b) add a `badRequest`/4xx kind here (`if (error.response.status >= 400 && error.response.status < 500) return 'badRequest';`) and surface `extractBackendMessage(error.response.data)` in the notification for that case.

packages/bahmni-services/src/errorHandling/utils.ts
ravinderkabli, line 76: 🟠 **MAJOR — `getErrorKind` name over-promises: partial, logout-shaped taxonomy that diverges from `getFormattedError`**

`getFormattedError` was **not** deleted — it still exists, is still exported, and still formats `{ title, message }`. This PR adds `getErrorKind` as a second exported sibling with a **different, incompatible taxonomy**:

| Status code | `getFormattedError` | `getErrorKind` |
|---|---|---|
| 401 | "Unauthorized" | `'unauthorized'` |
| 403 | "Unauthorized" | `'unknown'` ← diverges |
| 404 | "Not Found" | `'unknown'` ← diverges |
| 400 | backend message | `'unknown'` ← diverges |
| 5xx | "Server Error" | `'server'` |

The name `getErrorKind` reads as a general-purpose, exhaustive classifier exported from `@bahmni/services` as a shared primitive — but it's partial and logout-flow-shaped. A `403` during logout (e.g. CSRF mismatch) shows a generic toast instead of the `401` silent-redirect path. This divergence is also the root cause of the AC6 miss above.

**Fix — pick one**:
- **(a) Scope the name**: rename to `getLogoutErrorKind` and/or co-locate it with the consumer instead of exporting as a shared primitive.
- **(b) Complete the taxonomy (preferred)**: give `getErrorKind` a full set (`'unauthorized' | 'forbidden' | 'notFound' | 'badRequest' | 'server' | 'timeout' | 'network' | 'unknown'`) and refactor `getFormattedError` to build on top of it — one classification source of truth, and the logout flow gets a `'badRequest'` kind to satisfy AC6.

packages/bahmni-services/src/errorHandling/utils.ts
ravinderkabli, line 161: 🟠 **MAJOR — `getFormattedError` behavior change silently affects all ~17 existing consumers**

This line replaces the previous `else if (error instanceof Error) { message = error.message }` branch. Since every `AxiosError` is `instanceof Error`, the old code surfaced the raw `error.message` for all no-response Axios errors. Now every consumer of `getFormattedError` gets generic strings for network/timeout failures — a change in observable output across ~17 call sites (widgets, hooks, services — `FormsTable`, `LabInvestigation`, `ObservationFormsContainer`, the API `client.ts` interceptor, etc.) with no assertions updated for those consumers.

The new test comment explicitly acknowledges this: _"A real AxiosError is an `instanceof Error`, which previously short-circuited the network/timeout branches"_ — so the fix is intentional and correct, but it is a shared-utility behavioral change with wide blast radius.

**Fix**: Confirm affected consumers are fine losing the raw message on network errors (likely yes, since generic messages are better UX). Call out this change explicitly in the PR description. If the real `error.message` was useful for diagnostics, preserve it in the `console.error` path (it already logs `'Logout failed:', error`, which is good).

packages/bahmni-services/src/errorHandling/utils.ts
ravinderkabli, line 64: 🔵 **INFO — Truncation is size-limiting, not sanitization**

Truncating to 200 chars limits payload size but does not strip HTML or script tags — a crafted backend string could contain `<script>` within those 200 chars.

This is **safe today** because Carbon's `ToastNotification` renders text as React children (escaped by the JSX runtime, no `dangerouslySetInnerHTML`), so there is no XSS path in the current rendering chain.

**Suggestion**: Add a short comment here noting that future callers must never pass this string to `dangerouslySetInnerHTML`, to guard against someone inadvertently introducing that footgun later:
```ts
// NOTE: caller must never pass to dangerouslySetInnerHTML — React escaping is relied upon
```

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
ravinderkabli, line 36: 🟠 **MAJOR — AC4 partial: 401 redirects correctly but local session/auth data is not cleared**

AC4 requires that on a `401` the local session/auth data is **cleared** before the silent redirect. The redirect itself is correct — no notification is shown, which is the right UX — but nothing clears cookies, `sessionStorage`, or any Zustand auth state before navigating away.

The PR honestly flags cookie-clearing as a deferred interceptor change, so this is partial not missing. But the PR checklist marks AC4 as `✅` when it should be `⚠️`.

**Fix**: At minimum clear local auth keys (e.g. any `sessionStorage` auth tokens, Zustand auth slice reset) before `window.location.href = LOGIN_PATH`. Track the interceptor-level cookie clearing as a follow-up story that explicitly references AC4 so it doesn't get lost.

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
ravinderkabli, line 37: 🟡 **MINOR — `finally` re-runs `setIsLoggingOut(false)` on the 401 early-return path**

`return` inside `catch` does not skip `finally`. After `window.location.href = LOGIN_PATH` starts navigation, the component still calls `setIsLoggingOut(false)` — a redundant state update on a navigating/unmounting component.

Harmless in practice: React 18 suppresses the "state update on unmounted component" warning. But semantically the spinner should stay active until the page unloads, not be cleared mid-navigation.

**Fix**:
```tsx
// Option A — guard in finally
finally {
  if (kind !== 'unauthorized') setIsLoggingOut(false);
}

// Option B — explicit clearing only in non-redirect branches, no finally
catch (error) {
  const kind = getErrorKind(error);
  if (kind === 'unauthorized') {
    window.location.href = LOGIN_PATH;
    return; // spinner intentionally left active during navigation
  }
  // ... show notification ...
  setIsLoggingOut(false);
}
```

apps/home/src/components/UserProfileMenu/UserProfileMenu.tsx
ravinderkabli, line 40: 🟡 **MINOR — 4-level nested ternary for message-key mapping**

A deeply nested ternary is hard to read and brittle to extend — adding a new `ErrorKind` value requires finding the right nesting level and risks misplacing it.

A `Record<ErrorKind, string>` lookup is flat, self-documenting, and gets TypeScript exhaustiveness checking for free (the compiler will error if a new `ErrorKind` variant isn't handled):

```tsx
const kindToKey: Record<ErrorKind, string> = {
  network:      'HOME_ERROR_LOGOUT_NETWORK',
  timeout:      'HOME_ERROR_LOGOUT_TIMEOUT',
  server:       'HOME_ERROR_LOGOUT_SERVER',
  unauthorized: 'HOME_ERROR_LOGOUT_FAILED', // won't reach here (early return above)
  unknown:      'HOME_ERROR_LOGOUT_FAILED',
};
const messageKey = kindToKey[kind];
```

packages/bahmni-services/src/errorHandling/__tests__/utils.test.ts
ravinderkabli, line 312: 🟡 **MINOR — Missing tests: 400/403 logout paths and `getFormattedError` 400 extraction boundary**

Good coverage of `getErrorKind`'s happy paths. Two gaps worth closing:

**1. `getFormattedError` — 400 extraction path untested**
The new `extractBackendMessage()` helper added in this PR has no test asserting it is actually used for a 400 response in `getFormattedError`. Add:
```ts
it('extracts the backend error message for a 400 response', () => {
  const error = { response: { status: 400, data: { error: { message: 'Field X is required' } } } } as AxiosError;
  expect(getFormattedError(error).message).toBe('Field X is required');
});

it('truncates backend messages longer than 200 characters', () => {
  const long = 'x'.repeat(250);
  const error = { response: { status: 400, data: { message: long } } } as AxiosError;
  expect(getFormattedError(error).message).toHaveLength(203); // 200 + '...'
});
```

**2. Component-level 400/403 behaviour unlocked**
No test asserts that a 400 or 403 logout error shows the generic message (rather than a backend body). Without this, a future refactor that accidentally surfaces backend text for 4xx won't be caught.

packages/bahmni-services/src/cdssService/__tests__/cdssService.test.ts
ravinderkabli, line 1: 🔵 **INFO — Unrelated import cleanup (scope creep)**

Removing the unused `get` import is unrelated to BAH-4815. Harmless, but it widens the change surface and makes a targeted revert of this PR slightly more work.

Ideally in a separate lint-fix commit, or at minimum called out in the PR description so reviewers understand this is incidental cleanup.

packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -41,6 +41,51 @@ const parseFHIRError = (outcome: FHIROperationOutcome): string | null => {
...
+
+export type ErrorKind =
+  | 'unauthorized' // HTTP 401 — session already ended
+  | 'network' // no response received (offline / DNS / CORS)
+  | 'timeout' // request aborted because it timed out
+  | 'server' // HTTP 5xx
+  | 'unknown'; // anything else (non-Axios error, other 4xx, etc.)
+
+export const getErrorKind = (error: unknown): ErrorKind => {
+  if (axios.isAxiosError(error)) {
```
ravinderkabli, line 77: 🟠 **MAJOR — `getErrorKind` name over-promises: partial, logout-shaped taxonomy that diverges from `getFormattedError`**

`getFormattedError` was **not** deleted — it still exists, is still exported, and still formats `{ title, message }`. This PR adds `getErrorKind` as a second exported sibling, creating two overlapping error utilities with **divergent, incompatible taxonomies**:

| Status code | `getFormattedError` | `getErrorKind` |
|---|---|---|
| 401 | "Unauthorized" | `'unauthorized'` |
| 403 | "Unauthorized" | `'unknown'` ← diverges |
| 404 | "Not Found" | `'unknown'` ← diverges |
| 400 | backend message | `'unknown'` ← diverges |
| 5xx | "Server Error" | `'server'` |

The name `getErrorKind` reads as a general-purpose, exhaustive classifier (it's exported from `@bahmni/services` as a shared primitive), but the implementation is partial and logout-flow-shaped. This divergence is also the root cause of the AC6 miss (see next comment on line 42).

**Fix — pick one**:
- **(a) Scope the name**: rename to `getLogoutErrorKind` and/or co-locate it with the consumer instead of exporting as a shared primitive.
- **(b) Complete the taxonomy (preferred)**: give `getErrorKind` a full set (`'unauthorized' | 'forbidden' | 'notFound' | 'badRequest' | 'server' | 'timeout' | 'network' | 'unknown'`) and refactor `getFormattedError` to build on top of it — one classification source of truth, and the logout flow gets a `'badRequest'` kind to satisfy AC6.

packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -41,6 +41,51 @@ const parseFHIRError = (outcome: FHIROperationOutcome): string | null => {
...
+    if (error.response) {
+      if (error.response.status === 401) return 'unauthorized';
+      if (error.response.status >= 500) return 'server';
+      return 'unknown';
+    }
```
ravinderkabli, line 82: 🔴 **CRITICAL — AC6 claimed done but not implemented: 400 validation message is never shown**

The PR body checks `- [x] AC6: 400 validation message extracted and shown`. But this `return 'unknown'` is reached for **every** non-401, non-5xx response — including 400. The PR's own test confirms it: _"classifies a non-5xx/non-401 response (e.g. 400) as unknown"_.

`'unknown'` maps to `HOME_ERROR_LOGOUT_FAILED` ("Logout failed. Please try again.") in `UserProfileMenu`. A `400` with `{"error":"Field X is required"}` shows the generic fallback — the backend validation message is never read or shown.

**Root cause**: `extractBackendMessage()` is wired only into `getFormattedError()`, not into the `handleLogout` catch path which uses `getErrorKind()` exclusively. AC6's requirement ("error message from response is extracted and displayed") has no code path in the logout flow.

**Fix**: Either (a) correct the PR description to mark AC6 as deferred — matching AC2's honest treatment — or (b) add a `badRequest` kind here (`if (error.response.status >= 400 && error.response.status < 500) return 'badRequest';`) and surface `extractBackendMessage(error.response.data)` in the notification for that case in `UserProfileMenu`.

packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -108,18 +151,18 @@ export const getFormattedError = (
...
-            responseData?.message ??
+            extractBackendMessage(axiosError.response.data) ??
             axiosError.message ??
             'An unknown error occurred';
-        }
       }
-    } else if (error instanceof Error) {
-      message = error.message;
+    } else if (axiosError.code === 'ECONNABORTED') {
+      // Request timed out — a real AxiosError has no `response` here.
```
ravinderkabli, line 162: 🟠 **MAJOR — `getFormattedError` behavior change silently affects all ~17 existing consumers**

This line replaces the previous `else if (error instanceof Error) { message = error.message }` branch. Since every `AxiosError` is `instanceof Error`, the old code surfaced the raw `error.message` for all no-response Axios errors. Now every consumer of `getFormattedError` gets generic strings for network/timeout failures — a change in observable output across ~17 call sites (`FormsTable`, `LabInvestigation`, `ObservationFormsContainer`, `api/client.ts` interceptor, etc.) with no assertions updated for those consumers.

The new test comment explicitly acknowledges this: _"A real AxiosError is an `instanceof Error`, which previously short-circuited the network/timeout branches"_ — so the fix is intentional and correct, but it is a shared-utility behavioral change with wide blast radius.

**Fix**: Confirm affected consumers are fine losing the raw message on network errors, and call out this change explicitly in the PR description. If the real `error.message` was useful for diagnostics, preserve it in the `console.error` path (the component already logs `'Logout failed:', error` — good).

packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -41,6 +41,51 @@ const parseFHIRError = (outcome: FHIROperationOutcome): string | null => {
...
+    backendError !== null &&
+    typeof (backendError as Record<string, unknown>).message === 'string'
+  ) {
+    message = (backendError as Record<string, unknown>).message as string;
+  } else if (typeof responseData.message === 'string') {
+    message = responseData.message;
+  }
+  if (message && message.length > MAX_MESSAGE_LENGTH) {
+    return message.substring(0, MAX_MESSAGE_LENGTH) + '...';
+  }
```
ravinderkabli, line 65: 🔵 **INFO — Truncation is size-limiting, not sanitization**

Truncating to 200 chars limits payload size but does not strip HTML or script tags. This is **safe today** because Carbon's `ToastNotification` renders text as React children (escaped by the JSX runtime, no `dangerouslySetInnerHTML`), so there is no XSS path in the current rendering chain.

Consider adding a short comment noting that future callers must never pass this string to `dangerouslySetInnerHTML`, to guard against that footgun later.

ravinderkabli (approved):
Looks fine

PR #549, opened 2026-07-09
BAH-4732 | Edit Observation Forms from Form Display Control

hamsavarthiniR-Bahmni:
## Summary

- **Edit flow**: Edit icon in FormsTable rows and modal header opens observation form in right panel with pre-populated data; `Edit Observations` privilege gate
- **FHIR bundle**: Parent obsGroup → POST (with existing uuid); leaf obs → PUT; voided → DELETE; empty addMore slots skipped; status echo-back for PUT to avoid OpenMRS rejection
- **Done button**: Disabled on edit-form open; enables on any real field change; re-disables when user restores original value (snapshot comparison using CarbonContainer baseline)
- **File upload**: Session-level filename cache for round-trip via `valueAttachment.title`; Complex obs → string URL for CarbonContainer (prevents `indexOf` crash); ObsMapper null-safety guard
- **UI polish**: Form name section title below EncounterDetails in edit mode; Actions column per-accordion-group only when editable rows exist; pinned forms hide already-submitted entries; Continue Consultation title in ConsultationPad
- **Layout sections**: `extractControls` recurses into no-concept containers so nested obs are captured for change detection

## Summary by CodeRabbit

* **New Features**
  * Added an edit-only observation forms flow with FHIR preloading, correct form version selection, and edit-mode labels/titles.
  * Enabled privilege-gated inline editing in the forms table (including session-aware visibility).
  * Added support for observation values provided via attachment extensions.
* **Bug Fixes**
  * Improved active encounter handling to only show details when the session is in a matched state; hid edit actions for expired/mismatched sessions.
  * Removed submitted pinned forms from the pinned list and ensured observations use exact form-name matching.
  * Refined edit saving and grouped observation status/clearing behavior so the correct updates are sent.


packages/bahmni-widgets/src/utils/Observations.ts
```diff
@@ -114,6 +115,19 @@ export function extractObservationValue(
...
+  const attachmentExt = observation.extension?.find(
+    (ext) =>
+      ext.url === FHIR_OBSERVATION_VALUE_ATTACHMENT_URL &&
+      ext.valueAttachment?.url,
+  );
+  if (attachmentExt?.valueAttachment?.url) {
+    return createObservationValue(
+      attachmentExt.valueAttachment.url,
+      'string',
+      isAbnormal,
```
sumaztwcode, line 127: Is the Observation value always an abnormal or normal? or can it hold other reference values?

packages/bahmni-widgets/src/forms/FormsTable.tsx
```diff
@@ -229,14 +259,33 @@ const FormsTable: React.FC<WidgetProps> = ({
     return undefined;
   }, [metadataError, formDataError]);
 
-  const headers = useMemo(
+  // Base headers without Actions — shared across all groups.
```
sumaztwcode, line 281: remove comment, code is self explanatory

packages/bahmni-widgets/src/forms/__tests__/FormsTable.test.tsx
```diff
@@ -782,6 +784,293 @@ describe('FormsTable', () => {
...
+      // encounter-2 and encounter-3 are not the active encounter — no edit icon
+      expect(
+        screen.queryByTestId('edit-form-encounter-2'),
+      ).not.toBeInTheDocument();
+      expect(
+        screen.queryByTestId('edit-form-encounter-3'),
+      ).not.toBeInTheDocument();
+    });
+
+    it('hides Actions column when disableActions is true (no active visit, AC #2)', async () => {
```
sumaztwcode: do not give reference to ACs/ what is AC nos change? Give AC content in short, if required

packages/bahmni-widgets/src/forms/FormsTable.tsx
```diff
@@ -56,6 +64,28 @@ const FormsTable: React.FC<WidgetProps> = ({
     forms,
   } = (config ?? {}) as FormsTableConfig;
 
+  const canEditObservations = useHasPrivilege(
+    CONSULTATION_PAD_PRIVILEGES.EDIT_OBSERVATIONS,
+  );
+  // Actions column shown only when:
+  //  - user has Edit Observations privilege (AC #5)
```
sumaztwcode: do not give comments where code is self explanatory; do not give reference to ACs/ what is AC nos change? Give AC content in short, if required

apps/clinical/src/utils/fhir/observationResourceCreator.ts
```diff
@@ -26,3 +26,179 @@ export const createObservationResources = (
...
+ *
+ * For grouped observations (groupMembers):
+ *   - Children are processed first with the rules above.
+ *   - If the parent has a uuid and ALL children are deleted → parent is also DELETE.
+ *   - If the parent has a uuid and some children remain     → parent is POST with
+ *     resource.id set to the existing uuid.  OpenMRS recognises the id and updates
+ *     the obsGroup in place; new child entries are linked via hasMember.
+ *   - If the parent has no uuid                             → parent is POST.
+ */
+export function createObservationEntriesWithVerbs(
```
sumaztwcode: Naming convention not correct, do not include "WithVerbs"...if you have removed earlier API usage intentionally - can rename

apps/clinical/src/utils/fhir/observationResourceCreator.ts
```diff
@@ -1,6 +1,6 @@
 import { getFhirObservations, FhirReference } from '@bahmni/form2-controls';
-import { Form2Observation } from '@bahmni/services';
-import { Observation, Reference } from 'fhir/r4';
+import { Form2Observation, createBundleEntry } from '@bahmni/services';
+import { BundleEntry, Observation, Reference } from 'fhir/r4';
 
 export const createObservationResources = (
```
sumaztwcode: This does not seem to be called now, but tests refer it? createObservationResources - you could modiy this API itself, instead of introducing a new one?

apps/clinical/src/utils/fhir/observationResourceCreator.ts
```diff
@@ -26,3 +26,179 @@ export const createObservationResources = (
...
+  const entries: BundleEntry[] = [];
+
+  const options = {
+    patientReference: subjectReference as FhirReference,
+    encounterReference: encounterReference as FhirReference,
+    performerReference: performerReference as FhirReference,
+    basedOnReference: basedOn as FhirReference | undefined,
+  };
+
+  const processObservation = (obs: Form2Observation): string | null => {
```
sumaztwcode, line 38: The priority logic - post new observations (as this is applicable while creating new observations, and not just in edit case). Edit may hold secondary importance. Most frequently, it will be used to add new observations. So order the handling in possible priority of operations considering normal scenarios/ usecases

apps/clinical/src/utils/fhir/observationResourceCreator.ts
```diff
@@ -26,3 +26,179 @@ export const createObservationResources = (
...
+      }
+      return null;
+    }
+
+    // No uuid + voided = was never saved, skip
+    return null;
+  };
+
+  for (const obs of observations) {
+    processObservation(obs);
```
sumaztwcode: make sure notes that are added/ edited are preserved correctly in both new/ edit cases

apps/clinical/src/services/encounterBundleService.ts
```diff
@@ -22,7 +22,7 @@ import {
   createEncounterDiagnosisResource,
   createEncounterConditionResource,
 } from '../utils/fhir/conditionResourceCreator';
-import { createObservationResources } from '../utils/fhir/observationResourceCreator';
+import { createObservationEntriesWithVerbs } from '../utils/fhir/observationResourceCreator';
```
sumaztwcode: retain earlier name, also func name with "withVerbs" is not as per std

apps/clinical/src/components/forms/observations/ObservationFormsPanel.tsx
```diff
@@ -69,6 +72,80 @@ const ObservationFormsPanel: React.FC<ObservationFormsPanelProps> = ({
...
+    getObservationsBundleByEncounterUuid(editEncounterUuid)
+      .then((bundle) => {
+        const form2Observations = getObservationsFromFhir(bundle);
+
+        if (form2Observations.length > 0) {
+          // Build uuid → status map from the raw FHIR bundle so PUT requests can
+          // echo back the same status value (OpenMRS rejects status changes and
+          // also errors when status is absent on PUT).
+          const statusByUuid = buildStatusMap(bundle as Bundle);
+          const enrichedObservations = enrichObservationsWithStatus(
```
sumaztwcode: What is the significance of this method? enrichedObservations - can it be renamed to a more meaningful name as per usage context

Need to understand the flow in this API better - will re-review

apps/clinical/src/components/forms/observations/ObservationForms.tsx
```diff
@@ -136,6 +136,14 @@ const ObservationForms: React.FC<ObservationFormsProps> = React.memo(
     // Step 3: Combine - defaults first, then user-pinned
     const allPinnedForms = [...sortedDefaultForms, ...sortedUserPinnedForms];
 
+    // Step 4: Remove already-submitted forms so they don't appear in pinned tiles.
+    // Computed once here to avoid two separate .filter() calls in the JSX.
+    const visiblePinnedForms = useMemo(
+      () => allPinnedForms.filter((f) => !submittedFormUuids.has(f.uuid)),
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
sumaztwcode: ?? this is not expected, disabling lint

apps/clinical/src/components/forms/observations/styles/ObservationFormsContainer.module.scss
```diff
@@ -37,6 +37,15 @@
   }
 }
 
+// Section title shown in edit mode below EncounterDetails — mirrors the form
+// name title that appears in the ActionArea when adding a form.
+.editFormSectionTitle {
+  padding: $spacing-04 $spacing-06;
+  font-size: 1rem;
+  font-weight: 600;
```
sumaztwcode, line 45: hardcoded value - no carbon token?

apps/clinical/src/components/dashboardSection/DashboardSection.tsx
```diff
@@ -24,7 +24,13 @@ const DashboardSection: React.FC<DashboardSectionProps> = ({
   const { matchReasons, canEditOrCreate, activeEncounter } =
     useEncounterSessionStore();
   const noActiveVisit = matchReasons.includes('NO_ACTIVE_VISIT');
-  const activeEncounterUuid = activeEncounter?.id ?? null;
+  // Only expose the active encounter UUID when the session is genuinely MATCHED.
```
sumaztwcode: wherever code is self-explanatory, remove comments

sumaztwcode (changes requested):
Make sure you test edit notes as well; need to understand the flow in couple of places, will revisit; will share custom review file as well.

apps/clinical/src/components/forms/observations/styles/ObservationFormsContainer.module.scss
hamsavarthiniR-Bahmni: done

packages/bahmni-widgets/src/utils/Observations.ts
hamsavarthiniR-Bahmni: it is not always abnormal or normal — it reads the actual interpretation from the
  FHIR observation, same as all other value types. The code is valid and consistent with the existing pattern
  throughout the function.

apps/clinical/src/components/forms/observations/ObservationFormsPanel.tsx
```diff
@@ -69,6 +77,116 @@ const ObservationFormsPanel: React.FC<ObservationFormsPanelProps> = ({
...
+
+    const matchingForm = allForms.find(
+      (form) => form.name.toLowerCase() === editFormName.toLowerCase(),
+    );
+    if (!matchingForm) return;
+    if (selectedForms.some((f) => f.uuid === matchingForm.uuid)) return;
+
+    getObservationsBundleByEncounterUuid(editEncounterUuid)
+      .then(async (bundle) => {
+        const form2Observations = getObservationsFromFhir(bundle);
```
ravinderkabli: **🔴 CRITICAL — Edit fetch has no voided-observation guard; deleted observations can be resurrected on save**

Edit mode fetches via `Observation/$fetch-all?encounter={uuid}` and feeds every returned observation straight into `getObservationsFromFhir` and then the forms store, with **no client-side `entered-in-error` / voided filter**.

QA desk-check (Narmadha, item 4) explicitly flagged that `$fetch-all` should not return voided obs. Each fetched observation carries a `uuid`, so on the next Save it is emitted as a **PUT** by `createObservationEntries` — meaning a previously voided/deleted observation would be silently **un-deleted**. This is a data-integrity defect on a core clinical path.

**Suggested fix:**
Filter before populating the store, e.g.:
```ts
const form2Observations = getObservationsFromFhir(bundle).filter(
  (obs) => obs.status !== 'entered-in-error' && !obs.voided
);
```
Alternatively (or additionally) add `&status:not=entered-in-error` to the fetch URL. Add a test asserting that a voided obs in the fetched bundle never becomes a PUT on save.

apps/clinical/src/components/forms/observations/ObservationFormsPanel.tsx
```diff
@@ -69,6 +77,116 @@ const ObservationFormsPanel: React.FC<ObservationFormsPanelProps> = ({
...
+      !editEncounterUuid ||
+      isAllFormsLoading
+    )
+      return;
+
+    const matchingForm = allForms.find(
+      (form) => form.name.toLowerCase() === editFormName.toLowerCase(),
+    );
+    if (!matchingForm) return;
+    if (selectedForms.some((f) => f.uuid === matchingForm.uuid)) return;
```
ravinderkabli: **🟠 MAJOR — Edit-mode fetch effect double-fires due to stale-uuid guard**

The guard `selectedForms.some(f => f.uuid === matchingForm.uuid)` is checked before the fetch, but on success the form is added via `addForm({ ...matchingForm, uuid: savedFormUuid })` when `savedFormUuid !== matchingForm.uuid`. Since `selectedForms` is in the dependency array, the guard no longer short-circuits on the next run (the store holds `savedFormUuid`, not `matchingForm.uuid`), so the entire async fetch chain (`getObservationsBundleByEncounterUuid` + `getPatientFormData` + `fetchFormUuidByObservationDate`) fires a **second time** — a redundant network round-trip and a second `setState`. Bounded but a real stale-guard bug that happy-path tests miss.

**Suggested fix:**
Use a `useRef` latch keyed on `editEncounterUuid` (e.g. `hasInitiatedEditFetch`) set **before** the fetch starts, and remove `selectedForms` from the dependency array:
```ts
const hasInitiatedFetchRef = useRef<string | null>(null);
// ...
if (hasInitiatedFetchRef.current === editEncounterUuid) return;
hasInitiatedFetchRef.current = editEncounterUuid;
// proceed with fetch
```

ravinderkabli (changes requested):
## Review: BAH-4732 | Edit Observation Forms

ravinderkabli (dismissed):
Looks fine

ravinderkabli (approved):
Looks fine

sumaztwcode (approved):

PR #554, opened 2026-07-14
BAH - 4871 | Record waivers for a vaccine if unable to administer vaccine

devsingh05298:
JIRA → BAH-4871 (https://bahmni.atlassian.net/browse/BAH-4871)

Description

Implements the "Vaccine Waiver Capture" story: a Panel physician/Nurse can now record why a vaccine was not administered — a clinical contraindication or a non-clinical reason such as applicant refusal — as a FHIR R4 Immunization resource with status: "not-done" and a coded statusReason, for IOM and destination-country compliance/reporting. This adds a third immunization input control (immunizationWaiver) alongside the existing history/administration controls, and makes the immunizationHistory display control's "+" button action-specific so each configured display-control instance launches the correct input control instead of always opening the full consultation pad.

---
Key Features

- New waiver capture flow in the Consultation Pad: search/select a vaccine → pick a coded waiver reason → optionally add a free-text note (required when the reason is "Other").
- The vaccine picker for waivers is sourced from FHIR Medication (CVX-coded) resources — the same data already used by the "Vaccinations" display control — rather than the OpenMRS concept ValueSet used by the administered-vaccine forms.
- The immunization display control's "+" button now dispatches an editOnly target (existing startConsultation mechanism) based on configuration, so a display control scoped to "not-done" records opens only the waiver input control, not the entire pad.
- The "Not Administered" table now shows a note indicator (tooltip icon) next to the vaccine name when a waiver note was captured, and its vaccine-name column is now labeled "Type" with content-based column sizing instead of forced equal-width columns.

---
Implementation Details

- UI Components
  - SelectedImmunizationItem: added a statusReason ComboBox, gated by config attributes like the other optional fields; the existing free-text note field becomes required when the selected reason matches the configured "Other" concept.
  - NotAdministeredTab (display control): added a TooltipIcon for notes (mirrors the existing administered-table pattern), overrode the vaccine-name column header to "Type" for this table only, and added a notAdministeredTable layout class for content-sized columns.
  - No new components were added to Storybook — this reuses existing ComboBox/TooltipIcon primitives from @bahmni/design-system.
- State Management
  - Extended the immunization Zustand store: new statusReason field on ImmunizationInputEntry, waiverReasonConfig state (otherReasonConceptUuid), updateStatusReason action, and a value-conditional validation rule requiring a note when the "Other" reason is selected (independent of the attribute-driven required-fields table).
- Custom Hooks
  - None added.
- API Integration
  - Vaccine search for the waiver form uses the existing getVaccinations() service (GET /Medication?code=http://hl7.org/fhir/sid/cvx|) instead of searchFHIRConcepts(vaccineConceptSetUuid).
  - statusReason options are sourced via the existing searchFHIRConcepts(statusReasonValueSetUuid) pattern (same mechanism already used for route/site).
  - createImmunizationBundleEntries now accepts an isWaiver flag: when true, it sets status: "not-done", includes the statusReason coding, and omits administration-only fields (route, site, manufacturer, lotNumber, protocolApplied, expirationDate).
  - No backend contract changes — this only changes what the frontend sends to the existing FHIR Immunization create endpoint.
- Performance & Optimization
  - No new performance work; reuses existing staleTime: Infinity query caching conventions for the new statusReason ValueSet fetch.
- Internationalisation
  - Added new keys to apps/clinical/public/locales/locale_en.json: IMMUNIZATION_INPUT_CONTROL_STATUS_REASON_LABEL/PLACEHOLDER/REQUIRED, NO_MATCHING_STATUS_REASON_FOUND, IMMUNIZATION_HISTORY_WIDGET_COL_TYPE; renamed IMMUNIZATION_HISTORY_WIDGET_NOT_ADMINISTERED_TAB_TITLE to avoid duplicating the dashboard section title. es locale not yet updated — tracked as a follow-up.
- Accessibility
  - Reuses existing accessible primitives (ComboBox, TooltipIcon with aria-label); jest-axe accessibility tests extended and passing for the updated components.
- Limitations
  - Editing a previously recorded waiver is explicitly out of scope for this story.
  - Only one waiver reason per vaccine — enforced naturally by the single-select field, no separate multi-select guard needed.
  - The special case "capture date of immunity when reason = Already immune" was explicitly descoped for this story per product direction during implementation.

---
Testing & Type Definitions

- Unit Tests
  - getVaccineComboBoxItems (dedup, search filtering, fallback display logic) and the isWaiver branch of createImmunizationBundleEntries.
  - updateStatusReason store action and the conditional "note required for Other reason" validation logic.
  - statusReason ComboBox rendering/selection and conditional note behavior in SelectedImmunizationItem.
  - Waiver vaccine-picker sourcing from Medication resources instead of the concept ValueSet in ImmunizationForm.
  - notes extraction in createNotAdministeredImmunizationViewModel.
  - editOnly add-button targeting for both fixed-status and tabbed display-control configurations.
  - Note tooltip icon, "Type" header override, and notAdministeredTable layout class in NotAdministeredTab.
  - Full suites verified clean after these changes: 2181/2181 tests passing in @bahmni/clinical-app, 1464/1464 in @bahmni/widgets.
- Integration Tests
  - Existing ImmunizationForm.integration.test.tsx full-flow suite continues to pass unmodified; no new dedicated integration test beyond the component-level tests listed above.
- Type Definitions
  - Added ImmunizationStatusReason and WaiverReasonConfig types; extended ImmunizationInputEntry, CreateImmunizationBundleEntriesParams, and ImmunizationHistoryState.
  - Added ImmunizationHistoryWidgetConfig (typed display-control config, replacing loose as casts) and extended NotAdministeredImmunizationViewModel with notes.
  - tsc --noEmit diffed against the pre-change baseline for both affected projects — no new type errors introduced.

---
Screenshots

┌────────────────────────────────────────────────────────┬───────────────────┐
│                    Feature / State                     │    Screenshot     │
├─────────────
Screenshot 2026-07-14 at 8 23 26 AM
───────────────────────────────────────────┼

---

▎ [!IMPORTANT]
▎ Checklist
▎
▎ - [ ] Code adheres to project linting and formatting standards — repo-wide core.autocrlf corruption currently breaks the lint hook across unrelated projects too; the actual committed diffs are clean LF. Two minor real style nits were found on manual review (a ||→?? suggestion that would change behavior and was intentionally kept, and ternary line-wrapping in utils.ts) — happy to apply --fix for the latter if preferred.
▎ - [ ] New components added to Storybook — N/A, no new components added.
▎ - [x] Tests written and passing (unit + integration).
▎ - [ ] Labels and text support i18n — en complete, es pending (see Next Steps).
▎ - [x] Follows accessibility and responsive design guidelines.
▎ - [ ] PR reviewed by at least one other developer.

---
Reviewer(s)

@bahnew/developers
Kindly review the proposed changes when convenient. Your feedback is appreciated.

## Summary by CodeRabbit

* **New Features**
  * Added support for recording immunization waivers using medication-based vaccine selection.
  * Added status-reason selection, including conditional notes for “Other” reasons.
  * Added immunization type and recorded-date columns, with sorting and note tooltips for not-administered records.
  * Improved configuration for opening administered and not-administered immunization forms.

* **Bug Fixes**
  * Prevented errors when location data is incomplete.
  * Improved handling of missing immunization details and search results.

* **Documentation**
  * Updated English and Spanish translations for new labels and validation messages.


packages/bahmni-widgets/src/immunizationHistory/model.ts
```diff
@@ -32,3 +34,17 @@ export interface AdministeredTabConfig {
...
   columns: string[];
 }
+
+export interface ImmunizationHistoryWidgetConfig {
+  status?: 'completed' | 'not-done';
+  encounterType?: string;
+  startEncounterPrivilege?: string;
+  administeredFields?: string[];
+  notAdministeredFields?: string[];
+  /** Input control key the "+" button opens when `status` is fixed to one value. */
```
rahu1ramesh: avoid adding comments

apps/clinical/public/locales/locale_en.json
```diff
@@ -260,7 +263,7 @@
   "IMMUNIZATION_HISTORY_WIDGET_DETAIL_SITE": "Site",
   "IMMUNIZATION_HISTORY_WIDGET_ERROR_FETCHING_DATA": "Error fetching data",
   "IMMUNIZATION_HISTORY_WIDGET_NOT_ADMINISTERED_TABLE_ARIA": "Not administered immunizations table",
-  "IMMUNIZATION_HISTORY_WIDGET_NOT_ADMINISTERED_TAB_TITLE": "Vaccination waiver",
```
rahu1ramesh, line 263: Is waiver the default title for product as well @mohan-13, sounds a lot like a domain term might be okay, just cross checking.

apps/clinical/public/locales/locale_es.json
rahu1ramesh, line 1: Some of the keys are missing here in locale_es, please do verify and update the file.

apps/clinical/src/components/consultationPad/__tests__/index.test.tsx
```diff
@@ -533,6 +533,48 @@ describe('ConsultationPad', () => {
       });
     });
 
+    it('dispatches an ADD_VACCINE_NOT_ADMINISTERED_REASON audit event when an immunizationWaiver entry has data', async () => {
```
rahu1ramesh: Do we need this test here, most logic related to forms are handled within the component, this will bloat the tests unneccessarily.

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -463,6 +463,18 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
         messageParams: { encounterType: result.encounterTypeName },
       });
 
+      if (
```
rahu1ramesh: Let's not add a entry specific audit log here, this needs to be discussed and figured out. Add a todo and a card on jira to track this.

apps/clinical/src/components/forms/immunization/__tests__/ImmunizationForm.test.tsx
```diff
@@ -545,6 +547,108 @@ describe('ImmunizationForm', () => {
...
+  describe('Waiver form type', () => {
+    const waiverVaccinationBundle = {
+      resourceType: 'Bundle',
+      type: 'searchset',
+      entry: mockVaccineMedicationsWithDisplay.map((resource) => ({
+        resource,
+      })),
+    };
+
+    it('sources vaccine search results from Medication resources rather than the concept ValueSet', async () => {
```
rahu1ramesh: The test description is confusing, this gives me the impression that earlier it used to fetch from valueset. Is that the case.

apps/clinical/src/components/forms/immunization/__tests__/ImmunizationForm.test.tsx
```diff
@@ -545,6 +547,108 @@ describe('ImmunizationForm', () => {
...
+      await user.click(screen.getByText('Influenza vaccine'));
+      await waitFor(() => {
+        expect(mockStore.addImmunization).toHaveBeenCalledWith({
+          code: '88',
+          display: 'Influenza vaccine',
+        });
+      });
+    });
+
+    it('disables the vaccineConceptSetUuid query even when a vaccineConceptSetUuid is configured', () => {
```
rahu1ramesh: The test description can be rephrased to avoid confusion.

apps/clinical/src/components/forms/immunization/ImmunizationForm.tsx
```diff
@@ -126,7 +141,8 @@ const ImmunizationForm = ({
   } = useQuery({
     queryKey: ['vaccineConceptSetUuid', vaccineConceptSetUuid],
     queryFn: () => searchFHIRConcepts(vaccineConceptSetUuid!),
-    enabled: !!vaccineConceptSetUuid && !isConfigLoading && !configError,
+    enabled:
```
rahu1ramesh, line 146: All other calls expect what is relevant should be disabled for waiver, this needs to be fixed.

apps/clinical/src/components/forms/immunization/ImmunizationForm.tsx
```diff
@@ -175,6 +191,17 @@ const ImmunizationForm = ({
...
   });
 
+  const {
+    data: statusReasonConceptSet,
+    isLoading: statusReasonConceptSetLoading,
+    error: statusReasonConceptSetError,
+  } = useQuery({
+    queryKey: ['statusReasonValueSetUuid', statusReasonValueSetUuid],
+    queryFn: () => searchFHIRConcepts(statusReasonValueSetUuid!),
+    enabled: !!statusReasonValueSetUuid && !isConfigLoading && !configError,
```
rahu1ramesh: Enable this only if this the key is waiver.

apps/clinical/src/components/forms/immunization/ImmunizationForm.tsx
```diff
@@ -98,13 +103,23 @@ const ImmunizationForm = ({
...
+    | string
+    | undefined;
 
   useEffect(() => {
     if (attributes) {
       setAttributes(attributes);
     }
   }, [attributes, setAttributes]);
 
+  useEffect(() => {
```
rahu1ramesh, line 119: `setWaiverReasonConfig` called unconditionally. The effect fires even when `otherReasonConceptUuid` is undefined, we can avoid this call.

apps/clinical/src/components/forms/immunization/__tests__/SelectedImmunizationItem.test.tsx
```diff
@@ -786,6 +796,148 @@ describe('SelectedImmunizationItem', () => {
     });
   });
 
+  describe('Status reason field', () => {
+    const statusReasonAttributes = [{ name: 'statusReason', required: false }];
+
+    it('shows all reason options by default, without requiring the user to type', async () => {
```
rahu1ramesh, line 802: Test description confusing, the options only show up when we start typing.

apps/clinical/src/components/forms/immunization/__tests__/SelectedImmunizationItem.test.tsx
```diff
@@ -786,6 +796,148 @@ describe('SelectedImmunizationItem', () => {
...
+      await user.click(screen.getByText('Not age appropriate'));
+      await waitFor(() => {
+        expect(mockStore.updateStatusReason).toHaveBeenCalledWith(id, {
+          code: 'not-age-appropriate',
+          display: 'Not age appropriate',
+        });
+      });
+    });
+
+    it('calls updateStatusReason with null when the selection is cleared', async () => {
```
rahu1ramesh, line 836: Do we need this assertions?

apps/clinical/src/components/forms/immunization/__tests__/SelectedImmunizationItem.test.tsx
rahu1ramesh, line 1: We would also need to show the "Add Note" link if the reason is not "Other".

apps/clinical/src/components/forms/immunization/__tests__/stores.test.ts
```diff
@@ -437,6 +437,173 @@ describe('useImmunizationHistoryStore', () => {
...
+      });
+
+      expect(store().selectedImmunizations[0].statusReason).toEqual({
+        code: 'not-age-appropriate',
+        display: 'Not age appropriate',
+      });
+      expect(store().selectedImmunizations[1]).toEqual(otherEntryBefore);
+    });
+
+    it('is a no-op for a non-existent id', () => {
```
rahu1ramesh: What is no-op?

apps/clinical/src/components/forms/immunization/__tests__/utils.test.ts
```diff
@@ -134,6 +141,51 @@ describe('getValueSetComboBoxItems', () => {
...
+  it('returns all items when searchTerm is empty, without requiring a search', () => {
+    expect(
+      getAllValueSetComboBoxItems('', mockVaccineValueSet, 'No results'),
+    ).toEqual([
+      { code: 'covid-19', display: 'COVID-19 Vaccine' },
+      { code: 'flu', display: 'Influenza Vaccine' },
+    ]);
+  });
+
+  it('returns all items when searchTerm is whitespace-only', () => {
```
rahu1ramesh: Some of these tests can be clubbed together with it.each, better to refactor this.

apps/clinical/src/components/forms/immunization/components/SelectedImmunizationItem.tsx
rahu1ramesh, line 1: Would be a good time to extract the tanstack calls into a custom hook, worth looking into

apps/clinical/src/components/forms/immunization/utils.ts
```diff
@@ -107,6 +107,49 @@ export function getValueSetComboBoxItems(
...
+): ValueSetComboBoxItem[] {
+  if (!searchTerm.trim()) return [];
+  const seen = new Set<string>();
+  const items: ValueSetComboBoxItem[] = [];
+  (medications ?? []).forEach((med) => {
+    const code = med.code?.coding?.[0]?.code;
+    if (!code || seen.has(code)) return;
+    const display = med.code?.coding?.[0]?.display ?? med.code?.text ?? code;
+    if (!display.toLowerCase().includes(searchTerm.toLowerCase())) return;
+    seen.add(code);
```
rahu1ramesh, line 144: The seen seems redundant here.

apps/clinical/src/components/forms/immunization/utils.ts
```diff
@@ -251,17 +294,34 @@ export function createImmunizationBundleEntries({
...
   isAdministration,
+  isWaiver,
 }: CreateImmunizationBundleEntriesParams): BundleEntry[] {
   return selectedImmunizations.map((entry) => {
     const extensions = [
       ...(entry.drug ? resolveAdministeredProductExtension(entry.drug) : []),
       ...resolveBasedOnExtension(entry.basedOnReference),
       ...resolveStockLocationExtension(entry.stockLocation),
     ];
+    const administeredOnlyFields = isWaiver
```
rahu1ramesh, line 305: This also is little tricky to read, try to break it down

packages/bahmni-widgets/src/immunizationHistory/__tests__/NotAdministeredTab.test.tsx
```diff
@@ -89,6 +95,7 @@ describe('NotAdministeredTab', () => {
     expect(screen.getByText('Hepatitis B')).toBeInTheDocument();
     expect(screen.getByText('Patient refused')).toBeInTheDocument();
     expect(screen.getByText('19-3-2026')).toBeInTheDocument();
+    expect(screen.getByText('19-3-2026 10:30 AM')).toBeInTheDocument();
```
rahu1ramesh: Do we need this time here.

packages/bahmni-widgets/src/immunizationHistory/ImmunizationHistory.tsx
```diff
@@ -73,10 +77,20 @@ const ImmunizationHistory: React.FC<WidgetProps> = ({ config }) => {
...
+    if (status === 'completed' || status === 'not-done') {
+      return widgetConfig?.inputControlKey;
+    }
+    return selectedIndex === 0
+      ? widgetConfig?.administeredInputControlKey
+      : widgetConfig?.notAdministeredInputControlKey;
+  };
+
   const handleAddImmunization = () => {
+    const editOnly = resolveAddButtonInputControlKey();
```
rahu1ramesh, line 99: Why do we need `editOnly` in the event?

rahu1ramesh (changes requested):

apps/clinical/src/components/consultationPad/index.tsx
devsingh05298: removed.

apps/clinical/src/components/forms/immunization/__tests__/ImmunizationForm.test.tsx
devsingh05298: Improved to :   it('sources vaccine search results from Medication resources instead of the vaccine concept ValueSet used by the administration/history form', async () => {

apps/clinical/src/components/forms/immunization/__tests__/ImmunizationForm.test.tsx
devsingh05298: Improve to : it('leaves the vaccineConceptSetUuid query disabled for the waiver form type, even when vaccineConc
eptSetUuid is configured', () => {

apps/clinical/src/components/forms/immunization/__tests__/SelectedImmunizationItem.test.tsx
devsingh05298: Yes, for additional comments or notes we need "Add Note" link if the reason is not "Other".

apps/clinical/src/components/forms/immunization/__tests__/SelectedImmunizationItem.test.tsx
devsingh05298: By default ,here means , its not based on search of each character , the moment you click on combo box all options are shown in drop down

apps/clinical/src/components/forms/immunization/ImmunizationForm.tsx
devsingh05298: **Fixed**. Wrapped the setWaiverReasonConfig call in if (otherReasonConceptUuid) { ... }, matching the guard style already used by the attributes effect right above it. Now the store update is skipped entirely for immunizationHistory/immunizationAdministration controls, where otherReasonConceptUuid is always undefined.

apps/clinical/src/components/forms/immunization/__tests__/SelectedImmunizationItem.test.tsx
devsingh05298: Yes, this test is needed . The onChange handler has two branches:
if (selectedItem?.code && !selectedItem.disabled) {
  updateStatusReason(id, { code: ..., display: ... });
} else {
  updateStatusReason(id, null);
}
This is currently the only test in the file that exercises the else branch (calling updateStatusReason(id, null)) — nothing else clicks the "Clear selected item" button or selects a disabled sentinel item. Without it, that branch would have zero coverage.

It's also not a trivial/plumbing assertion — statusReason is a required field (per statusReasonAttributes config), and correctly resetting it to null on clear matters functionally: the store's validation (FIELD_VALIDATIONS) and the "Other reason requires a note" check (isOtherReasonSelected, based on immunization.statusReason?.code) both depend on the value actually being nulled out rather than left stale when the user clears their selection. So this guards a real behavior, not just Carbon's ComboBox internals.

apps/clinical/src/components/forms/immunization/__tests__/utils.test.ts
devsingh05298: Refactored.

apps/clinical/src/components/forms/immunization/utils.ts
devsingh05298: Clubbed the 6 tests in getAllValueSetComboBoxItems down to 3 it.each blocks:
- ''/'   ' searchTerm cases (both "return all items without requiring a search") → one it.each(['', '   ']).
- The three disabled-sentinel cases (no match / undefined valueSet / empty contains) all assert the identical output, just from different input conditions → one it.each varying (label, searchTerm, valueSet).

packages/bahmni-widgets/src/immunizationHistory/__tests__/NotAdministeredTab.test.tsx
devsingh05298: replaced with : expect(mockFormatDateTime).toHaveBeenCalledWith(mockRow.date, expect.anything());
expect(mockFormatDateTime).toHaveBeenCalledWith(mockRow.recordedOn, expect.anything(), true);

packages/bahmni-widgets/src/immunizationHistory/ImmunizationHistory.tsx
devsingh05298: handleAddImmunization → dispatches startConsultation with editOnly: resolveAddButtonInputControlKey() → ConsultationPad reads it as editOnlyKey (consultationPad/index.tsx:63) → passed into getActiveEntries(registry, encounterType, editOnlyKey) (consultationPad/utils.ts:40-59):

// When editOnly is set, show only the target form + encounterDetails.
if (editOnlyKey) {
  return (
    entry.key === editOnlyKey ||
    entry.key === ENCOUNTER_DETAILS_INPUT_CONTROL_KEY
  );
}
return true; // no editOnly → every input control registered for this encounterType

Why it's needed here specifically: three separate input controls now share encounterType: "Immunization" — immunizationHistory, immunizationAdministration, and the new immunizationWaiver. If editOnly were omitted, the pad would fall back to showing every input control matching that encounter type — i.e., clicking + on any one of the immunization display-control instances would open the pad with all three immunization forms stacked together. That's exactly the bug reported earlier in this session ("why do I see two forms on the immunization history display control") — root-caused to a display-control config missing inputControlKey/editOnly. editOnly is the mechanism that keeps each "+" button scoped to only the one form it's meant to open (administered vs. waiver, or whichever tab is active when status is unset).

apps/clinical/src/components/forms/immunization/utils.ts
devsingh05298: if its not a waiver it should show one Combo box for Reasons , if its administered vaccine the other fields ( route, batch number and manufacturer ,also needs to be there.

apps/clinical/src/components/forms/immunization/__tests__/stores.test.ts
devsingh05298: no-op : leaves selectedImmunizations unchanged when the id does not exist. Improving it to :  it('leaves selectedImmunizations unchanged when the id does not exist', () => {

apps/clinical/src/components/forms/immunization/ImmunizationForm.tsx
devsingh05298: done

apps/clinical/src/components/forms/immunization/ImmunizationForm.tsx
devsingh05298: done

apps/clinical/src/components/consultationPad/__tests__/index.test.tsx
devsingh05298: Removing as we longer going for Audit entry for Vaccine waiver.

apps/clinical/public/locales/locale_en.json
devsingh05298: It would be Not Administered for product, and waiver not the default title. Improved. 
 Added a title field to ImmunizationHistoryWidgetConfig — when configured, it overrides the hardcoded getTitleByStatus(status) default (IMMUNIZATION_HISTORY_WIDGET_ADMINISTERED_TAB_TITLE/_NOT_ADMINISTERED_TAB_TITLE/_WIDGET_TITLE), exactly mirroring how inputControlKey/editTitle already work — fully backward compatible when not set.
- Wired the "Vaccination Waiver" display control block in general.json with "title": "VACCINE_NOT_ADMINISTERED_REASON_WIDGET_TITLE" → renders as "Vaccine Not Administered Reason" instead of the hardcoded "Waiver Details". (FOR PRODUCT)

packages/bahmni-widgets/src/immunizationHistory/model.ts
devsingh05298: removed.

packages/bahmni-widgets/src/immunizationHistory/ImmunizationHistory.tsx
rahu1ramesh: I am a little unsure about this, @mohan-13 can you confirm if this is the agreed on pattern.

mohan-13: The event payload needs to be standardised, tech debt card has been created. https://bahmni.atlassian.net/browse/BAH-4838

rahu1ramesh (approved):

PR #555, opened 2026-07-15
BAH-4811 | Fix snomed url codeable concept

sahupadma:
https://bahmni.atlassian.net/browse/BAH-4811

Fix snomed url codeable concept

## Summary by CodeRabbit

* **Enhancements**
  * Preserve optional concept-system metadata when adding diagnoses and converting them to conditions.
  * When available, include the concept-system in generated FHIR Condition coding; fall back to previous behavior when it isn’t provided.
* **Tests**
  * Added unit and integration coverage to ensure concept-system is retained across workflows and correctly applied to FHIR coding with and without the optional value.


apps/clinical/src/utils/fhir/conditionResourceCreator.ts
```diff
@@ -15,6 +15,7 @@ export const createEncounterDiagnosisResource = (
   encounterReference: Reference,
   recorderReference: Reference,
   recordedDate: Date,
+  diagnosisConceptSystem?: string,
```
ravinderkabli, line 18: suggestion: The new `conceptSystem → coding.system` behaviour has no test coverage, despite existing suites for all three touched areas.

**conditionResourceCreator.test.ts** — the current tests mock `createCoding` and assert single-arg calls only. Add two cases:
1. Pass `'http://snomed.info/sct'` as `diagnosisConceptSystem` / `conditionConceptSystem` and assert `createCoding` was called with `(uuid, 'http://snomed.info/sct')`.
2. Omit the param and assert the single-arg call (regression guard).

**encounterBundleService.test.ts** — give the mock diagnosis/condition a `conceptSystem` and assert `(result[0].resource as Condition).code.coding[0].system` equals it.

**conditionsAndDiagnosesStore.test.ts** — add `conceptSystem` to a mock concept, call `addDiagnosis` then `markAsCondition`, and assert the field survives on both `selectedDiagnoses[0]` and `selectedConditions[0]` (Jest `toEqual` silently ignores `undefined` keys, so the existing tests will not catch a missing prop).

ravinderkabli (commented):
## Review for BAH-4811

ravinderkabli (dismissed):
Looks fine

apps/clinical/src/utils/fhir/conditionResourceCreator.ts
sahupadma: added test

ravinderkabli (dismissed):
Looks fine

ravinderkabli (approved):
Done

PR #560, opened 2026-07-20
BAH-4770 | Unify header user menu across apps

himanshiaaradhana:
## What
Unifies the header user menu across all apps onto the shared design-system `Header` and `@bahmni/widgets` `UserGlobalAction`.

- **Phase 2** – Add `brandName`/`brandPrefix`/`brandHref` + `globalFeatures` slots to `Header`.
- **Phase 3** – Migrate Home to those slots + shared `UserGlobalAction`; delete app-local `UserProfileMenu`.
- **Phase 4** – Deprecate `Header.extraContent` (+ README migration note).
- Fix brand-name color (white at rest and on hover).
- Fix user-menu dropdown keyboard navigation (first-item focus on open + reliable ArrowUp/ArrowDown).

## Testing
- Unit tests updated/added (widgets, home, design-system); clinical/appointments header snapshots refreshed.
- Verified in-browser: grouped menu renders in all four apps, brand text white, keyboard nav works both directions.

## Summary by CodeRabbit

- **New Features**
  - Added a unified user menu across Home, Appointments, Clinical, and Registration.
  - Users can view their greeting, change their password, and log out from the menu.
  - Added loading indicators, error notifications, and improved keyboard navigation.
  - Updated headers to support branding and location-selection content.
- **Localization**
  - Added English and Spanish translations for user-menu actions, greetings, loading states, and errors.
- **Style**
  - Improved header branding, menu layout, focus states, and responsive alignment.


packages/bahmni-design-system/src/organisms/header/models.ts
```diff
@@ -36,13 +36,33 @@ export interface HeaderGlobalAction {
  * Props for the Header component
  */
 export interface HeaderProps {
+  /**
```
ravinderkabli: Can we remove comments if not required

sumaztwcode: Raja has taken care @ravinderkabli

ravinderkabli (dismissed):
Looks fine

packages/bahmni-design-system/README.md
```diff
@@ -9,3 +9,43 @@ This library was generated with [Nx](https://nx.dev).
 ## Running unit tests
 
 Run `nx test @bahmni/design-system` to execute the unit tests via [Vitest](https://vitest.dev/)
+
```
sumaztwcode: @rajasurendra-tw Remove these; no need to add any comments in readme around this card

yarn.lock
```diff
@@ -11374,7 +11374,7 @@ json5@^2.1.2, json5@^2.2.0, json5@^2.2.2, json5@^2.2.3:
   resolved "https://registry.yarnpkg.com/json5/-/json5-2.2.3.tgz#78cd6f1a19bdc12b73db5ad0c61efd66c1e29283"
   integrity sha512-XmOWe7eyHYH14cLdVPoyg+GOH3rYX++KpzrylJwSW98t3Nk+U8XOl8FWKOgwtzdb8lXGf6zYwDUzeHMWfxasyg==
 
-jsonata@^2.2.1:
```
sumaztwcode: @rajasurendra-tw this may not be needed; pls do not check-in if this is not required

ravinderkabli (dismissed):

packages/bahmni-widgets/src/userGlobalAction/UserGlobalAction.tsx
```diff
@@ -1,32 +1,124 @@
...
+  }
+>;
+
 export const UserGlobalAction = () => {
   const { t } = useTranslation();
   const { userPrivileges } = useUserPrivilege();
+  // Consumed directly (not via the throwing `useNotification` hook) so the widget
+  // still works in an app that hasn't wired a `NotificationProvider` — it just
+  // falls back to logging instead of a toast.
+  const notification = useContext(NotificationContext);
```
sumaztwcode: cross check if this logic is required here; when I digged into this further - 

design is NOT correct as-is because:

  1. Provider should be mandatory. If Appointments, Registration, Clinical all have ActivePractitionerProvider (per the PR), Home should too — there's no good reason for the asymmetry.
  2. Errors should fail loudly. Silent failures (especially on action errors) are worse than crashes during config. A developer should discover immediately that they forgot to add NotificationProvider, not after users report "logout didn't work."
  3. The pattern should be consistent. All other widgets use safe hooks. This exception makes the codebase harder to understand.

packages/bahmni-widgets/src/userGlobalAction/UserGlobalAction.tsx
```diff
@@ -1,32 +1,124 @@
...
   const [isOpen, setIsOpen] = useState(false);
   const registry = useUserActionRegistry();
   const { getActions, version } = registry;
   const hasRegistered = useRef(false);
 
+  // Consumed directly (not via the throwing `useActivePractitioner` hook) so this
+  // widget keeps working — avatar-only, no greeting — in apps that haven't wired
+  // an `ActivePractitionerProvider` yet, instead of crashing the header.
+  const activePractitioner = useContext(ActivePractitionerContext);
```
sumaztwcode: cross check if this logic is required here
design is NOT correct as-is because:

  1. Provider should be mandatory. If Appointments, Registration, Clinical all have ActivePractitionerProvider (per the PR), Home should too — there's no good reason for the asymmetry.
  2. Errors should fail loudly. Silent failures (especially on action errors) are worse than crashes during config. A developer should discover immediately that they forgot to add NotificationProvider, not after users report "logout didn't work."
  3. The pattern should be consistent. All other widgets use safe hooks. This exception makes the codebase harder to understand.

sumaztwcode (commented):
## Review for BAH-4770

Good work unifying the header user menu across all four apps. Clean architectural layering, comprehensive test coverage (327 new test lines), and solid accessibility implementation. Two items need discussion before merge.

Found 0 blocking issues, 2 suggestions, and 2 nits.

**Acceptance Criteria**: 6/7 met (partial: long names — CSS truncation works but AC says "consider only first name", code uses full `user.display`)

### What's Done Well
- Clean separation: brand/layout slots in design-system, user menu widget in widgets, app-specific code removed from Home
- Graceful degradation via direct `useContext` with optional chaining — works without `ActivePractitionerProvider`
- Thorough accessibility: ARIA attributes, focus management, keyboard wrap-around, axe tests
- CSS specificity done right with targeted selectors and clear comments

### Info-level observations (not posted inline)

| File | Type | Issue |
|------|------|-------|
| `apps/*/src/App.tsx` | question | Provider nesting order (`ActivePractitionerProvider`, `UserPrivilegeProvider`, `UserActionProvider`) differs across apps — no functional bug today but worth standardizing |
| `UserGlobalAction.tsx:130` | question | `filteredActions` useMemo eslint-disable could be removed by adding stable `getActions` to deps |

packages/bahmni-widgets/src/userGlobalAction/UserGlobalAction.tsx
```diff
@@ -41,30 +133,48 @@ export const UserGlobalAction = () => {
...
-          id="user-icon"
-          data-testid="user-icon-button-test-id"
-          name="fa-user"
-          size={ICON_SIZE.LG}
-        />
-      </IconButton>
+        <UserAvatar id="user-icon" data-testid="user-icon-button-test-id" />
+        {user?.display && (
+          <span className={styles.greeting}>
+            {t('USER_GLOBAL_ACTION_GREETING', { name: user.display })}
```
sumaztwcode: suggestion: JIRA AC for "Long names handled" says _"fix the name width → consider only first name"_. This uses `user.display` (full name, e.g. "Dr. John Doe") with CSS truncation (`max-inline-size: 150px` + ellipsis). While truncation prevents layout breakage, the AC specifically asks for first-name-only rendering.

Consider extracting first name (e.g. `user.display.split(' ')[0]`) or confirming with the product team that CSS truncation of the full name is acceptable.

packages/bahmni-widgets/package.json
```diff
@@ -26,6 +26,8 @@
     "prepublishOnly": "yarn run build"
   },
   "dependencies": {
+    "@carbon/icons-react": "^11.65.0",
```
sumaztwcode: suggestion: `@carbon/icons-react` and `@carbon/react` are added as direct `dependencies`. The existing pattern in this package declares shared framework packages (`react`, `react-dom`, `@tanstack/react-query`, `@bahmni/design-system`, `@bahmni/services`) as `peerDependencies`. Carbon is the same category — a shared UI framework that consuming apps already bundle via `@bahmni/design-system`.

Hard `dependency` risks two independent copies of Carbon in the final bundle (duplicate CSS custom properties, broken theming context, increased bundle size). Move to `peerDependencies`.

apps/registration/src/pages/list/styles/index.module.scss
```diff
@@ -31,7 +31,7 @@
 .headerButton {
   position: fixed !important;
   top: $spacing-02;
-  right: $spacing-09;
+  right: calc(#{$spacing-12} + #{$spacing-08});
```
sumaztwcode: nit: This hardcoded right offset (`calc($spacing-12 + $spacing-08)`) will break if `UserGlobalAction` width changes (it's dynamic, up to 150px). Consider a layout-based approach (flexbox/grid spacing) rather than `position: fixed` with a magic-number offset.

packages/bahmni-widgets/src/userGlobalAction/UserGlobalAction.tsx
```diff
@@ -1,32 +1,124 @@
...
+      if (menu && items.length > 0) {
+        menuEl = menu;
+        items[0].focus();
+        menu.addEventListener('keydown', handleArrowKeys, true);
+      } else if (attempts < 5) {
+        attempts += 1;
+        frame = requestAnimationFrame(setup);
+      }
+    };
+    frame = requestAnimationFrame(setup);
```
sumaztwcode, line 122: nit: The rAF retry loop cleanup cancels the last `frame` ID, but if the user rapidly toggles the menu, a stale callback from a previous open cycle could fire after cleanup. Consider adding an `isCancelled` flag (set `true` in cleanup, checked at top of `setup()`) — the standard pattern for cancellable rAF loops.

```js
let cancelled = false;
const setup = () => {
  if (cancelled) return;
  // ... existing logic
};
return () => {
  cancelled = true;
  cancelAnimationFrame(frame);
  menuEl?.removeEventListener('keydown', handleArrowKeys, true);
};
```

packages/bahmni-design-system/README.md
rajasurendra-tw: Fixed — the extra migration-notes content after line 12 was removed in an earlier commit; current README.md only has the original 12 lines.

yarn.lock
rajasurendra-tw: Confirmed — checked yarn.lock, the stray jsonata diff is no longer present. Current entry is just `jsonata@2.2.1:` used by other deps, nothing extra checked in.

packages/bahmni-widgets/src/userGlobalAction/UserGlobalAction.tsx
rajasurendra-tw: Fixed — UserGlobalAction now uses the mandatory, throwing `useActivePractitioner()`/`useNotification()` hooks instead of optional `useContext` fallbacks. Verified all four apps (Home, Registration, Clinical, Appointments) wire both `ActivePractitionerProvider` and `NotificationProvider` at the App.tsx root, so the asymmetry this comment described no longer exists.

packages/bahmni-widgets/src/userGlobalAction/UserGlobalAction.tsx
rajasurendra-tw: Same fix as the sibling comment above — switched to the mandatory `useActivePractitioner()`/`useNotification()` hooks so missing providers fail loudly instead of silently degrading.

packages/bahmni-widgets/src/userGlobalAction/UserGlobalAction.tsx
rajasurendra-tw: Fixed — added a `getFirstName` helper (`display.split(' ')[0]`) and the greeting now renders only the first name, matching the JIRA AC. CSS truncation on `.greeting`/`.userAction` is left in place as a defense-in-depth safeguard.

packages/bahmni-widgets/package.json
rajasurendra-tw: Fixed — moved `@carbon/icons-react` and `@carbon/react` to `peerDependencies` in packages/bahmni-widgets/package.json, consistent with how other shared framework packages are declared.

apps/registration/src/pages/list/styles/index.module.scss
rajasurendra-tw: Fixed — moved the "Create new patient" button into Header's `globalFeatures` flex slot (the same layout-based pattern HomePageHeader already uses for LocationSelector) instead of `position: fixed` with a magic-number offset. It now flows naturally next to UserGlobalAction regardless of its width.

packages/bahmni-widgets/src/userGlobalAction/UserGlobalAction.tsx
rajasurendra-tw: Fixed — added a `cancelled` flag checked at the top of `setup()` and set in the cleanup function, so a stale retry callback from a previous open cycle can no longer fire after cleanup.

hamsavarthiniR-Bahmni (approved):

PR #563, opened 2026-07-21
BAH-4876 | Refactor. Wire Real Episode Search API

rahu1ramesh:
**JIRA** → [BAH-4876](https://bahmni.atlassian.net/browse/BAH-4876)

Replaces the local mock post function in CommonSearchWidget with the real post from @bahmni/services, wiring the widget to live API endpoints. Also fixes a bug where a failed search would navigate to the results view instead of returning to the search form.

## Key Features

- CommonSearchWidget now calls real REST endpoints using post from @bahmni/services
- On API failure, the widget correctly returns to the search form instead of rendering SearchSummary + ResultsTable with stale state

[BAH-4876]: https://bahmni.atlassian.net/browse/BAH-4876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/bahmni-widgets/src/search/commonSearch/constants.ts
```diff
@@ -0,0 +1,4 @@
+export const LOCATION_UUID_FIELD = 'location.uuid';
+export const LOCAL_ISO_DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSxx";
```
mohan-13, line 2: ISO is a standard format. Why have a custom constant defined. Why can't we leverage formatISO from date-fns package ? @rahu1ramesh

packages/bahmni-widgets/src/search/commonSearch/utils.ts
```diff
@@ -196,14 +209,41 @@ const buildCondition = ({ field, value }: ResolvedRow): SearchCondition => {
...
       ],
     };
   }
   return { field: field.key, comparator: 'eq', value: value.value };
 };
 
+const toLocalIso = (v: string): string =>
+  format(new Date(v), LOCAL_ISO_DATE_FORMAT);
+
+const localizeDateTime = (value: CriterionValue): CriterionValue => {
```
mohan-13, line 231: Are we supporting a date and time picker ? Why do we want to format all timestamps?
@rahu1ramesh

packages/bahmni-widgets/src/search/commonSearch/constants.ts
rahu1ramesh: On the backend to get started we have aligned with the Openmrs dateTime format for response (reusing the same for response). The frontend changes align with the backend date format. We can standardize this to plain ISO when we carve out the search to a library.

packages/bahmni-widgets/src/search/commonSearch/utils.ts
rahu1ramesh: We only support a date picker now. Since the frontend has limited knowledge on what the date is going to be used for instead of sending the date along, we are sending the whole date time here.

packages/bahmni-widgets/src/search/commonSearch/constants.ts
mohan-13: How is this format different from a ISO format ?

rahu1ramesh: The Java's Z pattern expects the timezone without a colon, but ISO from date-fns comes with one (+0530 and +05:30).

mohan-13: Okay, but lets further explore more on how Spring binds standard ISO format, so that we can remove such custom formats

mohan-13 (approved):

PR #567, opened 2026-07-22
Bah 4854 patient not found

AkshataKeshatwarTWer:
> [!NOTE]
> Thank you for your contribution. Please find the details of this pull request below.

**JIRA** → [BAH-#](https://bahmni.atlassian.net/browse/BAH-#)

### **Description**

 

---

### **Key Features**
 
 
 

---

### **Implementation Details**

* **UI Components**
   
   

* **State Management**
   

* **Custom Hooks**
   

* **API Integration**
   
   

* **Performance & Optimization**
   

* **Internationalisation**
   

* **Accessibility**
   

* **Limitations**
   

---

### **Testing & Type Definitions**

* **Unit Tests**
   

* **Integration Tests**
   

* **Type Definitions**
   
   

---

### **Screenshots**

| Feature / State | Screenshot |
|------------------|------------|
| Example: Empty State | Attach image or paste screenshot here |
| Example: Search Result | Attach image or paste screenshot here |
| Example: Validation Error | Attach image or paste screenshot here |

_Note: Please ensure sensitive information is not included in screenshots._

---

### **Next Steps**
 
 

---

> [!IMPORTANT]
> **Checklist**
>
> - [ ] Code adheres to project linting and formatting standards.
> - [ ] New components added to Storybook.
> - [ ] Tests written and passing (unit + integration).
> - [ ] Labels and text support i18n.
> - [ ] Follows accessibility and responsive design guidelines.
> - [ ] PR reviewed by at least one other developer.

---

### **Reviewer(s)**

@bahnew/developers 
_Kindly review the proposed changes when convenient. Your feedback is appreciated._

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Added clear loading feedback while patient data is being retrieved.
  * Prevented patient-specific dashboard content from appearing when the patient cannot be found.
  * Added a notification explaining when a patient record is unavailable.
  * Improved handling to distinguish missing patients from authorization and server errors.

* **Localization**
  * Added English and Spanish translations for patient loading and not-found messages.

* **Tests**
  * Expanded coverage for loading states, missing patients, and related error scenarios.

### DIFF
--- a/apps/clinical/src/pages/ConsultationPage.tsx
+++ b/apps/clinical/src/pages/ConsultationPage.tsx
@@ -53,6 +53,7 @@ import {
   getDefaultDashboard,
   getSidebarItems,
   filterSectionsByPrivileges,
+  isPatientNotFoundError,
 } from './util';
 
 const addSectionIds = (config: DashboardConfig): DashboardConfig => {
@@ -126,11 +127,27 @@ const ConsultationPage: React.FC = () => {
   const viewingForm = useObservationFormsStore((state) => state.viewingForm);
 
   const patientUUID = usePatientUUID();
-  const { data: patient } = useQuery({
+  const {
+    data: patient,
+    error: patientError,
+    isError: isPatientError,
+  } = useQuery({
     queryKey: ['patient', patientUUID],
     queryFn: () => getFormattedPatientById(patientUUID!),
     enabled: !!patientUUID,
   });
+
+  const isPatientNotFound =
+    isPatientError && isPatientNotFoundError(patientError);
+
+  // Only mount the patient-scoped content (patient header + dashboard widgets)
+  // once the patient is confirmed to exist. Those widgets each fetch data for
+  // the UUID as soon as they mount and raise their own error toasts, so
+  // rendering them for a missing patient would spam "Bad Request"/"Server Error"
+  // alongside the single "Patient not found" message. The header/breadcrumb/nav
+  // chrome always renders, so the screen is never blank. `patientUUID` is absent
+  // only outside the consultation route, where content should render as before.
+  const shouldRenderPatientContent = !patientUUID || !!patient;
 
   const breadcrumbItems = useMemo(
     () => [
@@ -198,6 +215,16 @@ const ConsultationPage: React.FC = () => {
       });
     }
   }, [dashboardConfigError]);
+
+  useEffect(() => {
+    if (isPatientNotFound && patientError instanceof Error) {
+      addNotification({
+        title: t('ERROR_DEFAULT_TITLE'),
+        message: t(patientError.message),
+        type: 'error',
+      });
+    }
+  }, [isPatientNotFound]);
 
   const filteredDashboardConfig = useMemo(() => {
     if (!dashboardConfig || !userPrivileges) return null;
@@ -326,24 +353,28 @@ const ConsultationPage: React.FC = () => {
               />
             }
           >
-            <div
-              id="section-sticky-header"
-              data-testid="section-sticky-header-test-id"
-              role="region"
-              aria-label={t('PATIENT_HEADER_SECTION')}
-              className={styles.stickySection}
-            >
-              <PatientHeader
-                isActionAreaVisible={isActionAreaVisible}
-                printOptions={filteredPrintOptions}
-              />
-              {renderContextInformation()}
-            </div>
-            <DashboardContainer
-              sections={filteredDashboardConfig!.sections}
-              activeItemId={activeItemId}
-              scrollTrigger={scrollTrigger}
-            />
+            {shouldRenderPatientContent && (
+              <>
+                <div
+                  id="section-sticky-header"
+                  data-testid="section-sticky-header-test-id"
+                  role="region"
+                  aria-label={t('PATIENT_HEADER_SECTION')}
+                  className={styles.stickySection}
+                >
+                  <PatientHeader
+                    isActionAreaVisible={isActionAreaVisible}
+                    printOptions={filteredPrintOptions}
+                  />
+                  {renderContextInformation()}
+                </div>
+                <DashboardContainer
+                  sections={filteredDashboardConfig!.sections}
+                  activeItemId={activeItemId}
+                  scrollTrigger={scrollTrigger}
+                />
+              </>
+            )}
           </Suspense>
         }
         isActionAreaVisible={isActionAreaVisible}
--- a/apps/clinical/src/pages/ConsultationPage.tsx
+++ b/apps/clinical/src/pages/ConsultationPage.tsx
@@ -14,6 +14,7 @@ import {
   getFormattedPatientById,
   capitalize,
   hasPrivilege,
+  PATIENT_NOT_FOUND_ERROR_KEY,
 } from '@bahmni/services';
 import {
   ProgramDetails,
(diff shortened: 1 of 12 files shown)


apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -126,12 +127,29 @@ const ConsultationPage: React.FC = () => {
...
+    isPatientError && isPatientNotFoundError(patientError);
+
+  // Only mount the patient-scoped content (patient header + dashboard widgets)
+  // once the patient is confirmed to exist. Those widgets each fetch data for
+  // the UUID as soon as they mount and raise their own error toasts, so
+  // rendering them for a missing patient would spam "Bad Request"/"Server Error"
+  // alongside the single "Patient not found" message. The header/breadcrumb/nav
+  // chrome always renders, so the screen is never blank. `patientUUID` is absent
+  // only outside the consultation route, where content should render as before.
+  const shouldRenderPatientContent = !patientUUID || !!patient;
```
ravinderkabli, line 164: **Gap: invalid UUID format gives a silent blank content area**

When the UUID in the URL is not a valid UUID format (e.g. `abc-not-a-uuid`), `getPatientById` throws a JS validation error _before_ making any network call:

```ts
throw new Error(`Invalid patient UUID format: ${patientUUID}`);
```

This error is not an Axios error, so `getFormattedError` never runs and the message is never `ERROR_PATIENT_NOT_FOUND`. As a result:
- `isPatientNotFoundError(patientError)` → `false`
- `isPatientNotFound` → `false`
- The `useEffect` notification never fires
- But `shouldRenderPatientContent = !patientUUID || !!patient` is also `false` (patient is undefined)

The user sees: **no notification, no dashboard, no explanation** — a blank content area.

packages/bahmni-services/src/errorHandling/utils.ts
```diff
@@ -41,6 +41,20 @@ const parseFHIRError = (outcome: FHIROperationOutcome): string | null => {
...
+export const PATIENT_NOT_FOUND_ERROR_KEY = 'ERROR_PATIENT_NOT_FOUND';
+
+// A GET of a single FHIR patient resource, e.g. /ws/fhir2/R4/Patient/{uuid}.
+// A 400 (UUID looks valid but does not exist) or 404 (truly missing) on this
+// request means the patient could not be found, so we surface a dedicated
+// "patient not found" message (as a translation key) instead of a generic one.
+// 5xx is intentionally excluded — that indicates a real server error, not a
+// missing patient, and must not be masked.
+const PATIENT_RESOURCE_REQUEST = /\/Patient\/[0-9a-f-]{36}/i;
+const PATIENT_NOT_FOUND_STATUSES = [400, 404];
```
ravinderkabli, line 56: **Note: regex intentionally excludes non-UUID-format strings**

The regex `/\/Patient\/[0-9a-f-]{36}/i` only matches URLs where the patient segment looks like a UUID. This is correct for the HTTP error case (a valid-format but non-existent UUID reaches the API and gets a 400/404).

However, combined with the UUID format validation in `getPatientById` that throws _before_ the API call for invalid-format strings, there are two separate error paths for "bad UUID in URL":
1. Invalid format (e.g. `abc`) → JS validation error, never hits this interceptor
2. Valid format but non-existent (e.g. `00000000-0000-0000-0000-000000000000`) → API returns 400/404, this interceptor catches it ✅

Path 1 has no user-facing feedback currently (see comment on `ConsultationPage.tsx`). Worth aligning both paths to the same outcome.

ravinderkabli (commented):
## Code Review

packages/bahmni-services/src/errorHandling/utils.ts
AkshataKeshatwarTWer: Agreed with the analysis, and both paths now land on the same outcome as you suggested.

- **Path 1** (invalid format, e.g. `abc`) — `getPatientById` throws `PATIENT_NOT_FOUND_ERROR_KEY` directly, before the request is made.
- **Path 2** (valid format, non-existent) — unchanged; the 400/404 interceptor classifies it here.

Both surface `ERROR_PATIENT_NOT_FOUND`, so the regex staying UUID-shaped is now correct by design rather than a gap — it only ever has to handle path 2. Path 1 never reaches this code, and no longer needs to.

While in here I also tightened the matcher per the other review comment, so it no longer catches sub-resources or non-fetch calls on the same path:

```ts
// packages/bahmni-services/src/errorHandling/utils.ts:55
const PATIENT_RESOURCE_REQUEST = /\/Patient\/[0-9a-f-]{36}(?:[?#]|$)/i;
```

plus a GET check in `isPatientResourceFetch`, so `/Patient/{uuid}/$photo` and a failed create/update on the same URL keep their own errors instead of being mislabelled as a missing patient.

`errorHandling/__tests__/utils.test.ts` asserts all of these: sub-resource 404, patient update 400, 500, 401 and non-patient 400 are each verified *not* to classify as patient-not-found, alongside a positive case for an explicit GET.

apps/clinical/src/pages/ConsultationPage.tsx
AkshataKeshatwarTWer: "Gap: invalid UUID format gives a silent blank content area"

Good catch — this is fixed. The root cause was exactly what you identified: the two "bad UUID" paths ended in different places, and only one of them produced a key the UI could act on.

`getPatientById` now rejects malformed input with the *same* key the 400/404 response path produces, before any network call happens:

```ts
// packages/bahmni-services/src/patientService/patientService.ts:55-61
if (
  !patientUUID ||
  patientUUID.trim() === '' ||
  !UUID_PATTERN.test(patientUUID)
) {
  throw new Error(PATIENT_NOT_FOUND_ERROR_KEY);
}
```

So `isPatientNotFoundError(patientError)` → `true` and the notification fires. `getFormattedError` never running on this path is no longer a problem, because the key is produced at the source rather than derived from an Axios error shape.

Separately, the notification no longer depends on the error being a not-found at all — every fetch failure now produces feedback, with a generic key as the fallback:

```ts
// apps/clinical/src/pages/ConsultationPage.tsx:149-154
const patientErrorMessageKey = useMemo(() => {
  if (!isPatientError) return null;
  return isPatientNotFoundError(patientError)
    ? PATIENT_NOT_FOUND_ERROR_KEY
    : 'ERROR_FETCHING_PATIENT_DATA';
}, [isPatientError, patientError]);
```

On your "no explanation" point specifically — a toast dismisses itself, so the content area now states the reason directly instead of relying on the notification alone:

```tsx
// apps/clinical/src/pages/ConsultationPage.tsx:377-385
{patientErrorMessageKey && (
  <div
    id="error-loading-patient"
    data-testid="error-loading-patient-test-id"
    role="alert"
  >
    {t(patientErrorMessageKey)}
  </div>
)}
```

The patient-scoped widgets stay held back (`shouldRenderPatientContent`, line 164) — mounting them for a missing patient makes each one fire its own "Bad Request"/"Server Error" toast on top of the real message. So the outcome is: header/breadcrumb/nav chrome, one explanatory message in the content area, and one notification.

Tests covering this:
- `patientService.test.ts` — empty, whitespace-only, invalid-format and invalid-character UUIDs all reject with the not-found key
- `ConsultationPage.test.tsx` — "shows the patient-not-found error notification when the URL holds a malformed UUID", plus "renders the reason in the content area when the patient is not found", "renders the generic reason in the content area for a non-not-found error", and "does not render the patient error state when the patient loads"

ravinderkabli (approved):

PR #568, opened 2026-07-23
BAH-4887 | Encounter reference added for conditions

sahupadma:
https://bahmni.atlassian.net/browse/BAH-4887

* **New Features**
  * Conditions marked inactive now use the active encounter session context when available, including whether the encounter was matched.
  * Inactivation updates now use an encounter-aware transaction workflow, reusing a matched encounter when possible or creating a finished ambulatory encounter when not.
* **Bug Fixes**
  * Inactive updates now consistently set both inactive clinical status and the appropriate condition category.
  * When there isn’t enough encounter context to proceed, the operation now fails instead of performing a standalone condition update.

## Summary by CodeRabbit

* **New Features**
  * Marking a condition inactive now uses the active encounter session context (including whether it was matched) when available.
  * The flow updates encounter/session state after success, and refreshes consultation data when patient details are present.
* **Bug Fixes**
  * Inactivation consistently sets `clinicalStatus` to inactive and applies the correct problem-list category.
  * If insufficient encounter context is available, the action now fails rather than falling back to a Condition-only update.
* **Tests**
  * Expanded unit/integration coverage for the encounter-aware inactivation flow and widget prop wiring.


ravinderkabli (commented):
## Review for BAH-4887

packages/bahmni-services/src/conditionService/conditionService.ts
```diff
@@ -67,17 +65,83 @@ export async function getConditionPage(
...
+      : undefined,
+    location: [
+      {
+        location: {
+          reference: `Location/${locationUuid}`,
+          type: 'Location',
+        },
+      },
+    ],
+    period: { start: new Date().toISOString() },
```
ravinderkabli: blocking: On the matched-reuse path, `buildConditionEncounter` always sets `period.start = now` (here), plus `location` = current login location and `participant` = only the passed practitioner. This encounter is then PUT onto the existing `activeEncounter.id`, and an OpenMRS FHIR PUT is a full-resource replace — so the original encounter's start time is rewritten on every AC1 reuse, and its location/participants are overwritten too.

The established pattern deliberately preserves these: `submitConsultation` calls `createEncounterResource(..., deps.activeEncounter?.period?.start ?? null)`. Please pass `activeEncounter.period?.start` through and preserve `activeEncounter.location`/`.participant` on the reuse branch, defaulting to now/login-location only on the genuine create paths.

packages/bahmni-services/src/conditionService/conditionService.ts
```diff
@@ -89,8 +153,114 @@ export async function markConditionAsInactive(
...
+    const newEncounter = buildConditionEncounter(
+      newEncounterType,
+      newEncounterPartOf,
+      newEncounterSubject,
+      practitionerUUID,
+    );
+    // Create the encounter first so we have its server-assigned UUID before
+    // building the bundle. OpenMRS does not reliably return entry.response.location
+    // in transaction-response bundles, so a standalone POST is used to obtain the UUID.
+    const createdEncounter = await createFhirEncounter(newEncounter);
```
ravinderkabli, line 222: suggestion: This standalone `createFhirEncounter` POST happens *outside* the bundle transaction. If the following bundle POST fails, the encounter is already persisted with no condition linked and no rollback — an orphan encounter. That breaks AC4's atomicity guarantee for the create path (the matched path is atomic).

Either create the encounter inside the transaction bundle (POST + `urn:uuid` bundle-local ref — the existing `createEncounterBundleEntry` handles this), or add best-effort compensating cleanup on bundle failure and document the residual non-atomicity.

packages/bahmni-widgets/src/conditions/ConditionsTable.tsx
```diff
@@ -90,13 +97,35 @@ const ConditionsTable: React.FC<WidgetProps> = ({
...
+      const encounter = await markConditionAsInactive(
+        conditionToMarkInactive.rawFhirResource,
+        activeEncounter ?? undefined,
+        activeEncounterMatched ?? false,
+        encounterTypeName,
+        patientUUID ?? undefined,
+        practitioner?.uuid,
+      );
+      if (encounter.id) {
+        setEncounterSessionDecision({ reasons: ['MATCHED'], encounter });
```
ravinderkabli, line 109: suggestion: This forces `reasons: ['MATCHED']` into the session store after *any* successful mark-inactive — including the `matched=false` create paths and cases where the original session was PROVIDER_MISMATCH/LOCATION_MISMATCH. Combined with the new `useEncounterSession` seed, a mismatched or freshly-created encounter then gets inherited by the consultation pad as a clean MATCHED resume, and `canEditOrCreate` treats it as a full match.

Consider writing the reason that reflects the actual outcome rather than hardcoding `['MATCHED']`. Also note `encounterSessionStore` documents a single-writer convention (PatientHeader writes, widgets read) — worth updating that doc for this new writer.

apps/clinical/src/hooks/useEncounterSession.ts
```diff
@@ -61,6 +62,24 @@ export function useEncounterSession(
       // location cookie unavailable — location check will be skipped in mapper
     }
 
+    // Seed immediately from the global store when it already reflects a MATCHED
+    // session. This lets the consultation pad inherit an encounter created by a
+    // dashboard widget action (e.g. mark-condition-inactive) without waiting for
+    // the OpenMRS FHIR search index to catch up to the freshly created encounter.
+    const storeState = getEncounterSessionSnapshot();
+    if (
+      storeState.matchReasons.includes('MATCHED') &&
```
ravinderkabli, line 71: suggestion: This seed adopts the store snapshot's encounter for the current `patientUUID` without checking it actually belongs to this patient. Safety relies entirely on PatientHeader's reset effect firing before this hook's effect on a patient switch — both are keyed on `patientUUID`, so ordering isn't guaranteed. A stale cross-patient MATCHED snapshot could be seeded onto the new patient.

Since the encounter carries `subject`, a cheap guard removes the ordering dependency, e.g. also require `storeState.activeEncounter.subject?.reference?.endsWith(patientUUID)`.

packages/bahmni-services/src/conditionService/conditionService.ts
```diff
@@ -67,17 +65,83 @@ export async function getConditionPage(
   };
 }
 
-/**
- * Marks a condition as inactive via FHIR PUT.
- * Preserves all existing fields from the raw resource; only clinicalStatus is changed.
- * @param condition - The full raw FHIR Condition resource to update
- * @returns Promise resolving to the updated Condition resource
- */
+function buildConditionEncounter(
```
ravinderkabli, line 67: suggestion: `buildConditionEncounter` re-hand-rolls the encounter shape already produced by `createEncounterResource` (apps/clinical/src/utils/fhir/encounterResourceCreator.ts), and this function now carries the full reuse-vs-create orchestration that in this codebase lives in the app layer (`submitConsultation`). It's partly forced (that builder is in `apps/clinical` and can't be imported by `packages`), but it's a second source of truth and is the root of the reuse-semantics divergence in the blocking comment.

Consider moving the orchestration to an app-layer service that reuses `createEncounterResource`/`createEncounterBundleEntry`, or extracting one shared builder into `packages/bahmni-services`.

packages/bahmni-widgets/src/conditions/ConditionsTable.tsx
```diff
@@ -90,13 +97,35 @@ const ConditionsTable: React.FC<WidgetProps> = ({
...
+      }
+      if (patientUUID) {
+        dispatchAuditEvent({
+          eventType: AUDIT_LOG_EVENT_DETAILS.EDIT_ENCOUNTER
+            .eventType as AuditEventType,
+          patientUuid: patientUUID,
+          messageParams: { conditionDisplay: conditionToMarkInactive.display },
```
ravinderkabli: nit: AC5 asks for the audit entry to carry encounter context, but only `conditionDisplay` is passed. The returned `encounter.id` is already in scope (used just above for `setEncounterSessionDecision`) and the dispatcher supports `encounterUuid`/`encounterType`. Consider adding `encounterUuid: encounter.id` to `messageParams`.

packages/bahmni-services/src/conditionService/conditionService.ts
```diff
@@ -67,17 +65,83 @@ export async function getConditionPage(
...
+          reference: `Location/${locationUuid}`,
+          type: 'Location',
+        },
+      },
+    ],
+    period: { start: new Date().toISOString() },
+  };
+}
+
 export async function markConditionAsInactive(
```
ravinderkabli, line 96: nit: This now returns `Promise<Encounter>` (not `Condition`) and takes 6 args, but the JSDoc was removed and not replaced. A short JSDoc documenting the argument matrix and the returned encounter would help call sites; the name `markConditionAsInactive` also no longer reflects the return value.

packages/bahmni-services/src/conditionService/conditionService.ts
```diff
@@ -89,8 +153,114 @@ export async function markConditionAsInactive(
...
+    ];
+    await post<Bundle>(ENCOUNTER_BUNDLE_URL, createEncounterBundle(entries));
+    return freshEncounter;
+  }
+
+  let newEncounterType: Encounter['type'];
+  let newEncounterPartOf: Encounter['partOf'];
+  let newEncounterSubject: Encounter['subject'];
+
+  if (!matched && activeEncounter) {
```
ravinderkabli, line 180: nit: When `matched=false` but a mismatched `activeEncounter` is present, this never falls through to the `encounterTypeName`/`patientUuid` resolution. If the mismatched encounter lacks `type`/`partOf`, the whole call throws even though a valid config-driven fallback was supplied. The comment says this is intentional — flagging as a known edge case in case a fallback is preferred.

apps/clinical/src/hooks/useEncounterSession.ts
```diff
@@ -61,6 +62,24 @@ export function useEncounterSession(
...
+      storeState.activeEncounter?.id &&
+      !signal.ignored
+    ) {
+      setHasActiveSession(true);
+      setActiveEncounter(storeState.activeEncounter);
+      setIsPractitionerMatch(true);
+      setMatchReason(['MATCHED']);
```
ravinderkabli: nit: `setMatchReason(['MATCHED'])` discards any other codes in the store snapshot's `matchReasons`. `fetchSessionState` preserves `decision.reasons` verbatim — consider `setMatchReason(storeState.matchReasons)` for consistency.

packages/bahmni-services/src/conditionService/conditionService.ts
```diff
@@ -67,17 +65,83 @@ export async function getConditionPage(
...
+  encounterTypeName?: string,
+  patientUuid?: string,
+  practitionerUUID?: string,
+): Promise<Encounter> {
+  // Category is intentionally always set to problem-list-item. Conditions managed via
+  // this widget are problem-list conditions by definition, regardless of their original
+  // stored category (e.g. encounter-diagnosis).
+  const updatedCondition: Condition = {
     ...condition,
+    category: [
```
ravinderkabli, line 109: question: `category` is unconditionally rewritten to `problem-list-item` on every mark-inactive, which would silently reclassify an `encounter-diagnosis` condition. The comment says this is intentional — just confirming with product that no encounter-diagnosis condition ever reaches this action, since the change is silent and irreversible.

sahupadma: The conditions table queries with category=problem-list-item at the FHIR layer (see PATIENT_CONDITION_PAGE_URL), so only problem-list conditions are ever returned to this widget — an encounter-diagnosis condition cannot reach this action through normal usage. The rewrite is a defensive normalisation for conditions that may have been historically stored under a different category. The existing inline comment captures this rationale.

apps/clinical/src/hooks/useEncounterSession.ts
sahupadma: Fixed.

packages/bahmni-services/src/conditionService/conditionService.ts
sahupadma: Confirmed intentional. Any activeEncounter in the session store was persisted by the server, so it will always carry type and partOf. A missing field there indicates corrupt session state, not a normal mismatch scenario, and failing loudly is preferable to silently falling back to the config-driven lookup and creating an unintended encounter under the wrong type. If a graceful fallback is ever needed (e.g. a degraded-mode story), it can be added explicitly to the if (!matched && activeEncounter) branch.

packages/bahmni-services/src/conditionService/conditionService.ts
sahupadma: added

packages/bahmni-widgets/src/conditions/ConditionsTable.tsx
sahupadma: Fixed.

packages/bahmni-services/src/conditionService/conditionService.ts
sahupadma: Fixed.

apps/clinical/src/hooks/useEncounterSession.ts
sahupadma: Fixed.

packages/bahmni-widgets/src/conditions/ConditionsTable.tsx
sahupadma: Intentional. Every encounter returned by markConditionAsInactive is either the reused MATCHED encounter (AC1) or one freshly created for the current session in this patient's context (AC2 / NO_ACTIVE_ENCOUNTER). In both create cases, seeding the store as ['MATCHED'] is correct — it tells the consultation pad "this encounter is yours, resume it" which is exactly what should happen. Hardcoding avoids forwarding a stale reason code (e.g. PROVIDER_MISMATCH from the old encounter that was just replaced by a new one).

packages/bahmni-services/src/conditionService/conditionService.ts
sahupadma: documented

packages/bahmni-services/src/conditionService/conditionService.ts
sahupadma: Fixed.

packages/bahmni-services/src/encounterService/encounterService.ts
```diff
@@ -1,12 +1,80 @@
...
+      {
+        location: {
+          reference: `Location/${locationUuid}`,
+          type: 'Location' as const,
+        },
+      },
+    ],
+    period: { start: periodStart },
+  };
+}
```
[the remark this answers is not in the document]
sahupadma: The suggestion to pass 'finished' status in buildConditionEncounter directly contradicts the core BAH-4887 design decision. Let me show the exact evidence:

The encounter built by markConditionAsInactive is seeded into the session store via setEncounterSessionDecision({ reasons: ['MATCHED'], encounter }) in ConditionsTable.tsx. That encounter is then inherited by the consultation pad's useEncounterSession (via the snapshot seeding we added). The consultation pad uses it as activeEncounter to PUT data into on "Continue Consultation".

The searchEncounters call in findActiveEncounterInSession has no status filter — it queries by patient, _tag, _lastUpdated, and type. So a finished encounter would be returned. But it would then be PUT with clinical data attached to a closed encounter, which is invalid in FHIR semantics. More critically, the memory from BAH-4887 implementation records:

▎ encounter status = in-progress — finished caused consultation pad to create a new encounter on "Continue Consultation" instead of reusing.

This was a real regression that was debugged and fixed by keeping in-progress. Adding the status param to the interface and using 'finished' in buildConditionEncounter would re-introduce that break.

Making status configurable in BuildEncounterResourceParams would also be premature — no current caller needs anything other than 'in-progress', and the default-and-override pattern adds surface area with no benefit. The hard-coded status is correct by design, not an oversight.

ravinderkabli (dismissed):
Looks fine

ravinderkabli (approved):
Done

PR #570, opened 2026-07-23
BAH-4888 | Add. Configurable Result Field Transformer For Common Search Widget

skjameela276:
BAH-4888 | Add configurable result-field transforms to Common Search widget

What

Adds a config-driven transform option on Common Search result fields, so a column's raw value can be formatted before display.

Changes

- utils.ts — new resultTransforms map (keyed by the config transform string) with a Translator-aware ResultTransform type:
  - formatDate — date only
  - formatTime — time only (h:mm a)
  - formatDateTime — date + time
  - formatAge — birthdate → human-readable age
  - formatGender — M/F → localized label via t('GENDER_<value>'), falls back to raw value
  - formatCountry — ISO code → country name via Intl.DisplayNames (locale-aware), falls back to raw code
- ResultsTable.tsx — evaluateRows looks up each field's transform in the map and applies it (with t) after jsonata evaluation; unknown keys and null values pass through unchanged.
- models.ts / schema.json — added optional transform on ResultFieldConfig.
- api.ts — appointment time mock values changed from bare '08:00' to full ISO date-times so time transforms can parse them.
- Locale files — added GENDER_M/GENDER_F (en + es).

## Summary by CodeRabbit

* **New Features**
  * Added configurable result-field transformations for common search results (date, time, datetime, age, gender, country, and search-result formatting).
  * Added localized gender labels (English/Spanish) for gender display.
* **Bug Fixes**
  * Improved transformed-value rendering to apply translations when available and display `-` for null/empty results.
* **Tests**
  * Expanded coverage for known/unknown transforms and null/empty transform outputs, plus gender/country/search-result formatting behavior.
* **Documentation**
  * Updated the common search result-field configuration schema to support the new optional `transform` option.


apps/clinical/public/locales/locale_en.json
```diff
@@ -91,6 +91,7 @@
   "COMMON_SEARCH_RESULT_FIELD_PATIENT_AGE": "Age",
   "COMMON_SEARCH_RESULT_FIELD_PATIENT_APPLICANT_ID": "Applicant ID",
   "COMMON_SEARCH_RESULT_FIELD_PATIENT_BIRTHDATE": "Birthdate",
+  "COMMON_SEARCH_RESULT_FIELD_PATIENT_COUNTRY" : "Country",
```
rahu1ramesh: We can remove the result field translations from the locale. These should ideally come from the config. Not worth adding.

packages/bahmni-widgets/src/search/commonSearch/__tests__/ResultsTable.test.tsx
```diff
@@ -144,4 +146,48 @@ describe('ResultsTable', () => {
...
+      renderTable({
+        resultFields: mockResultFieldsWithTransform,
+        results: [{ id: '1', country: 'us' }],
+      });
+      await waitFor(() => {
+        expect(screen.getByText('United States')).toBeInTheDocument();
+      });
+    });
+
+    it('falls back to the raw value when the transform key is not registered', async () => {
```
rahu1ramesh, line 154: The ideal flow should be the other way around. We start by appending a constant key to the value (eg US -> `COUNTRY_CODE_${camelToScreamingSnakeCase(context)}`. Then if i18n returns the same key back (no translation available scenario), then we fallback to Intl package. If the look up also fails, we fall back to the value as is.

packages/bahmni-widgets/src/search/commonSearch/__tests__/ResultsTable.test.tsx
```diff
@@ -144,4 +146,48 @@ describe('ResultsTable', () => {
...
+      mockJsonata.mockReturnValue({
+        evaluate: jest.fn().mockResolvedValue('us'),
+      });
+      renderTable({
+        resultFields: mockResultFieldsWithUnknownTransform,
+        results: [{ id: '1', country: 'us' }],
+      });
+      await waitFor(() => {
+        expect(screen.getByText('us')).toBeInTheDocument();
```
rahu1ramesh: This assertion is also brittle, "us" is a very common word. See if we can improve this.

packages/bahmni-widgets/src/search/commonSearch/utils.ts
rahu1ramesh, line 1: Please move all the transform functions to bahmni-services. These are reusable and should be kept there for other consumers.

packages/bahmni-widgets/src/search/commonSearch/utils.ts
```diff
@@ -12,6 +17,47 @@ import {
...
+  options?: { count?: number; defaultValue?: string },
+) => string;
+
+export type ResultTransform = (value: unknown, t: Translator) => string;
+
+export type DateTimeValue = string | Date | number;
+
+export const formatGender = (value: unknown, t: Translator): string =>
+  t(`GENDER_${value ?? ''}`, {
+    defaultValue: typeof value === 'object' ? '' : String(value ?? ''),
```
rahu1ramesh: This seems a little confusing to me. Do we need to have the defaultValue here?

packages/bahmni-widgets/src/search/commonSearch/utils.ts
```diff
@@ -12,6 +17,47 @@ import {
...
+  });
+
+export const formatCountry = (value: unknown): string => {
```
rahu1ramesh: The ideal flow should be the other way around. We start by appending a constant key to the value (eg US -> COUNTRY_CODE_${camelToScreamingSnakeCase(context)}. Then if i18n returns the same key back (no translation available scenario), then we fallback to Intl package. If the look up also fails, we fall back to the value as is.

packages/bahmni-widgets/src/search/commonSearch/schema.json
```diff
@@ -91,6 +91,11 @@
           "type": "string",
           "enum": ["text", "select", "dateRange", "numeric"],
           "description": "When present, a filter control appears for this column"
+        },
+        "transform": {
+          "type": "string",
```
rahu1ramesh, line 96: Better to support an enum here (eg line 92).

packages/bahmni-widgets/src/search/commonSearch/ResultsTable.tsx
```diff
@@ -32,8 +35,9 @@ const evaluateRows = async (
             ? String((item as Record<string, unknown>).id)
             : generateUUID(),
       };
-      for (const { key, expr } of compiled) {
-        row[key] = await expr.evaluate(item as Record<string, unknown>);
+      for (const { key, expr, transform } of compiled) {
+        const value = await expr.evaluate(item as Record<string, unknown>);
+        row[key] = transform && value != null ? transform(value, t) : value;
```
rahu1ramesh: If value is falsy, there should be some default like "-" that we provide.

rahu1ramesh (changes requested):

packages/bahmni-services/src/utils/__tests__/utils.test.ts
```diff
@@ -1330,3 +1333,78 @@ describe('common utility functions', () => {
...
+  const t: Translator = (key) => translations[key] ?? key;
+
+  it.each([
+    { label: 'maps M to the male label', value: 'M', expected: 'Male' },
+    { label: 'maps F to the female label', value: 'F', expected: 'Female' },
+  ])('$label', ({ value, expected }) => {
+    expect(formatGender(value, t)).toBe(expected);
+  });
+
+  it('builds the translation key from the value', () => {
```
rahu1ramesh: The above test is enough, this one is a duplicate. Maybe add a test to see what happens in case of a value that is not supported ("P", "R" etc).

packages/bahmni-services/src/utils/__tests__/utils.test.ts
```diff
@@ -1330,3 +1333,78 @@ describe('common utility functions', () => {
...
+
+describe('formatCountry', () => {
+  const noTranslations: Translator = (key) => key;
+  const withTranslations =
+    (map: Record<string, string>): Translator =>
+    (key) =>
+      map[key] ?? key;
+
+  it.each([
+    {
```
rahu1ramesh, line 1373: It should take the i18n key first, then fall back on i18n

packages/bahmni-services/src/utils/__tests__/utils.test.ts
```diff
@@ -1330,3 +1333,78 @@ describe('common utility functions', () => {
...
+    },
+    {
+      label: 'falls back to Intl for an uppercase ISO code',
+      value: 'IN',
+      expected: 'India',
+    },
+    {
+      label: 'returns empty string for empty input',
+      value: '',
+      expected: '',
```
rahu1ramesh: For an empty / falsy value return null, let the consumer decide whether to use a "" or a "-".

packages/bahmni-services/src/utils/__tests__/utils.test.ts
```diff
@@ -1330,3 +1333,78 @@ describe('common utility functions', () => {
...
+    {
+      label: 'returns the raw code when Intl does not recognise the region',
+      value: '1',
+      expected: '1',
+    },
+  ])('$label', ({ value, expected }) => {
+    expect(formatCountry(value, noTranslations)).toBe(expected);
+  });
+
+  it('uses the i18n translation when the key is resolved', () => {
```
rahu1ramesh: Can we club this with the above it.each.

packages/bahmni-services/src/utils/utils.ts
```diff
@@ -402,3 +402,27 @@ export function convertToSentenceCase(str: string): string {
     .toLowerCase()
     .replace(/^./, (char) => char.toUpperCase());
 }
+
+export type Translator = (
```
rahu1ramesh: Do we have to export this type?

packages/bahmni-services/src/utils/utils.ts
```diff
@@ -402,3 +402,27 @@ export function convertToSentenceCase(str: string): string {
...
+  key: string,
+  options?: { count?: number; defaultValue?: string },
+) => string;
+
+export const formatGender = (value: unknown, t: Translator): string =>
```
rahu1ramesh: Have an early return here, if value.trim() is falsy, return null. Else t(`GENDER_${value)}`); -> t(`GENDER_${camelToScreamingSnakeCase(value) ?? ''}`);

packages/bahmni-services/src/utils/utils.ts
```diff
@@ -402,3 +402,27 @@ export function convertToSentenceCase(str: string): string {
...
+  t(`GENDER_${value ?? ''}`);
+
+export const formatCountry = (value: unknown, t: Translator): string => {
+  const code = (typeof value === 'object' ? '' : String(value ?? '')).trim();
```
rahu1ramesh: I don't think we should support unknown, keep value as string. This is a very brittle contract.

packages/bahmni-services/src/index.ts
```diff
@@ -113,6 +113,9 @@ export {
   camelToScreamingSnakeCase,
   convertToSentenceCase,
   resolveComboBoxItems,
+  type Translator,
```
rahu1ramesh: No need to export this.

packages/bahmni-widgets/src/search/commonSearch/__tests__/ResultsTable.test.tsx
```diff
@@ -144,4 +149,78 @@ describe('ResultsTable', () => {
       expect(await axe(container)).toHaveNoViolations();
     });
   });
+
+  describe('Result field transforms', () => {
```
rahu1ramesh, line 153: Do we need to assert these again. We already have unit test to cover this at the service layer.

packages/bahmni-widgets/src/search/commonSearch/__tests__/utils.test.ts
rahu1ramesh, line 1: Try to use it.each for tests, less code duplication.

packages/bahmni-widgets/src/search/commonSearch/ResultsTable.tsx
```diff
@@ -32,8 +39,14 @@ const evaluateRows = async (
             ? String((item as Record<string, unknown>).id)
             : generateUUID(),
       };
-      for (const { key, expr } of compiled) {
-        row[key] = await expr.evaluate(item as Record<string, unknown>);
+      for (const { key, expr, transform } of compiled) {
+        const value = await expr.evaluate(item as Record<string, unknown>);
+        row[key] =
+          value != null && value !== ''
```
rahu1ramesh: value != null && value !== '' are both falsy checks, can be replaced with 

```
const value = await expr.evaluate(item as Record<string, unknown>);
if(!value) {
 row[key] = "-"; 
 continue; 
}
row[key] = transform ? transform(value, t) : value;
```

rahu1ramesh (changes requested):

apps/clinical/public/locales/locale_en.json
kousalyapt-tw: done

packages/bahmni-widgets/src/search/commonSearch/__tests__/ResultsTable.test.tsx
kousalyapt-tw: done

packages/bahmni-widgets/src/search/commonSearch/utils.ts
kousalyapt-tw: done

packages/bahmni-widgets/src/search/commonSearch/utils.ts
kousalyapt-tw: removed  the defaultValue.

packages/bahmni-widgets/src/search/commonSearch/utils.ts
kousalyapt-tw: done

packages/bahmni-widgets/src/search/commonSearch/ResultsTable.tsx
kousalyapt-tw: done

packages/bahmni-services/src/utils/utils.ts
kousalyapt-tw: Translator is used in bahmni-widgets — ResultsTable.tsx, utils.ts, and utils.test.ts all import it from @bahmni/services. It needs to be exported for those consumers.

packages/bahmni-services/src/index.ts
kousalyapt-tw: Translator is consumed in bahmni-widgets — ResultsTable.tsx, utils.ts, and utils.test.ts all import it directly from @bahmni/services. Removing the export would break those.

packages/bahmni-widgets/src/search/commonSearch/ResultsTable.tsx
kousalyapt-tw: done

packages/bahmni-widgets/src/search/commonSearch/__tests__/utils.test.ts
kousalyapt-tw: done

packages/bahmni-widgets/src/search/commonSearch/__tests__/ResultsTable.test.tsx
kousalyapt-tw: done

packages/bahmni-services/src/utils/utils.ts
kousalyapt-tw: done

packages/bahmni-services/src/utils/utils.ts
kousalyapt-tw: done

packages/bahmni-services/src/utils/__tests__/utils.test.ts
kousalyapt-tw: done

packages/bahmni-services/src/utils/__tests__/utils.test.ts
kousalyapt-tw: done

packages/bahmni-services/src/utils/__tests__/utils.test.ts
kousalyapt-tw: done

packages/bahmni-services/src/utils/__tests__/utils.test.ts
kousalyapt-tw: done

packages/bahmni-services/src/utils/utils.ts
rahu1ramesh: Do we need the type to be exported? Why does the consumer have to know of this.

packages/bahmni-services/src/utils/utils.ts
```diff
@@ -402,3 +402,39 @@ export function convertToSentenceCase(str: string): string {
...
+  options?: { count?: number; defaultValue?: string },
+) => string;
+
+export const formatGender = (value: string, t: Translator): string | null => {
+  const raw = value.trim();
+  if (!raw) return null;
+  return t(`GENDER_${camelToScreamingSnakeCase(raw)}`);
+};
+
+export const formatSearchResult = (
```
rahu1ramesh: This fn is specific to common search this can be kept in common search, no need to move this to bahmni-services

rahu1ramesh (changes requested):

rahu1ramesh (approved):

PR #79, opened 2025-10-10
BN-124|BN-175  New patient registration page with basic details and Additional Details

devsingh05298:
**JIRA** → [BN-124](https://bahmni.atlassian.net/browse/BN-124), [BN-175](https://bahmni.atlassian.net/browse/BN-175)

### **Description**
This PR introduces a comprehensive patient registration feature that enables healthcare staff to create new patient records in the Bahmni system. The implementation includes a multi-section form with robust validation, address hierarchy integration, and real-time age/DOB calculations.

### 1. __Patient Registration Form__

- Multi-section form layout with Basic Information, Address Information, Contact Information, and Additional Information sections
- Photo upload and capture functionality placeholders
- Patient identifier format selection with dynamic prefix management
- Manual entry toggle for identifier creation

### 2. __Form Validations__

- __Name Validation__: Ensures only alphabetic characters and spaces are allowed in name fields

- __Required Field Validation__: First name, last name, gender, and date of birth are mandatory

- __Age Validation__:

  - Years: Maximum 120
  - Months: Maximum 11
  - Days: Maximum 31

- __Phone Number Validation__: Accepts only numeric input

- __Address Validation__: Enforces selection from dropdown suggestions for district, state, and pincode

### 3. __Address Hierarchy Integration__

- Real-time address suggestions with auto-complete functionality
- Hierarchical auto-population: When pincode is selected, district and state auto-populate
- Debounced API calls (300ms) for optimal performance
- Minimum 2 characters required to trigger suggestions
- Visual feedback with parent address information in suggestions

### 4. __Age/DOB Bidirectional Calculation__

- Automatic DOB calculation when age is entered (years/months/days)
- Automatic age calculation when DOB is selected
- "Estimated" checkbox automatically checked when age is entered manually
- Supports entry in years, months, and days for precise age specification

### 5. __Gender Selection__

- Dynamic gender options fetched from global properties API
- Support for Male, Female, and Other gender options
- Internationalized gender labels

### UI Components

__Core Components Used:__

- `BaseLayout` - Main layout wrapper with header
- `Header` - Custom header component with breadcrumb navigation
- `TextInput` - For name, contact, and address fields
- `Dropdown` - For gender and patient ID format selection
- `DatePicker` & `DatePickerInput` - For date of birth selection
- `Checkbox` - For manual entry and estimated DOB
- `Button` - For form actions (Save, Print, Start OPD Visit)
- `Tile` - For section grouping

__Custom Styling:__

- Module-based SCSS for component-specific styles
[...]

PR #572, opened 2026-07-24
BAH-4704 | Refactor. Remove stale frontend validation workaround and validation moved to form2-controls

ROHIT-V-TW:
## Summary
- Removes `hasMissingMandatoryVisibleField()` and all related session-validation logic from `ObservationFormsContainer` and `observationFormsTransformer`
- Removes the corresponding tests for the removed transformer code
- The proper fix for mandatory field validation on hidden→visible fields is being implemented in the form2-controls library (see form2-controls PR #79)

## Why
The workaround tightly coupled the clinical app to form2-controls' internal Immutable.js state structure (`formContainerRef.current.state.data`), which is fragile and breaks the abstraction where `getValue().errors` should be the single source of truth. Moving the fix upstream to form2-controls eliminates this coupling.

## How Tested
- [x] Existing `ObservationFormsContainer` tests pass (31 tests)
- [x] No regressions in affected packages

## Acceptance Criteria Met
- [x] Stale frontend validation workaround removed
- [x] Fix delegated to form2-controls library

JIRA: BAH-4704

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated observation form validation so empty forms are consistently identified as empty.
  * Improved consistency in validation outcomes by relying on recorded values and validation errors.

* **New Features**
  * Expanded access to observation form transformation and note-extraction utilities for supported integrations.


apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
```diff
@@ -289,32 +288,19 @@ const ObservationFormsContainer: React.FC<ObservationFormsContainerProps> = ({
...
-        } | null
-      )?.state?.data;
-      const hasMissingMandatory = hasMissingMandatoryVisibleField(
-        convertImmutableToPlainObject(containerStateData) as
-          | Record<string, unknown>
-          | undefined,
-      );
-
-      if (isEmpty && !hasMissingMandatory) {
+      if (isEmpty) {
```
ravinderkabli: suggestion: For a form containing only mandatory fields that were never touched, this now shows the `EMPTY` state instead of `MANDATORY`.

Previously the guard was `if (isEmpty && !hasMissingMandatory)`, so an untouched mandatory-only form fell through to the mandatory branch and showed "required fields missing". With the unconditional `if (isEmpty)`, it now shows "form is empty" — misleading for a form that is actually missing required input (the save is still blocked either way).

Once form2-controls surfaces the mandatory error via `getValue().errors` for never-touched fields this resolves itself; if not, consider consulting `errors` for mandatory violations before returning `EMPTY` here.

ravinderkabli (commented):
## Review for BAH-4704

apps/clinical/src/components/forms/observations/ObservationFormsContainer.tsx
ROHIT-V-TW: Fixed — swapped the guard order so `errors` (including mandatory) are evaluated before the `isEmpty` check. An untouched mandatory-only form now correctly surfaces `MANDATORY` instead of `EMPTY`. Added a test case covering this scenario.

ravinderkabli (dismissed):
Approved

ravinderkabli (approved):
Looks fine

PR #580, opened 2026-07-29
BAH-4872 | Fix. Composite Criterion Keys For Disambiguation

rahu1ramesh:
**JIRA** → [BAH-4876](https://bahmni.atlassian.net/browse/BAH-4876)

## Description

Criteria that share the same field.key but differ by keyType (e.g. multiple identifier types like UMI, IME, PASSPORT) were being treated as the same criterion because comparisons used field.key alone. This caused incorrect deduplication in the search form and wrong criterion resolution at submission time. This fix introduces a criterionId utility that produces a composite key:keyType identifier, and applies it consistently across all lookup and comparison sites.

## Key Features

- Multiple identifier-type criteria (same field.key, different keyType) can now be selected and used simultaneously without collision.
- Search form correctly preselectss, deduplicates, resolves, and validates each identifier type independently.

[BAH-4876]: https://bahmni.atlassian.net/browse/BAH-4876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/bahmni-widgets/src/search/commonSearch/utils.ts
```diff
@@ -111,6 +112,9 @@ export const getRangeOrderError = (
   return new Date(fromVal) > new Date(toVal) ? errorMessage : null;
 };
 
+export const criterionId = (field: FieldConfig): string =>
```
mohan-13: Can we name the method better ? May be getCriterionId(field) ?

mohan-13 (commented):
@rahu1ramesh I see that we are generating the criterionId at multiple places. We should generate the Ids as soon as we process the config and have it stored.

This can provide some performance improvement and will be more consistent

rahu1ramesh:
> @rahu1ramesh I see that we are generating the criterionId at multiple places. We should generate the Ids as soon as we process the config and have it stored.
> 
> This can provide some performance improvement and will be more consistent

The reason, I did not go with this approach was to avoid a bigger blast radius of change. But sure makes sense, will make the change.

mohan-13 (approved):

PR #596, opened 2026-08-04
BAH-4878 |Fix. for sortOrder default-sort arrow direction

devsingh05298:
A fix for default sort direction for Column fields in common search page. 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved search results sorting to consistently respect the configured sort order.
  - Fields without a configured sort order now retain their original row order and correctly indicate that no sorting is applied.


devsingh05298:
The fix is already been done by another developer and merged. Hence closing this PR.

PR #597, opened 2026-08-04
BAH 4397 - One-Click Collapse/Expand Toggle

rajasurendra-tw:
> [!NOTE]
> Thank you for your contribution. Please find the details of this pull request below.

**JIRA** → [BAH-#](https://bahmni.atlassian.net/browse/BAH-#)

### **Description**

 

---

### **Key Features**
 
 
 

---

### **Implementation Details**

* **UI Components**
   
   

* **State Management**
   

* **Custom Hooks**
   

* **API Integration**
   
   

* **Performance & Optimization**
   

* **Internationalisation**
   

* **Accessibility**
   

* **Limitations**
   

---

### **Testing & Type Definitions**

* **Unit Tests**
   

* **Integration Tests**
   

* **Type Definitions**
   
   

---

### **Screenshots**

| Feature / State | Screenshot |
|------------------|------------|
| Example: Empty State | Attach image or paste screenshot here |
| Example: Search Result | Attach image or paste screenshot here |
| Example: Validation Error | Attach image or paste screenshot here |

_Note: Please ensure sensitive information is not included in screenshots._

---

### **Next Steps**
 
 

---

> [!IMPORTANT]
> **Checklist**
>
> - [ ] Code adheres to project linting and formatting standards.
> - [ ] New components added to Storybook.
> - [ ] Tests written and passing (unit + integration).
> - [ ] Labels and text support i18n.
> - [ ] Follows accessibility and responsive design guidelines.
> - [ ] PR reviewed by at least one other developer.

---

### **Reviewer(s)**

@bahnew/developers 
_Kindly review the proposed changes when convenient. Your feedback is appreciated._

---

## Summary by CodeRabbit

* **New Features**
  * Added the ability to expand and collapse the consultation action area.
  * Expanded mode hides side navigation and gives the consultation panel more space.
  * Added accessible expand/collapse labels in English and Spanish.
  * Improved action-area headers, including placement of form titles and pin controls.
  * Improved keyboard focus handling when expanding or collapsing the action area.

* **Bug Fixes**
  * Closing and reopening a consultation now resets the action area to its collapsed state.


packages/bahmni-design-system/src/templates/actionAreaLayout/index.tsx
```diff
@@ -52,15 +55,28 @@ const ActionAreaLayout: React.FC<ActionAreaLayoutProps> = ({
...
+              ? MAIN_DISPLAY_PANEL_EXPANDED_SIZE
+              : PANEL_MIN_SIZE
+          }
         >
           <div
             id="main-display-area"
             data-testid="main-display-area-test-id"
             aria-label="main-display-area-aria-label"
+            aria-hidden={isActionAreaExpanded}
+            inert={isActionAreaExpanded}
```
hamsavarthiniR-Bahmni: blocking: `inert={isActionAreaExpanded}` will misbehave on React 18. React only treats `inert` as a true boolean HTML attribute (omitted from the DOM when `false`) starting in React 19 — on React 18 it renders as `inert="false"`, and per the HTML spec the mere *presence* of the attribute makes the element inert regardless of its string value. This package's `package.json` still declares `"react": ">=18.0.0"` as a peer dependency, so this would permanently disable the main display panel for any React 18 consumer.

Either bump the peer dependency floor to `>=19.0.0`, or spread the attribute conditionally: `{...(isActionAreaExpanded ? { inert: true } : {})}`.

packages/bahmni-design-system/src/molecules/actionArea/styles/ActionArea.module.scss
```diff
@@ -18,17 +18,44 @@
...
+  gap: $spacing-03;
 }
 
 .content {
   flex: 1;
   overflow-y: auto;
 }
 
+.cappedWidth {
+  max-width: 75rem;
```
hamsavarthiniR-Bahmni, line 55: suggestion: AC2 requires controls/forms to resize and "fit the screen space" when expanded to 100%, but `.cappedWidth { max-width: 75rem }` leaves large empty gutters on screens wider than ~1200px instead of truly filling the width. This may be an intentional readability choice — worth explicit product/QA sign-off that it satisfies the AC as worded, otherwise consider removing/raising the cap.

apps/clinical/src/components/consultationPadContainer/index.tsx
```diff
@@ -218,6 +224,10 @@ const ConsultationPadContainer: React.FC<ConsultationPadContainerProps> = ({
           </div>
         }
         ariaLabel={t('CONSULTATION_PAD_TITLE')}
+        isExpanded={isActionAreaExpanded}
```
hamsavarthiniR-Bahmni: suggestion: This 4-line expand/collapse prop block (`isExpanded`, `onToggleExpand`, `expandAriaLabel`, `collapseAriaLabel`) is duplicated verbatim 4 times across this file, `consultationPad/index.tsx`, and `ObservationFormsContainer.tsx`. Consider extracting a small shared hook (e.g. `useActionAreaExpandProps()`) and spreading it at each `<ActionArea>` call site to avoid drift when this needs to change.

apps/clinical/src/components/consultationPad/index.tsx
```diff
@@ -569,6 +573,10 @@ const ConsultationPad: React.FC<ConsultationPadProps> = ({
         secondaryButtonText={t('CONSULTATION_PAD_CANCEL_BUTTON')}
         onSecondaryButtonClick={handleCancel}
         content={renderPadContent}
+        isExpanded={isActionAreaExpanded}
```
hamsavarthiniR-Bahmni: nit: This pad-level `ActionArea` is `hidden={!!viewingForm}` (CSS `visibility: hidden`, still mounted) but still receives `isExpanded`/`onToggleExpand`/aria labels, so while a form is open there are two DOM elements simultaneously carrying `data-testid="action-area-expand-toggle"` — this one and the one in `ObservationFormsContainer`. It's currently harmless only because `visibility:hidden` removes it from tab order/pointer events. Consider omitting these props when `viewingForm` is truthy, since this instance is hidden anyway.

packages/bahmni-design-system/src/molecules/actionArea/ActionArea.tsx
```diff
@@ -1,13 +1,19 @@
+import { Maximize, Minimize } from '@carbon/icons-react';
 import { Button, ButtonSet } from '@carbon/react';
 import classNames from 'classnames';
-import React, { ReactNode } from 'react';
+import React, { ReactNode, useEffect, useRef } from 'react';
+import { IconButton } from '../../atoms/iconButton';
 import styles from './styles/ActionArea.module.scss';
 
+const TOGGLE_BUTTON_SELECTOR = '[data-testid="action-area-expand-toggle"]';
```
hamsavarthiniR-Bahmni: nit: Focus-restoration logic queries the DOM by `data-testid`, coupling production behavior to a testing attribute — renaming/removing the test id would silently break keyboard-focus restoration with no compile-time signal. Since `IconButton` doesn't forward refs, consider wrapping it with `React.forwardRef` and attaching a real ref instead of `querySelector('[data-testid=...]')`.

apps/clinical/src/pages/ConsultationPage.tsx
```diff
@@ -82,6 +82,7 @@ const ConsultationPage: React.FC = () => {
   const { userPrivileges } = useUserPrivilege();
   const { addNotification } = useNotification();
   const [isActionAreaVisible, setIsActionAreaVisible] = useState(false);
+  const [isActionAreaExpanded, setIsActionAreaExpanded] = useState(false);
```
hamsavarthiniR-Bahmni, line 85: question: `isActionAreaExpanded` is reset to `false` in the pad's `onClose` handler, but not in the `useSubscribeConsultationStart` callback below. Currently harmless since the only caller doesn't hit this path while expanded, but if a future/alternate entry point fires a consultation-start event while the pad is already expanded, the new session would render already-expanded. Worth also resetting it here for defense-in-depth?

apps/clinical/src/pages/__tests__/ConsultationPage.test.tsx
```diff
@@ -729,4 +750,39 @@ describe('ConsultationPage', () => {
       });
     });
   });
+
+  describe('Side nav visibility when consultation pad is expanded', () => {
```
hamsavarthiniR-Bahmni, line 759: nit: This test covers toggling expand/collapse, but there's no test exercising expand → `onClose` → reopen to verify the pad comes back collapsed (the side-nav-restore-on-close behavior mentioned in the PR description). Worth adding one given it's called out as a fix in this PR.

hamsavarthiniR-Bahmni (commented):
## Review for BAH-4397

Solid feature overall — architectural placement of generic UI logic vs. domain wiring is correct, and i18n/testing groundwork is in good shape. Found 1 blocking issue, 3 suggestions, and 4 nits/questions worth addressing before merge.

packages/bahmni-design-system/src/molecules/actionArea/styles/ActionArea.module.scss
rajasurendra-tw: Suman has discussed with the UI team and told that any expansion of the screen after that point, the form size should not expand more than the size mentioned and further expansion will need to be applied on the empty space rather than the form

apps/clinical/src/pages/ConsultationPage.tsx
rajasurendra-tw: Addressed this comment.  added the reset in useSubscribeConsultationStart's callback too (setIsActionAreaExpanded(false) right after setIsActionAreaVisible(true)), so a new consultation session always starts collapsed regardless of entry point.

apps/clinical/src/pages/__tests__/ConsultationPage.test.tsx
rajasurendra-tw: Addressed this comment. Added 'reopens collapsed after being expanded and then closed', it expands the pad, closes it, starts a new consultation, and asserts the toggle reads 'Expand' (not 'Collapse') on reopen. That last assertion is what actually catches a regression if the reset above were removed. Had to extend the consultationPadContainer mock to expose an onClose trigger button since it previously only wired up the expand toggle.

hamsavarthiniR-Bahmni (approved):

PR #600, opened 2026-08-05
BAH-4689 | Stop Medication — Standard FHIR POST, note alignment, configurable stop reason

ROHIT-V-TW:
## Summary

Refactors the stop medication feature from a custom `$stop` FHIR operation to standard FHIR resources (`POST Encounter` + `POST MedicationRequest`), ensuring AtomFeed events fire correctly for downstream sync.

### Key Changes

**Stop Flow (major refactor)**
- Replaced `POST /MedicationRequest/{id}/$stop` (custom op) with standard FHIR:
  1. `POST /Encounter` — creates/reuses session encounter (AtomFeed `ENCOUNTER_CREATED` ✅)
  2. `POST /MedicationRequest` with `status=stopped + priorPrescription + encounter + statusReason + note`
- Reuses existing session encounter when available; falls back to creating a new one with the practitioner as participant (fixes "Edit blocked after stop" bug)

**Backend (`bahmni-module-fhir2-addl-extension`)**
- Removed `$stop` custom operation from `BahmniMedicationRequestFhirR4Provider`
- Added `translateStopMedicationOrder()` to `BahmniMedicationRequestTranslatorImpl.toOpenmrsType()` — handles stop write path:
  - `statusReason.coding` → `order_reason` (concept FK)
  - `statusReason.text` → `order_reason_non_coded`
  - `note[cancellation-note]` → `comment_to_fulfiller`
  - `dateStopped` extension → DC order `dateActivated`
  - Copies `concept`, `drug`, `orderer`, `asNeeded` from prior order

**Frontend (`bahmni-apps-frontend`)**
- `stopMedicationService.ts` — two-call stop flow: Encounter POST then MedicationRequest POST
- Note alignment fix — `TextAreaWClose` subgrid padding overridden via `noteColumn` style
- Stop reason concept set name made configurable via `stopOrderedReasonConceptSet` in `medication.json` (fallback: `"Stopped Order Reason"`)
- `sessionEncounterUuid` plumbing removed from store/form/consultationPad (encounter resolved in service)
- FHIR `note-category` extension (`"order-note"` / `"cancellation-note"`) for tooltip display

**Bug fixes**
- Edit button blocked after stop: fixed by ensuring stop encounter has practitioner participant so `findActiveEncounterInSession` can find it
- Stop reason dropdown not loading: removed `enabled: medicationConfig !== undefined` guard

## Related
- Backend PR: https://github.com/Bahmni/bahmni-module-fhir2-addl-extension/pull/105
- Standard config: https://github.com/Bahmni/standard-config/pull/177

## Summary by CodeRabbit

- **New Features**
  - Added configurable stop-medication reasons.
  - Replaced the always-visible note field with an **Add Note** option and closable textarea.
  - Added support for categorized cancellation notes.
  - Medication lists now display order notes and cancellation notes separately in tooltips.

- **Bug Fixes**
  - Improved stop-medication processing by reliably associating actions with the correct patient encounter.
  - Added clearer validation and handling when required encounter information is unavailable.

- **Documentation**
  - Added guidance on medication stopping, editing, notes, and instruction data flow.


apps/clinical/src/services/stopMedicationService.ts
```diff
@@ -41,36 +52,118 @@ export async function fetchStopReasons(): Promise<StopReason[]> {
   }
 }
 
+async function resolveStopEncounter(patientUuid: string): Promise<string> {
```
ravinderkabli: **resolveStopEncounter()** exists

It exists because the PR kept the onDirectSubmit pattern while changing the underlying operation to one that now requires an encounter. That combination forced the method into existence as a workaround.

The onDirectSubmit pattern was designed for the old $stop custom operation specifically because $stop did not need an encounter — it called orderService.discontinueOrder() directly on the existing drug order. No new encounter, no bundle.

Now the operation has changed to POST /MedicationRequest with status=stopped, which does require an encounter reference. But instead of changing the submit path to match, the PR kept onDirectSubmit and bolted encounter resolution onto the service.

---
**What should have happened instead**

**The stop flow should switch from onDirectSubmit to createBundleEntries — exactly the same path that edit medication uses. This would:**

Before (current PR):
  onDirectSubmit → resolveStopEncounter() [separate HTTP call]
                 → POST /MedicationRequest [separate HTTP call]

After (correct approach):
  createBundleEntries → ConsultationPad bundles Encounter + MedicationRequest together
                      → POST /Bundle [single atomic transaction] ✅ AC4 met

The ConsultationPad already handles all of this for every other form:
- Finding or creating an encounter
- Wrapping it with the FHIR resource in a single transaction Bundle
- Submitting atomically
- Error handling with i18n keys
- Resetting state on success

**If stop medication routed through createBundleEntries, resolveStopEncounter() would be deleted entirely.**

Edit_Stop_Medications_Deep_Dive.md
```diff
@@ -0,0 +1,1293 @@
...
+User clicks Stop
+  → actionHandlers.ts dispatches startConsultation
+  → ConsultationPage renders ConsultationPad (editOnly: 'stopMedications')
+  → StopMedicationForm renders
+  → TanStack Query fetches: date range + stop reasons
+  → User fills form (date, reason, note)
+  → Submit → stopMedications index.ts onDirectSubmit
+  → stopMedicationService.ts: POST $stop FHIR operation
+  → consultationSaved event → MedicationsTable refetch
+```
```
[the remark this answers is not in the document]
ROHIT-V-TW: Removed file as not valid to keep

Edit_Stop_Medications_Deep_Dive.md
```diff
@@ -0,0 +1,1293 @@
...
+
+Validation respects field visibility config:
+```ts
+validate: () => {
+  const cfg = get().fieldConfig;
+  if (cfg.stopDate?.isVisible !== false && cfg.stopDate?.isMandatory && !get().stopDate) return false;
+  if (cfg.stopReason?.isVisible !== false && cfg.stopReason?.isMandatory && !get().stopReason) return false;
+  return true;
+}
```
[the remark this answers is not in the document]
ROHIT-V-TW: Removed the file

Edit_Stop_Medications_Deep_Dive.md
```diff
@@ -0,0 +1,1293 @@
...
+
+{
+  "resourceType": "Parameters",
+  "parameter": [
+    { "name": "reason",        "valueString": "Side effects" },
+    { "name": "effectiveDate", "valueDate": "2025-07-02" },
+    { "name": "note",          "valueString": "Patient requested" }
+  ]
+}
+```
```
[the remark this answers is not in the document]
ROHIT-V-TW: Removed the file

Note_Additional-Instructions Analysis.md
```diff
@@ -0,0 +1,218 @@
...
+| **instructions** | As directed | As directed |
+| **additionalInstructions_from_json** | `Old Bahmni (Additional)` | `NULL` |
+| **comment_to_fulfiller** | `NULL` | `New Bahmni (Note)` |
+| **treatmentNotes_as_print_would_show** | `Old Bahmni (Additional)` ✅ | `New Bahmni (Note)` ✅ |
+
+### What this confirms
+
+- **Old Bahmni** stores "Additional Instructions" in `dosing_instructions` JSON → `additionalInstructions_from_json` populated, `comment_to_fulfiller` is NULL
+- **New Bahmni** stores "Note" in `comment_to_fulfiller` → `additionalInstructions_from_json` is NULL, `comment_to_fulfiller` populated
+- **`treatmentNotes_as_print_would_show`** resolves correctly for **both** via the CASE fallback logic — validating that the proposed 1-line fix to `compute.js` will work
```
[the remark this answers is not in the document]
ROHIT-V-TW: Removed the file

Note_Additional-Instructions Analysis.md
```diff
@@ -0,0 +1,218 @@
...
+
+### Write path
+Write to the new field going forward. Old records stay in old format — the layered read path handles both. **No DB migration needed** since `dosing_instructions` is a free-text column.
+
+### Print template
+Add fallback read in `parseAdditionalInstructions` or at the call site in `compute.js`:
+```js
+treatmentNotes: parseAdditionalInstructions(mr.dosageInstruction?.[0]?.text)
+             || mr.note?.[0]?.text  // fallback for new Bahmni medications
+             || '';
```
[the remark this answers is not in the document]
ROHIT-V-TW: Removed the file

Note_Additional-Instructions Analysis.md
```diff
@@ -0,0 +1,218 @@
...
+                                (never set by new FE)                                                                 → printed as "Treatment Notes"
+```
+
+### Print Template — Exact Read Path (`compute.js:56`)
+```js
+treatmentNotes: parseAdditionalInstructions(mr.dosageInstruction?.[0]?.text)
+// where:
+function parseAdditionalInstructions(text) {
+  return JSON.parse(text)?.additionalInstructions ?? '';
+}
```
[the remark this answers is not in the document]
ROHIT-V-TW: Removed the file

Edit_Stop_Medications_Deep_Dive.md
```diff
@@ -0,0 +1,1293 @@
...
+
+**Deferred dispatch with `setTimeout(fn, 0)`**:
+```ts
+// Deferred to next event loop tick — lets UI updates complete first
+export const dispatchConsultationSaved = (payload): void => {
+  const event = new CustomEvent(CONSULTATION_SAVED_EVENT, { detail: payload });
+  window.dispatchEvent(event);   // synchronous in this version
+};
```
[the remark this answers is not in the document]
ROHIT-V-TW: Removed the file

apps/clinical/src/services/stopMedicationService.ts
```diff
@@ -1,36 +1,39 @@
...
 }
 
-/**
- * Fetches stop reasons from the FHIR ValueSet "Stopped Order Reason".
- *
- * 1. Searches: GET /ws/fhir2/R4/ValueSet?title=Stopped+Order+Reason
- * 2. Expands: GET /ws/fhir2/R4/ValueSet/{uuid}/$expand
- * 3. Returns the expanded concepts as stop reasons
- */
-export async function fetchStopReasons(): Promise<StopReason[]> {
```
SrijanRamSN, line 21: can we accept the Uuid of the stopOrderedReasonConcept directly than by name ?

ROHIT-V-TW: nice suggestion Srijan, have made the change now

SrijanRamSN (approved):

apps/clinical/src/services/stopMedicationService.ts
```diff
@@ -1,36 +1,51 @@
...
+      const valueSetEntry = searchBundle.entry?.[0]?.resource as
+        | ValueSet
+        | undefined;
+      valueSetId = valueSetEntry?.id;
+    }
 
+    if (!valueSetId) return [];
     const expanded = await get<ValueSet>(
-      STOP_REASON_VALUESET_EXPAND_URL(valueSetEntry.id),
+      STOP_REASON_VALUESET_EXPAND_URL(valueSetId),
```
SrijanRamSN, line 42: wouldn't this create a double fall back ? if conceptSetNameOrUuid is empty the method will go through both callbacks . can we  call expand<uuid> first ?

apps/clinical/src/services/stopMedicationService.ts
```diff
@@ -1,36 +1,51 @@
...
- * 1. Searches: GET /ws/fhir2/R4/ValueSet?title=Stopped+Order+Reason
- * 2. Expands: GET /ws/fhir2/R4/ValueSet/{uuid}/$expand
- * 3. Returns the expanded concepts as stop reasons
- */
-export async function fetchStopReasons(): Promise<StopReason[]> {
+const UUID_REGEX =
+  /^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i;
+
+export async function fetchStopReasons(
+  conceptSetNameOrUuid?: string,
```
SrijanRamSN: we are not passing the Name now right ? if name is taken from the STOP_REASON_VALUESET_URL as a fall back only , can we keep this as conceptSetUuid/stopReasonConceptSetUuid  then ?

apps/clinical/src/components/forms/stopMedications/StopMedicationForm.tsx
```diff
@@ -80,9 +88,13 @@ const StopMedicationForm: React.FC<StopMedicationFormProps> = React.memo(
       },
     });
 
+    const stopReasonConceptSet = inputControlConfig?.metadata
```
SrijanRamSN: can we keep the STOP_REASON_VALUESET_TITLE here ? if the  stoppedOrderReasonConceptSet from inputControlConfig is undefined passing the fall back name here itself can avoid the double fall back logic in the service ?

SrijanRamSN (changes requested):
don't we need to capture the stoppedOrderReasonConceptSet in the actionhandler  / handleStopAction ?

apps/clinical/src/services/stopMedicationService.ts
ROHIT-V-TW: Done — moved the fallback title to the form itself, so the service always receives a string. No double fallback now.

apps/clinical/src/services/stopMedicationService.ts
ROHIT-V-TW: Renamed to conceptSetUuid — good call, since we only pass UUIDs now (or the fallback title name as a string, handled before it reaches the service).

apps/clinical/src/components/forms/stopMedications/StopMedicationForm.tsx
ROHIT-V-TW: Done exactly this — form now resolves the fallback (STOP_REASON_VALUESET_TITLE) when stoppedOrderReasonConceptSet is undefined, so the service always gets a non-null value and doesn't need to handle the fallback itself.

ROHIT-V-TW:
stoppedOrderReasonConceptSet in handleStopAction — the metadata doesn't need to flow through the action handler. When the ConsultationPad opens in editOnly mode, loadEncounterInputControls processes all inputControls from app.json (including the stopMedications entry with its metadata). The StopMedicationForm then receives inputControlConfig as a prop from the renderer, so it already has access to stoppedOrderReasonConceptSet via inputControlConfig.metadata. The action handler only needs to tell the pad which form to open — the form reads its own config independently.

SrijanRamSN (approved):

PR #601, opened 2026-08-05
BAH-4955|Fix. Date fields are not sorted appropriately

devsingh05298:
Fix. for sorting the date on common search page , initially it was sorting on basis of date and month. 

## Bug Fixes

* Improved search result tables displaying formatted date, time, datetime, and age values.
* Sorting and filtering now use the underlying values while showing readable formatted text.
* Prevented transformed display values from affecting ordering and filtering accuracy.
* Fixed cases where raw date values appeared instead of their formatted representation.
* Added support for reliable sorting of age values based on birth dates.

Sorting for Transform Age : 
image (8)
Sorting for Transform Date : 
image (9)


packages/bahmni-widgets/src/search/commonSearch/ResultsTable.tsx
```diff
@@ -55,15 +55,24 @@ const evaluateRows = async (
...
           }
         }
-      }
 
-      for (const { key, expr, transform } of compiled) {
-        const value = await expr.evaluate(item as Record<string, unknown>);
         if (!value) {
           row[key] = '-';
           continue;
         }
```
PratitiP, line 66: Here, we are checking only for date fields which will are getting formatted for display. But there can be other transformed fields like age, gender, country. Can we make this generic to handle for all the fields which have transform function configured

packages/bahmni-widgets/src/search/commonSearch/utils.ts
```diff
@@ -63,6 +63,16 @@ export const resultTransforms: Record<string, ResultTransform> = {
   formatSearchResult,
 };
```
PratitiP, line 77: Can we rename this constant simply TRANSFORM_KEYS ? as we use transform key on config.

PratitiP (dismissed):

PratitiP (approved):

PR #609, opened 2026-08-11
BAH-4956 | Applicant context search - Add search criterias

devsingh05298:
**JIRA** → [BAH-4956](https://bahmni.atlassian.net/browse/BAH-4956)

## **Description**

As a clinical or registration user, I want to search for an applicant using applicant details such as name, ID, or phone number, So that I can locate the applicant record I need and take action.

## Assumptions

Searching applicant specific to location is out of scope of this card. and maybe tackled as part of multi-tenancy. 

[BAH-4956]: https://bahmni.atlassian.net/browse/BAH-4956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/bahmni-widgets/src/search/commonSearch/schema.json
```diff
@@ -129,7 +128,11 @@
           "minLength": 1,
           "description": "i18n key for the criterion label"
         },
-        "input": { "$ref": "#/definitions/inputConfig" }
+        "input": { "$ref": "#/definitions/inputConfig" },
+        "requiresAdditionalCriterion": {
```
rahu1ramesh: @devsingh05298 can we have a less verbose key like `standalone`?

packages/bahmni-widgets/src/search/commonSearch/utils.ts
```diff
@@ -375,8 +387,37 @@ export const validateRows = (
       rangeOrderMessage,
       t,
     );
+    if (
+      !validationError &&
+      !rangeOrderError &&
+      criterion.requiresAdditionalCriterion &&
```
rahu1ramesh: @devsingh05298 what happens when 2 fields with requiresAdditionalCriterion set to true is selected, will we block the search or does this qualify validation?

rahu1ramesh (changes requested):

packages/bahmni-widgets/src/search/commonSearch/utils.ts
devsingh05298: With the current implementation, it would NOT be blocked — the search would be allowed to proceed.
 If you select two fields that both have requiresAdditionalCriterion: true (e.g., hypothetically Sex + some other field), neither gets the error — search passes.

packages/bahmni-widgets/src/search/commonSearch/schema.json
devsingh05298: no need of this now.

PR #586, opened 2026-07-31
BAH-4911 | Refactor. SearchPatient As A Search Extension Widget

rahu1ramesh:
**JIRA** → [BAH-4911](https://bahmni.atlassian.net/browse/BAH-4911)

## Description

Refactors SearchPatient from a directly-prop-driven component into a self-contained search extension widget. It now accepts extensionParams (consistent with CommonSearchWidget) and fetches its own configuration via configUrl, enabling it to be registered and rendered through the extension system in RegistrationList without any app-level prop wiring. The patientSearch configuration block is removed from RegistrationConfig entirely, as it is now owned by the widget's own config URL.

## Key Features

- SearchPatient is now a proper search extension widget — registered under the key defaultSearch and rendered via the SearchHandler extension point in RegistrationList.
- RegistrationList replaces PatientSearchPage as the registration app's main list/search page, supporting multiple tabbed search extensions.
- Extension models and utilities (Extension, groupExtensionsByPoint, filterExtensionsByPrivileges) promoted from the clinical app into @bahmni/services for shared use.
- patientSearch block removed from RegistrationConfig and its schema — search config is now fetched independently by the widget.

[BAH-4911]: https://bahmni.atlassian.net/browse/BAH-4911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #612, opened 2026-08-12
Bah 4581 print registration card

AkshataKeshatwarTWer:
> [!NOTE]
Please find the details of this pull request below.

**JIRA** → [BAH-4581](https://bahmni.atlassian.net/browse/BAH-4581)

[BAH-4581]: https://bahmni.atlassian.net/browse/BAH-4581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added printing support for patient registration cards.
  * Print options are now filtered according to user privileges.
  * Added configurable registration print options and templates.
  * Added English and Spanish text for document preparation and printing.

* **Documentation**
  * Updated print button guidance, configuration examples, privilege settings, and template registration instructions.


apps/registration/src/providers/registrationConfig/models.ts
```diff
@@ -1,4 +1,15 @@
-import { Extension } from '@bahmni/services';
+import {
+  AppointmentSearchField,
+  Extension,
+  PatientSearchField,
+} from '@bahmni/services';
+import type { PrintOption } from '@bahmni/widgets';
+
+export interface PatientSearchConfig {
```
hamsavarthiniR-Bahmni: blocking: This `PatientSearchConfig` interface is a stale, pre-BAH-4911 shape (it has an extra `patientDetailUrl?: string` field not present on `main`), and it's now completely orphaned — zero references anywhere in `apps/registration/src`.

More importantly: `main` already has `patientSearch: PatientSearchConfig` as a **required** field on `RegistrationConfig` (added by the separately-merged BAH-4911 PR), plus a full `patientSearch` JSON-schema block in `schema.json`. This branch's merge with `origin/main` (commit `e5acb336`) dropped both — confirmed via `git show main:...models.ts` (has the field) vs. this branch (doesn't) and `git show main:...schema.json` (has the block) vs. this branch (zero occurrences).

Since the schema's root has `additionalProperties: false`, any existing deployment with `patientSearch` configured (which is now every deployment, per BAH-4911) would fail schema validation and fail to load its config if this merges as-is.

Please redo the merge from `main` for this file and `schema.json`: restore `patientSearch: PatientSearchConfig` on `RegistrationConfig` and the full schema block using `main`'s `PatientSearchConfig` shape (no `patientDetailUrl`), then layer only the print-feature additions on top.

apps/registration/src/providers/registrationConfig/models.ts
```diff
@@ -104,6 +115,7 @@ export interface RegistrationConfig {
   registrationForm?: RegistrationFormConfig;
   extensionPoints?: ExtensionPoint[];
   registrationAppExtensions?: AppExtensionConfig[];
+  printOptions?: PrintOption[];
```
hamsavarthiniR-Bahmni, line 108: question: Just flagging alongside the comment above — this `printOptions` field is the new, correct addition. It's `patientSearch` that went missing during the merge, not this line.

apps/registration/src/pages/PatientRegister/PatientRegister.tsx
```diff
@@ -356,6 +371,17 @@ const PatientRegister = () => {
                 >
                   {t('CREATE_PATIENT_SAVE')}
                 </Button>
+                {patientUuid && (
```
hamsavarthiniR-Bahmni, line 391: blocking: JIRA AC #1/#5/#6 describe the print button as configured via `registrationAppExtensions` — "using the same extension mechanism as Start Visit" — with an extension carrying `type="print"` and `templateId`. This implementation instead uses a completely separate, new `printOptions` config array, bypassing the `useFilteredExtensions`/`RegistrationActions` pipeline entirely.

`AppExtensionConfig.type` is still `'link' | 'startVisit'` (unchanged), and the schema enum for `registrationAppExtensions[].type` is still `["link","startVisit"]`. An implementer literally following AC #6 (add a `type="print"` extension to `registrationAppExtensions`) would produce a config that fails schema validation, not a working print button.

The end-user-visible behavior (button in the footer, shows/hides correctly) does satisfy the intent, but the configuration contract promised by the ticket isn't what was built. Please get explicit sign-off from the PO/ticket owner on this deviation before merge — either update the AC to describe `printOptions`, or rework this to actually extend the extension mechanism.

apps/registration/src/pages/PatientRegister/PatientRegister.tsx
```diff
@@ -46,6 +53,14 @@ const PatientRegister = () => {
   const { shouldShowAdditionalIdentifiers } = useAdditionalIdentifiers();
   const { relationshipTypes } = useRelationshipValidation();
   const { registrationConfig } = useRegistrationConfig();
+  const { userPrivileges } = useUserPrivilege();
+
+  const filteredPrintOptions = useMemo<PrintOption[]>(() => {
```
hamsavarthiniR-Bahmni, line 61: suggestion: `hasPrivilege` returns `false` whenever `userPrivileges` is empty — even for a `printOptions` entry with no `privileges` restriction at all (meant to be visible to everyone per the schema, where `privileges` is optional). A user with a genuinely empty privilege array would have every "public" print option filtered out here. Your own new docs (`docs/document-print-button-guide.md`) call this out explicitly but don't fix it.

Consider special-casing `!option.privileges?.length` to always include the option, or fixing `hasPrivilege` itself to check for an empty required-privilege list before checking the user's privileges.

Also worth a second look: this only guards `!userPrivileges` (null), not the loading state from `useUserPrivilege()` — contrast with `useFilteredExtensions.ts`, which explicitly returns an empty result while privileges are still loading. This could cause a flash-of-missing-button on initial load.

apps/registration/src/pages/PatientRegister/PatientRegister.tsx
```diff
@@ -356,6 +371,17 @@ const PatientRegister = () => {
                 >
                   {t('CREATE_PATIENT_SAVE')}
                 </Button>
+                {patientUuid && (
+                  <DocumentPrintButton
+                    printOptions={filteredPrintOptions}
+                    renderContext={{
+                      patientUuid,
```
hamsavarthiniR-Bahmni, line 395: blocking: `patientUuid` here can be stale for the previous patient. `/patient/:patientUuid` is a single route — React Router won't remount this component across navigations to a different patient within the same route match (e.g. browser back/forward). `patientUuid` state is only ever moved forward via an effect keyed on `metadata`, and during the loading window for a newly-navigated-to patient, the previous patient's `metadata` reference is unchanged, so that effect doesn't fire.

Net effect: for a window after navigating to a different patient's edit URL, this `DocumentPrintButton` can render with the *previous* patient's UUID — a user could print the wrong patient's registration card, which is a real patient-safety concern.

Suggest resetting `patientUuid`/`metadata` whenever the URL's `patientUuid` param changes, or additionally gating this button on the patient-details query's loading state.

apps/registration/src/pages/PatientRegister/__tests__/PatientRegister.test.tsx
```diff
@@ -1351,4 +1353,105 @@ describe('PatientRegister', () => {
       });
     });
   });
+
+  describe('Print Registration Card', () => {
```
hamsavarthiniR-Bahmni, line 1381: nit: A few coverage gaps in this describe block worth filling in:
- No test asserts what `renderAsHtml` is actually called with after clicking print — nothing verifies `renderContext` reaches the render API with both `patientUuid`/`patientUUID` keys, despite your own docs calling this dual-key behavior out as important.
- No test sets `isPending: true` and checks the print button is disabled, even though this exact pattern already exists for the Save button elsewhere in this file.
- All privilege tests are all-or-nothing (single option, present or absent) — no test with two options (one allowed, one not) to confirm the filter keeps only the allowed one.

hamsavarthiniR-Bahmni (changes requested):
## Review for BAH-4581

The end-user-visible print feature is well-built and well-tested for the happy path, but this PR has a serious regression and a significant requirements gap that need to be resolved before merge.

Found 1 blocking issue, 2 critical, 1 suggestion, and 2 nits.

**Acceptance Criteria**: 5/8 fully met, 3 at-risk/not-met — the print button's configuration mechanism doesn't match what AC #1/#5/#6 literally specify (`registrationAppExtensions` extension pattern), and there's a merge-artifact regression unrelated to this PR's stated scope.

### Most important: this PR silently drops an already-shipped, required config field

`main` already has `patientSearch: PatientSearchConfig` (required) on `RegistrationConfig`, plus a full JSON-schema block for it (from the separately-merged BAH-4911 SearchPatient-widget PR). I verified directly via `git show main:...` vs. this PR's branch: this PR's `models.ts` keeps a stale, orphaned `PatientSearchConfig` interface but has **lost the `patientSearch` field on `RegistrationConfig` entirely**, and `schema.json` has **zero** occurrences of `patientSearch` (vs. a full block on `main`). Because the schema's root has `additionalProperties: false`, any existing deployment with `patientSearch` configured (which every deployment now has, per BAH-4911) would fail schema validation and fail to load its registration config if this PR merges as-is. This looks like a bad merge-conflict resolution when this branch was merged with `origin/main` (commit `e5acb336`), not deliberate print-feature work.

Full local review: `PR_612_code_review.md`

apps/registration/src/providers/registrationConfig/models.ts
AkshataKeshatwarTWer: `patientSearch` was **removed** from the registration config by BAH-4911, not added by it. The commit is `870b8276` ("BAH-4911 | Refactor. SearchPatient As A Search Extension Widget", #586):

```
$ git show 870b8276 -- apps/registration/src/providers/registrationConfig/models.ts
-export interface PatientSearchConfig {
-  customAttributes: PatientSearchField[];
-  patientSearch: PatientSearchConfig;
```

Current `main` confirms it — both of these return `0`:

```
$ git show origin/main:apps/registration/src/providers/registrationConfig/models.ts  | grep -c patientSearch
$ git show origin/main:apps/registration/src/providers/registrationConfig/schema.json | grep -c patientSearch
```

`PatientSearchConfig` now exists only in clinical (`apps/clinical/src/providers/clinicalConfig/models.ts:64`), and with a different shape — `{ displayFields?: PatientSearchDisplayField[] }`.

The matching half of that refactor landed in clinic-config: `001a5e2` ("BAH-4911 | Add. Default Search Config For Registration", #302) deleted the top-level `patientSearch` block from `openmrs/apps/registration/v2/app.json` and replaced it with an `extensions` entry pointing at `defaultSearchConfig.json`. So restoring `patientSearch` here would resurrect a contract BAH-4911 retired on both sides — and because the root schema is `additionalProperties: false`, adding it back is the change that would risk breaking config loads, not leaving it out.

You were right about the stale orphaned interface, though. The `PatientSearchConfig` carrying the extra `patientDetailUrl` was a merge artifact from `e5acb336`; that commit is gone. The branch has been rebased onto `main`, and `models.ts` now differs from `main` by exactly two lines — the `PrintOption` import and `printOptions?`. Worth re-reading from the current head (`0598fa3d`).

apps/registration/src/pages/PatientRegister/PatientRegister.tsx
AkshataKeshatwarTWer: The reason it's built as a `printOptions` array is that printing is already modelled that way in this codebase. `main` ships dashboard-level `printOptions` in clinical (`clinicalConfig/schema.json:206`, consumed at `ConsultationPage.tsx:199`), both apps render the same `DocumentPrintButton` from `@bahmni/widgets`, and the extension pipeline is navigation-oriented (`onBeforeNavigate`, `order`, `link`/`startVisit` semantics) — a `type="print"` extension would be one that never navigates.

Could the ticket owner confirm which contract is wanted:

1. update AC #1/#5/#6 to describe `printOptions`, keeping registration consistent with clinical; or
2. rework to `registrationAppExtensions` with `type="print"`, which also needs the `AppExtensionConfig.type` union widened, the schema enum extended, and a matching clinic-config rewrite.

Happy to do (2) if that's the decision — leaving this unresolved rather than closing it.

apps/registration/src/pages/PatientRegister/PatientRegister.tsx
AkshataKeshatwarTWer: An option with no `privileges` is now short-circuited as unrestricted, so an empty privilege array no longer filters out options meant to be visible to everyone:

```ts
!option.privileges?.length || hasPrivilege(userPrivileges, option.privileges)
```

That's the same shape `useFilteredExtensions` already uses (`!ext.requiredPrivilege || hasPrivilege(...)`), so I left the shared `hasPrivilege` untouched rather than changing behaviour for its other callers.

Also took the loading point — the filter now returns `[]` while `useUserPrivilege()` reports `isLoading`, matching `useFilteredExtensions`, so the button no longer flashes in before privileges resolve. Both cases have tests.

---

apps/registration/src/pages/PatientRegister/PatientRegister.tsx
AkshataKeshatwarTWer: Confirmed and fixed in `0598fa3d`. The route element is keyed by `route.path` rather than the uuid (`routes/index.tsx`), so the component is indeed reused across patients.

Rather than resetting in an effect, the URL is now authoritative:

```ts
const patientUuid = patientUuidFromUrl ?? savedPatientUuid;
```

`savedPatientUuid` only covers the `/patient/new` flow before its redirect lands. An effect-based reset still left one render holding the stale uuid before effects flushed — I tried that first and the new regression test caught it — whereas deriving removes the window entirely. Patient-scoped state is still cleared when the route param changes so stale details don't linger on screen.

Worth noting this also closes the same exposure on the update mutation, which reads the same value; that path pre-dates this PR.

New test: `should not print the previous patient after navigating to another patient`.

apps/registration/src/pages/PatientRegister/__tests__/PatientRegister.test.tsx
AkshataKeshatwarTWer: All three added in `0598fa3d`, plus regression tests for the two fixes above.

One deviation worth calling out: asserting on `renderAsHtml` from this test isn't achievable. `@bahmni/widgets` resolves to its built `dist/index.js`, which inlines its own minified copy of `@bahmni/services` — so `jest.mock('@bahmni/services')` in a page test never intercepts the widget's call, and the assertion reports 0 calls regardless of actual behaviour. That wiring is already covered one layer down by `DocumentPrintButton.test.tsx:191` ("calls renderAsHtml with templateId, context and locale").

So the split is: the widget test asserts the render call, and this page test asserts the props contract — `renderContext` carrying both `patientUuid` and `patientUUID`. That required stubbing `DocumentPrintButton` here, consistent with how the other child components are already mocked in this file.

Added: render-context keys, disabled-while-saving, two options with only one privileged, an unrestricted option with empty privileges, hidden while privileges load, and the cross-patient navigation case.

